PR #3, opened 2020-10-14
Add a Dockerfile for Containerization of REST API.

jmorgan74:
## Changes

- Added a multi-stage dockerfile to compile and run the app.
- Updated the README with a brain dump of info.

### DIFF
--- a/api-microservice/Dockerfile
+++ b/api-microservice/Dockerfile
@@ -1,0 +1,29 @@
+FROM node:12-alpine as builder
+ARG APP_NAME
+ARG BUILD_ID
+ARG VERSION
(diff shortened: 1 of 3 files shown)


api-microservice/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM node:12-alpine as builder
```
MrSimonEmms, line 1: Bloke who gave you that is a clever chap

MrSimonEmms (approved):

api-microservice/Dockerfile
jmorgan74: He is! :-)

PR #4, opened 2020-10-14
chore(setup): Linting, Formatting & Loggin

jmorgan74:
## Changes

- Add ESLint + Airbnb Rules + TypeScript Rules + Prettier.
- Added Winston for logging and created a logging service which logs everything to combined.log and errors to error.log.  Also enabled the Console logger for now.


api-microservice/src/services/Logger.ts
```diff
@@ -0,0 +1,23 @@
...
+export default class LoggerService {
+  private logger: winston.Logger;
+
+  public constructor() {
+    const loggerSettings = {
+      format: winston.format.combine(winston.format.timestamp(), winston.format.json()),
+      level: 'info',
+      transports: [
+        new winston.transports.Console(),
+        new winston.transports.File({ filename: 'error.log', level: 'error' }),
```
MrSimonEmms, line 12: Can we just support console logging? Adding it into logs isn't necessary for Docker (and can cause problems)

We might need to revisit the format in future - I'm not 100% how it plays with Azure logging

MrSimonEmms (changes requested):

api-microservice/src/services/Logger.ts
jmorgan74: yep, no problem, wasn't 100% we needed file logging

jmorgan74: format config is easily maintainable so yes, can change it anytime.

MrSimonEmms (approved):

PR #5, opened 2020-10-15
Configure infrastructure and release process

MrSimonEmms:
This is the base process for building the infrastructure and configuring the release mechanism.

# Infrastructure

The infrastructure is built using Terraform to ensure repeatability. There are two sections:
 - `/infrastructure/common`: common pieces of infrastructure, effectively assigned to the subscription
 - `/infrastructure/environments`: the different environment stacks

## Common

This is infrastructure that is common to the subscription. This is for things where it makes sense for only one, shared instance throughout the application. Things that belong in here are the likes of the container registry, or the key vault. For instance, the container registry will have Docker images published which will be deployed between all environments (it's poor practice to build artifacts between releases - using the version that was tested in all environments).

## Environments

This is the where the application will be run. This current version produces a Kubernetes cluster which, currently, is the entirety of the infrastructure. Future iterations will have CosmosDB, storage accounts etc although currently not required.

# GitHub Actions

There are actions added to run the infrastructures on a per-environment basis. Currently, only `dev` is configured - future environments will be iterated shortly.

## To Do
- There is currently a manual process to associate the IP address of the load balancer. Need to understand how this will be linked up with the .gov.uk domain name so will leave as manual until I have answers.

# Releases

Releases are done using the GitOps workflow. Lots can be found about [GitOps online](https://www.gitops.tech/), but in summary, we have a release manifest (in `/releases`) which describes the Helm charts (in `/charts/app`). The release manifest has a few variables, but the important ones are the image and the tag to track, the URL of the Docker registry and any variables which we want to apply (in this instance, just the URL)

Currently, this is only the `forms-web-app` as there is no release for the `appeals-service-api` - this can be iterated quickly

### DIFF
--- a/.github/actions/kubectl-helm/Dockerfile
+++ b/.github/actions/kubectl-helm/Dockerfile
@@ -1,0 +1,23 @@
+FROM alpine:latest as builder
+ARG KUBE_VERSION=1.18.2
+ARG HELM_VERSION=3.2.0
+ARG OS_TYPE=linux
+ARG ARCH=amd64
+WORKDIR /opt/app
+RUN apk --no-cache add curl \
+  && curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBE_VERSION}/bin/${OS_TYPE}/${ARCH}/kubectl \
+  && chmod +x ./kubectl \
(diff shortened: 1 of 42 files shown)


jmorgan74 (approved):
💥

PR #10, opened 2020-10-19
chore(setup): Setup UX service lint

jmorgan74:
N.B. I'm aware that we can configure one eslint, pretter ruleset for both api and ux services at the top level but I've never done it, still learning. :-)

- Add .eslintrc, .prettierrc and tsconfig files to UX sub dir.


ux-microservice/.eslintrc
```diff
@@ -0,0 +1,17 @@
+{
```
MrSimonEmms, line 1: This is the Typescript config - let me work out a JS version of this and I'll send it you

ux-microservice/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
MrSimonEmms, line 1: If using JS-not-TS, then we don't need this file

MrSimonEmms (changes requested):

ux-microservice/tsconfig.json
jmorgan74: yes but we don't know yet until tomorrow

jmorgan74: I was hoping to go TypeScript with this too so just got it ready.

ux-microservice/.eslintrc
jmorgan74: am aware, see comment below.

MrSimonEmms:
@jmorgan74 I think this should work for you. I need to work on the trailing comma in a function issue, but that's just a rule with I need to disable

```
{
  "parserOptions": {
    "ecmaVersion": 6,
    "sourceType": "module"
  },
  "extends": [ "airbnb-base", "prettier" ],
  "plugins": [ "jest", "prettier" ],
  "root": true,
  "ignorePatterns": [ "node_modules/**", "dist/**" ],
  "rules": {
    "prettier/prettier": "error"
  },
  "env": {
    "es2021": true,
    "node": true,
    "jest/globals": true
  }
}
```

MrSimonEmms (approved):

PR #14, opened 2020-10-26
chore(setup): Add linting for UX service

jmorgan74:
Add code style checking, formatting and husky for formatting on pre-commit hook.

### DIFF
--- a/ux-microservice/bin/www
+++ b/ux-microservice/bin/www
@@ -4,39 +4,17 @@
  * Module dependencies.
  */
 
-var app = require('../app');
-var debug = require('debug')('govuk-fe:server');
-var http = require('http');
-
-/**
- * Get port from environment and store in Express.
- */
-
-var port = normalizePort(process.env.PORT || '3000');
-app.set('port', port);
-
-/**
- * Create HTTP server.
- */
-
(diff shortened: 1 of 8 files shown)


ux-microservice/bin/www
```diff
@@ -80,11 +65,15 @@ function onError(error) {
...
-
 function onListening() {
-  var addr = server.address();
-  var bind = typeof addr === 'string'
-    ? 'pipe ' + addr
-    : 'port ' + addr.port;
-  debug('Listening on ' + bind);
+  const addr = server.address();
+  const bind = typeof addr === 'string' ? `pipe ${addr}` : `port ${addr.port}`;
+  debug(`Listening on ${bind}`);
```
MrSimonEmms, line 71: As this is the only (I think) use of `debug` as the logger, can we change this to `morgan` (or change `morgan` in the app to `debug`)?

MrSimonEmms (changes requested):
Can we select only one logging library - don't care whether `morgan`, `debug` or something else

ux-microservice/bin/www
jmorgan74: I want to install 'winston' as the logger, separate pr

PR #18, opened 2020-10-26
chore(setup): remove the npm ci from the docker run commands

MrSimonEmms:
This was working on Linux, but killing Macs


jmorgan74 (approved):
👍

PR #24, opened 2020-10-27
fix(no-decision) - Create controllers for routes

jmorgan74:
## Changes

- Add controller for eligibility router.
- Add controller for index router.


mmuchlinski (approved):
All looks good

PR #30, opened 2020-10-29
feature(decision-date-expired) - Added the decision date expired page and fixed missing govuk assets

jmorgan74:
## Changes

- Added path to express for 'assets'.
- Created "decision date expired" view template.
- Added jest and an eligibility controller unit test (with dummy passing test which needs to be implemented).


services/forms-web-app/.prettierrc
```diff
@@ -1,5 +1,5 @@
 {
   "singleQuote": true,
   "printWidth": 100,
-  "trailingComma": "all"
+  "trailingComma": "none"
```
MrSimonEmms, line 4: How come we've changed this?

MrSimonEmms (commented):
The image looks ok but not sure on changes to prettier - don't object as such, but the trailing comma does make reading `git diff` and PRs easier

services/forms-web-app/.prettierrc
jmorgan74: I was having issues with this line:

```
app.use(
  '/assets',
  express.static(path.join(__dirname, '../node_modules/govuk-frontend/govuk/assets'))
);
```

MrSimonEmms: Oh right, that would have been adding a trailing to the function - can you try `trailingComma: "es5"` and see if that works https://prettier.io/docs/en/options.html#trailing-commas

I'd prefer to keep it in so it's in the AirBnB style guide

jmorgan74: lovely sir! done! 👍

MrSimonEmms (approved):
Glad to help

PR #46, opened 2020-11-02
fix(deadline-date): date formatting added

mmuchlinski:
## Changes

- Installed 'nunjucks-date-filter' dependency.
- Used the 'date' filter to format the date on the 'expired' page.

### DIFF
--- a/forms-web-app/package-lock.json
+++ b/forms-web-app/package-lock.json
@@ -1186,12 +1186,6 @@
       "integrity": "sha512-f5j5b/Gf71L+dbqxIpQ4Z2WlmI/mPJ0fOkGGmFgtb6sAu97EPczzbS3/tJKxmcYDj55OX6ssqwDAWOHIYDRDGA==",
       "dev": true
     },
-    "@types/parse-json": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/@types/parse-json/-/parse-json-4.0.0.tgz",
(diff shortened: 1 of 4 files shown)


jmorgan74 (dismissed):
👍

package.json
```diff
@@ -21,7 +21,6 @@
     "url": "https://github.com/foundry4/appeal-planning-decision/issues"
   },
   "homepage": "https://github.com/foundry4/appeal-planning-decision#readme",
-  "dependencies": {},
```
MrSimonEmms, line 24: Please restore this line

package-lock.json
```diff
@@ -857,6 +857,11 @@
         "through": ">=2.2.7 <3"
       }
     },
+    "a-sync-waterfall": {
```
MrSimonEmms, line 860: I wouldn't have expected any changes in this file - can you restore the original file and then run `make uninstall` and then `make install`. If it does make any changes, that's fine but please confirm this

MrSimonEmms (changes requested):
Root level changes that I'm not expecting have happened

MrSimonEmms:
Can we also have a linear-commit history? It would be better to rebase this commit to the second "remove" didn't exist (it also breaks things because it's required in semantic-release)

Also, please don't make manual changes in package-lock.json - this is an autogenerated file.

image

package.json
jmorgan74: done

jmorgan74 (approved):

MrSimonEmms (approved):

MrSimonEmms (approved):

PR #49, opened 2020-11-02
fix(invalid-date): added condtional rendering of dateInput macro to show errors

jmorgan74:
## Changes

- Added a condition in the decision-date nunjucks template to render a date input differently if errors have been found.

Screenshot 2020-11-02 at 18 32 16

### DIFF
--- a/forms-web-app/src/controllers/eligibility.js
+++ b/forms-web-app/src/controllers/eligibility.js
@@ -10,7 +10,7 @@ exports.getDecisionDate = (req, res) => {
 
 exports.postDecisionDate = (req, res) => {
   const { body } = req;
-  const { errors } = body;
+  const { errors = [] } = body;
 
   // TO DO: Move this validation out to a service and try to
   // build a custom validator in express-validator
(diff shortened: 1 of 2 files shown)


forms-web-app/src/controllers/eligibility.js
```diff
@@ -10,7 +10,7 @@ exports.getDecisionDate = (req, res) => {
 
 exports.postDecisionDate = (req, res) => {
   const { body } = req;
-  const { errors } = body;
+  const errors = body.errors || [];
```
MrSimonEmms, line 13: If this is to default to an array, you can write this as `const { errors = [] } = body;`

MrSimonEmms (approved):
lgtm

forms-web-app/src/controllers/eligibility.js
jmorgan74: ah, smashing, ta

MrSimonEmms: No probs. I've no issue with the way you did it - it's perfectly valid - but that's just an FYI for how to do it with destructuring

MrSimonEmms (approved):

PR #51, opened 2020-11-03
fix(future-date): entering a date later than today will now throw error

jmorgan74:
## Changes

- Added a check for date in the future.

### DIFF
--- a/forms-web-app/src/controllers/eligibility.js
+++ b/forms-web-app/src/controllers/eligibility.js
@@ -21,14 +21,15 @@ exports.postDecisionDate = (req, res) => {
 
   const date = moment(`${year}-${month}-${day}`, 'Y-M-D', true);
 
-  if (!date.isValid()) {
+  const currentDate = moment();
+
+  if (!date.isValid() || date.isAfter(currentDate)) {
     errors.push({ 'decision-date': 'Invalid date' });
   }
 
-  if (errors) {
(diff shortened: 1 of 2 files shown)


forms-web-app/src/controllers/eligibility.js
```diff
@@ -21,14 +21,15 @@ exports.postDecisionDate = (req, res) => {
...
 
-  if (!date.isValid()) {
+  const currentDate = moment();
+
+  if (!date.isValid() || date.isAfter(currentDate)) {
     errors.push({ 'decision-date': 'Invalid date' });
   }
 
-  if (errors) {
+  if (errors.length) {
```
MrSimonEmms, line 30: Only just thought this, but there's no guarantee that `req.body.errors` will be an error - if someone send in a body with `{ errors: true }` then this will fail (presumably with a 500).

Are we ok with this or do we want some for of `if (!Array.isArray(errors)) { /* do something */ }`?

MrSimonEmms (approved):
I'm happy with this as-is, but may want to consider the error array question.

forms-web-app/src/controllers/eligibility.js
jmorgan74: Happy to change that now.

MrSimonEmms (approved):

PR #56, opened 2020-11-03
fix(link-padding): removed the duplicate form-group div

jmorgan74:
## Changes

- Fixed link padding
- Fixed size of year field


mmuchlinski (approved):
That should do it

PR #59, opened 2020-11-03
Add MongoDB and set up appeals-api-service in deployed infrastructure

MrSimonEmms:
See tickets UCD-865 and UCD-808

### DIFF
--- a/charts/app/templates/appealsServiceApi.yaml
+++ b/charts/app/templates/appealsServiceApi.yaml
@@ -1,12 +1,12 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ include "app.fullname" . }}-form-web-app
+  name: {{ include "app.fullname" . }}-appeals-service-api
   labels:
     {{- include "app.labels" . | nindent 4 }}
 spec:
 {{- if not .Values.autoscaling.enabled }}
-  replicas: {{ .Values.formsWebApp.replicaCount }}
+  replicas: {{ .Values.appealsServiceApi.replicaCount }}
 {{- end }}
   selector:
     matchLabels:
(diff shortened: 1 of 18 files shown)


charts/app/templates/appealsServiceApi.yaml
```diff
@@ -0,0 +1,61 @@
...
+              containerPort: 3000
+              protocol: TCP
+          env:
+            - name: MONGODB_URL
+              valueFrom:
+                secretKeyRef:
+                  key: url
+                  name: {{ .Values.appealsServiceApi.secrets.mongodb.appeals }}
+            - name: SERVER_PORT
+              value: {{ 3000 | quote }}
```
a6software, line 47: Is it possible to use the variable (https://github.com/foundry4/appeal-planning-decision/pull/59/files#diff-f9b76d9e4b03537e44f2a2103cb3a267a0bf92478b0d975bc974fe17e0cde7b6R9) here, and on line 38 also? 

(no idea if it **is** possible, just a Q and potential improvement if so?)

docker-compose.yml
```diff
@@ -55,6 +55,6 @@ services:
 
   # Third-party services
   mongodb:
-    image: mongo:3.4.4
+    image: mongo:3.6.0
```
a6software, line 58: When services are updated like this, is it something that we will "get" automatically when we next run `make serve`, or do we need to explicitly `docker pull` this first?

https://github.com/foundry4/appeal-planning-decision/blob/master/Makefile#L16

e.g. 

```
serve:
	docker-compose pull && docker-compose up
.PHONY: serve
```

?

infrastructure/environments/README.md
```diff
@@ -31,8 +31,12 @@ Infrastructure which the applications are deployed toInfrastructure which is com
...
+| mongodb\_consistency\_policy | Cosmos consistency policy | `string` | `"BoundedStaleness"` | no |
+| mongodb\_databases | List of databases and collections to provision | <pre>list(object({<br>    name = string<br>    collections = list(object({<br>      name = string<br>      default_ttl_seconds = number<br>      indexes = list(object({<br>        keys = set(string)<br>        unique = bool<br>      }))<br>    }))<br>  }))</pre> | n/a | yes |
 | mongodb\_failover\_read\_locations | Locations where read failover replicas are created for MongoDB | `list(string)` | `[]` | no |
-| mongodb\_max\_throughput | Max throughput of the MongoDB database - set in increments of 1,000 between 4,000 and 1,000,000 | `number` | `4000` | no |
+| mongodb\_max\_staleness\_prefix | Represents the number of state requests that are tolerated - min 100,000 for global replication | `number` | `100000` | no |
+| mongodb\_max\_throughput | Max throughput of the MongoDB database - set in increments of 1,000 between 4,000 and 1,000,000 | `number` | `400` | no |
```
a6software, line 39: checking - potential typo here? 

Was previously `4000`, now this has changed to `400` - doc states between 1,000-1,000,000 - is this correct?

```diff
-| `number` | `4000` | no |`
+| `number` | `400` | no |
```

infrastructure/environments/mongodb.tf
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+resource "random_integer" "mongodb" {
+  min = 1000
+  max = 9999
+}
+
+resource "azurerm_cosmosdb_account" "mongodb" {
+  name = format(local.name_format, "mongodb-${random_integer.mongodb.result}")
```
a6software, line 18: not sure what's happening here honestly, but if there's any concern about two deploys ever hitting the same integer, could we potentially use a microsecond timestamp here instead?

a6software (dismissed):
Approving with some potential feedback which may be wrong as I am not devop'sy enough to fully understand all this

:+1:

docker-compose.yml
MrSimonEmms: If the images don't exist on your machine, `docker-compose up` does the `pull` beforehand - same as running `docker run some-image-what-i-dont-have`

infrastructure/environments/README.md
MrSimonEmms: Good catch - the documentation is wrong.

infrastructure/environments/mongodb.tf
MrSimonEmms: > tl;dr - good question, it's fine

The reason we need a pseudo-random integer is because Cosmos requires a unique name and it can take some time to register in Azure that the name is available - you can get into a weird state once deleting a resource and can't recreate it. By creating a random resource, this random number is linked to the workspace and only changes if the resource is destroyed (or tainted)

charts/app/templates/appealsServiceApi.yaml
MrSimonEmms: Yeah, this is one of those things I was in 50/50 about. The fact you've asked the question has tipped the balance in favour of using a variable

MrSimonEmms:
@a6software never apologise for asking questions - it's worth having someone sanity check things as we all make mistakes

a6software (approved):
LGTM!

a6software (approved):

PR #65, opened 2020-11-06
Configure health endpoints for appeal service api

MrSimonEmms:
Ticket UCD-861.

This exposes health endpoints. I've created a common module for things that will be used across packages - the health checks will be required on the future document service, identity service _et al_ so it makes sense to put the boilerplate in a way that can be reused across services.

I've tested the built image (`pinscommonukscontainersdefault.azurecr.io/appeals-service-api:feature-health-endpoints`) in Docker Compose and in my K8S cluster and that's all worked as expected with the health checks running.

One important thing to note is how the [Mongoose health check runs](https://github.com/foundry4/appeal-planning-decision/blob/37f3ad639036d9cda29bb0de213da292e968688b/appeals-service-api/src/lib/healthchecks.js#L10). It may seem strange at first glance, but this is testing the existing pool is able to connect and ping rather than creating a new connection and checking that - I've had it in the past where a pooled connection is not closed properly, but it's able to create a new connection meaning that the pod is unhealthy, but the health check doesn't know it (an edge case admittedly). Using this command tests that the pools that already exist is able to connect.

This is the SQL equivalent of `SELECT 1 + 1`

I've not put this health check in the web app because it doesn't yet need it to my mind - the existing health check of hitting `GET:/` is telling us all we need to know about the app. There may be a time when we need to check it can access the appeals-service-api, but that's a future conversation.


common/src/health.js
```diff
@@ -0,0 +1,78 @@
...
+
+        await new Promise((resolve) => setTimeout(resolve, terminationGrace));
+      }
+    },
+    healthChecks: {
+      '/health': async () => {
+        const startTime = Date.now();
+
+        const results = (
+          await Promise.allSettled(
```
a6software, line 34: til: the result of `.allSettled` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

could not work out where `value` was coming from on line 51

a6software (approved):
LGTM - not able to validate the github workflow as I don't have experience with that, but the rest looks good

danWatkinson (approved):
interesting; i literally did the dependency-wiring exercise for another project last week but hacked it with local file-references, and managed to get away without needing to do any volume-mounting in docker-compose.. just got the docker images to build in such a way that the modules were there and lined up properly.. not saying either's better but as a comparison..

assume we're not exposing ourselves to any overly-zealous healthchecking when we ping the db? eg. I know azure likes to ping every second; healthchecks that do database calls can be expensive then? (straw man- i understand why you want to see db connections working, might be that we end up wanting multiple healthchecks for different purposes but if so that's tomorrow's problem..)

MrSimonEmms:
There should be no issues with the health checks timings. As we using Kubernetes, we have complete control over the `periodTime` - I've left it as the default 10 seconds, which should be fine.

PR #66, opened 2020-11-09
feat(forms-web-app): add Listed Building form

a6software:
Adds listed building form happy & sad paths, express validator, and template for next step

### DIFF
--- a/forms-web-app/package.json
+++ b/forms-web-app/package.json
@@ -9,7 +9,8 @@
     "lint": "eslint ./",
     "lint:fix": "eslint ./ --fix",
     "format": "prettier --write './**/*.js'",
-    "test": "jest"
+    "test": "jest",
+    "test:watch": "jest --watchAll --runInBand"
   },
   "lint-staged": {
     "./**/*.js": "npm run lint:fix",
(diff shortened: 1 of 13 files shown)


MrSimonEmms (approved):
LGTM

Have run this locally and all seems to run ok. Good work on getting some tests in.

Only comment (and it is a comment rather than a criticism or RFC) is that your consts are all capitalised which seems to be a divergence on the existing style - it all passes the linting, so I'm fine with that.

a6software:
> LGTM
> 
> Have run this locally and all seems to run ok. Good work on getting some tests in. I don't know the functionality that you're trying to achieve, so may be worth getting someone better placed to look at that.
> 
> Only comment (and it is a comment rather than a criticism or RFC) is that your consts are all capitalised which seems to be a divergence on the existing style - it all passes the linting, so I'm fine with that.

Cheers. 

I'm definitely open to changing the upper case constants. Just a personal preference. My bigger concern is the abundance of strings dotted around the place, and the potential for mistakes with these. I'm compiling a list of things like this that I'd like to discuss and standardise as a team. One for the retro or similar discussion.

I'm going to merge this as not had a further review yet, and I need this as the basis of my next ticket. 

Thanks

PR #69, opened 2020-11-10
Fix/ucd 744 decision date

Fenisay:
2 small fixes to close the ucd 744
https://pins-ds.atlassian.net/jira/software/projects/UCD/boards/30/backlog?selectedIssue=UCD-744

## Changes

fix(decision-date): Inherit from the good govuk stylesheets.
fix(decision-date): The decision date year field has to be 4 digits.

### DIFF
--- a/forms-web-app/src/controllers/eligibility.js
+++ b/forms-web-app/src/controllers/eligibility.js
@@ -19,7 +19,7 @@ exports.postDecisionDate = (req, res) => {
   const month = req.body['decision-date-month'];
   const year = req.body['decision-date-year'];
 
-  const date = moment(`${year}-${month}-${day}`, 'Y-M-D', true);
+  const date = moment(`${year}-${month}-${day}`, 'YYYY-M-D', true);
 
   const currentDate = moment();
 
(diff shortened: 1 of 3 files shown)


forms-web-app/src/controllers/eligibility.js
```diff
@@ -19,7 +19,7 @@ exports.postDecisionDate = (req, res) => {
   const month = req.body['decision-date-month'];
   const year = req.body['decision-date-year'];
 
-  const date = moment(`${year}-${month}-${day}`, 'Y-M-D', true);
+  const date = moment(`${year}-${month}-${day}`, 'YYYY-M-D', true);
```
MrSimonEmms, line 22: I know moment was done before your time (so this is just a comment, not a request to change), but just noticed the [project status](https://momentjs.com/docs/)

> We now generally consider Moment to be a legacy project in maintenance mode. It is not dead, but it is indeed done.

It may well be worth a future piece of work to replace moment with either the native `Date` or `Intl` class or [Luxon](https://moment.github.io/luxon/) (the interface is similar as it's made by the same people)

MrSimonEmms (approved):
Note for me in the future...

GH Actions needs to have `pull_requests` enabled as well as this is coming from a fork. I need to update the GH Actions to also respond to PRs. The local build is failing on the Docker build because it doesn't (correctly) have access to the secrets

forms-web-app/src/controllers/eligibility.js
Fenisay: Ok I keep this in mind, thanks for the info.

PR #70, opened 2020-11-10
Feature/ucd 744 decision date

Fenisay:
2 small fixes in 

https://pins-ds.atlassian.net/browse/UCD-744


MrSimonEmms (approved):
Thoughts about replacing moment are documented in #69. Not essential, but may be worth considering

PR #71, opened 2020-11-10
fix: use pre-existing public ip address for nginx ingress load balancer

MrSimonEmms:


MrSimonEmms:
This PR cannot be merged immediately as it will change the IP address of the load balancer that Kubernetes uses - when this is done must be approved by a wider audience (with @karunak21 having the final say)

danWatkinson (approved):

PR #72, opened 2020-11-10
feat(forms-web-app): add "your appeal statement" page

a6software:
Adds "your appeal statement" page and placeholder for next step - /task-list

### DIFF
--- a/forms-web-app/src/app.js
+++ b/forms-web-app/src/app.js
@@ -8,6 +8,7 @@ const dateFilter = require('nunjucks-date-filter');
 
 const indexRouter = require('./routes/index');
 const eligibilityRouter = require('./routes/eligibility');
+const taskListRouter = require('./routes/task-list');
 
 const app = express();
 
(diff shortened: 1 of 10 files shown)


Fenisay:
Just so you know, on my PR I added govuk-body in the main.njk because otherwise we had to write it everywhere as you did and from what I understood everything should inherit or override it. I'm not 100% confident, what do you think about that ?

Fenisay (dismissed):

a6software:
> Just so you know, on my PR I added govuk-body in the main.njk because otherwise we had to write it everywhere as you did and from what I understood everything should inherit or override it. I'm not 100% confident, what do you think about that ?

Nice, thanks for flagging. CSS isn't my strong point. I was wondering how that was working differently on the prototype. 

I have removed all the extra class attributes. 

Please could you re-review? Thanks

Fenisay (approved):

Fenisay:
Well me neither, at least at the beginning of each project... I think I have a brain issue and no matter how much time I spend on it, if I do not practice 12h everything is being flushed. Frameworks like saas helped me a little bit though.

a6software:
thanks

PR #74, opened 2020-11-10
Local Planning Authorities list

MrSimonEmms:
Ticket UCD-918

Creates two endpoints in the appeals-service-api (a list and a get) to show the local planning authorities.

This data comes from a static file, but there is a future job to get this mounted as ConfigMap in K8S and automatically ingest the data from https://geoportal.statistics.gov.uk/datasets/local-planning-authorities-april-2020-uk-bfc

By default, there are two "in trial" LPAs:
 - Adur (as it's top alphabetically)
 - Telford and Wrekin (because I live there)

These won't necessarily be the real trialists in production. Again, we have complete control over who they are - these are just some trialists so we can check things work

This exposes two endpoints:
 - `GET:/local-planning-authorities` - returns all LPAs. Optional `?name=...` for filtering by name only
- `GET/:local-planning-authorities/:id` - returns the specific LPA by ID


danWatkinson (approved):
looks good; liking the jest mocking etc in the unit-tests. also taking note of the kind of documentation expected..

mmuchlinski (approved):

PR #75, opened 2020-11-11
feat(forms-web-app): add check your answers

a6software:
Adds task list & check your answer pages, basic implementation for your details (first task), and
save / retrieve form submission via middleware


appeals-service-api/src/schemas/appeals.js
```diff
@@ -11,10 +11,17 @@ const AppealSchema = new Schema(
...
+    appellantName: {
       type: String,
       required: true,
     },
+    appellantEmail: {
+      type: String,
+      required: true,
+    },
+    planningApplicationNumber: {
+      type: Number,
```
a6software, line 23: note: not required, this is something we need to discuss more I think

docker-compose.yml
```diff
@@ -29,6 +29,7 @@ services:
     image: node:14-alpine
     environment:
       APPEALS_SERVICE_API_URL: http://appeals-service-api:3000
+      SESSION_KEY: some_secure_key_goes_here
```
a6software, line 32: not sure if this is the best place for this.

I'm using the session to keep track of the current `uuid` for the "in progress" form submission

forms-web-app/.eslintrc
```diff
@@ -1,6 +1,6 @@
 {
   "parserOptions": {
-    "ecmaVersion": 6,
+    "ecmaVersion": 9,
```
a6software, line 3: allows promises and `...` spreading in objects

forms-web-app/Dockerfile
```diff
@@ -9,6 +9,7 @@ ENV VERSION="${VERSION}"
 ENV PORT=3000
 # Do not rely on NODE_ENV - exists for performance reasons only
 ENV NODE_ENV=production
+RUN npm run sass
```
a6software, line 12: This might need to be captured somewhere else in the build process?

Right now it's a manual step.

forms-web-app/package.json
```diff
@@ -17,14 +17,17 @@
     "./**/*.{js, json}": "npm run format"
   },
   "dependencies": {
+    "@ministryofjustice/frontend": "0.0.21",
```
a6software, line 20: This is what the alpha uses for the task list view

forms-web-app/src/app.js
```diff
@@ -5,19 +5,38 @@ const path = require('path');
...
+const sessionConfig = {
+  secret: process.env.SESSION_KEY,
+  resave: false,
+  saveUninitialized: true,
+  cookie: {},
+};
+
+if (app.get('env') === 'production') {
+  app.set('trust proxy', 1); // trust first proxy
+  sessionConfig.cookie.secure = true; // serve secure cookies
```
a6software, line 29: might not be correct, copied wholesale from the npm docs

forms-web-app/src/config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  APPEALS_SERVICE_API_URL: process.env.APPEALS_SERVICE_API_URL,
```
a6software, line 2: not sure how else to do this. Don't want references to `process.env` dotted everywhere we need an API call. There may be a better / more preferable way?

forms-web-app/src/controllers/task-list.js
```diff
@@ -1,3 +1,117 @@
 exports.getTaskList = (req, res) => {
-  res.render('task-list/index');
+  // TODO: derive these values from the current session data
```
a6software, line 2: outside the scope of this ticket

forms-web-app/src/controllers/your-details.js
```diff
@@ -0,0 +1,31 @@
...
+exports.getYourDetails = (req, res) => {
+  res.render(VIEW.YOUR_DETAILS, {
+    appeal: req.session.appeal,
+  });
+};
+
+exports.postYourDetails = async (req, res) => {
+  const body = {
+    appellantName: req.body['appellant-name'],
+    appellantEmail: req.body['appellant-email'],
```
a6software, line 16: not validated - outside the scope of this ticket as this page was only added to prove the concept of save / reload / redisplay form data

forms-web-app/src/controllers/your-details.js
```diff
@@ -0,0 +1,31 @@
...
+
+  try {
+    const uuid = (req.session && req.session.uuid) || '';
+    const appealJson = await appealsApi(body, uuid);
+
+    req.session.uuid = uuid;
+    req.session.appeal = appealJson;
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(e);
```
a6software, line 27: do we have a more preferable way of logging / tracking errors ?

forms-web-app/src/lib/appeals-api-wrapper.js
```diff
@@ -0,0 +1,18 @@
...
+    appealsServiceApiUrl = `${appealsServiceApiUrl}/${uuid}`;
+    method = 'PUT';
+  }
+
+  return fetch(appealsServiceApiUrl, {
+    method,
+    body: JSON.stringify(body),
+    headers: { 'Content-Type': 'application/json' },
+  }).then((apiResponse) => apiResponse.json());
+};
```
a6software, line 18: this is a stab at a single call to provide upsert functionality

forms-web-app/src/routes/check-answers.js
```diff
@@ -0,0 +1,10 @@
+const express = require('express');
+const fetchExistingAppealMiddleware = require('../middleware/fetch-existing-appeal');
+
+const checkAnswersController = require('../controllers/check-answers');
+
+const router = express.Router();
+
+router.get('/', [fetchExistingAppealMiddleware], checkAnswersController.getCheckAnswers);
```
a6software, line 8: an example of a route that uses the middleware to pre-populate the appeal form data

forms-web-app/src/routes/validators/listed-building.js
```diff
@@ -20,22 +17,6 @@ const rules = () => {
   ];
 };
 
-const validator = (req, res, next) => {
```
a6software, line 23: extracted to own file

forms-web-app/src/routes/validators/decision-date.js
```diff
@@ -0,0 +1,36 @@
...
+    body('decision-date-year').notEmpty(),
+    // body('decision-date-year').custom(decisionDateCombiner).isDate().custom(deadlineDateValidator),
+  ];
+};
+
+module.exports = {
+  decisionDateCombiner,
+  deadlineDateValidator,
+  rules,
+};
```
a6software, line 36: changes here:

1. remove the validator
2. rename the file

rest is as was

forms-web-app/src/sass/save-return.scss
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.pins-sidebar {
+  margin-top: 35px;
+}
+
+.pins-reference-code-panel {
+  margin-bottom: 50px;
+  margin-top: 0;
+}
```
a6software, line 20: copied wholesale from the alpha

forms-web-app/src/views/check-answers/index.njk
```diff
@@ -0,0 +1,381 @@
...
+                }
+              ]
+            }
+          },
+          {
+            key: {
+              text: "Your email"
+            },
+            value: {
+              text: appeal.appellantEmail
```
a6software, line 72: how this all comes together

forms-web-app/src/views/includes/save.njk
```diff
@@ -0,0 +1,6 @@
+<div class="govuk-grid-column-one-third">
+  <aside class="pins-sidebar">
+    <h3 class="govuk-heading-m">Return to this later</h3>
+    <p><a href="save-return/return-info">Save my progress</a></p>
+  </aside>
+</div>
```
a6software, line 6: essentially a stub - it's there on the alpha, but this functionality is not implemented as part of this ticket

forms-web-app/tests/unit/controllers/your-details.test.js
```diff
@@ -0,0 +1,23 @@
...
+describe('controller/your-details', () => {
+  describe('getYourDetails', () => {
+    it('should call the correct template', () => {
+      yourDetailsController.getYourDetails(req, res);
+
+      expect(res.render).toHaveBeenCalledWith('your-details/index', { appeal: undefined });
+    });
+  });
+
+  xdescribe('postYourDetails', () => {
```
a6software, line 16: intentionally skipped as this approach may not be wanted

forms-web-app/Dockerfile
a6software: Seems to have broken the build.

appeals-service-api/src/schemas/appeals.js
MrSimonEmms: Is this the UUID for the user?

docker-compose.yml
MrSimonEmms: I like that. Makes it nice and easy for me to generate a random secret in the deployment

forms-web-app/.eslintrc
MrSimonEmms: Booyah!

forms-web-app/Dockerfile
MrSimonEmms: The idea with the Dockerfile is that the artifact is built before the `docker build...` is run. It's probably more useful having an `npm run build` step which consists of:
```
{
  "scripts": {
    "prebuild": "rm -Rf ./dist",
    "build": "npm run build:sass && npm run build:app",
    "build:app": "cp -Rf ./src ./dist",
    "build:sass": "<command to build sass files in ./dist>"
  }
}
```

Then change all the commands that are `src/server.js` to be `dist/server` and change the `ADD src src` to `ADD dist dist` in the Dockerfile

forms-web-app/package.json
```diff
@@ -17,14 +17,17 @@
...
     "compression": "^1.7.4",
     "cookie-parser": "~1.4.4",
     "express": "~4.16.1",
+    "express-session": "^1.17.1",
     "express-validator": "^6.6.1",
     "govuk-frontend": "^3.9.1",
     "http-errors": "~1.6.3",
     "lusca": "^1.6.1",
     "moment": "^2.29.1",
+    "node-fetch": "^2.6.1",
```
MrSimonEmms, line 30: Take it this is for making the API calls?

forms-web-app/src/app.js
MrSimonEmms: I'd prefer not relying on `production` - would prefer having a `config.server.useSecureSessionCookie` boolean that I can control instead of `if (app.get('env') === 'production') { ... }`

For the most part, any HTTP headers for production will be set in the nginx-ingress section so you don't need to worry about that.

forms-web-app/src/config.js
MrSimonEmms: Yeah, basically correct. See how I did it in [appeals-service-api](https://github.com/foundry4/appeal-planning-decision/blob/master/appeals-service-api/src/lib/config.js). Separating them into sections makes it more readable. Suggest...

```
module.exports = {
   appeals: {
     url: process.env.APPEALS_SERVICE_API_URL
   }
}
```

forms-web-app/src/controllers/your-details.js
MrSimonEmms: I would prefer to use [pino](https://getpino.io) as that gives a nice JSON object and can use [pino-pretty](https://github.com/pinojs/pino-pretty). See [appeals-service-api](https://github.com/foundry4/appeal-planning-decision/blob/master/appeals-service-api/src/lib/logger.js)

MrSimonEmms (commented):
Looking really good so far. I've put some comments into answer your questions/provide guidance. If you're feeling better tomorrow, we can talk more about this

appeals-service-api/src/schemas/appeals.js
a6software: No, this appears to be something else - https://has-appeal-alpha.herokuapp.com/v7/application-number

Unless I have misunderstood. 

I'm using the `uuid` from Mongo as the user's `uuid` - maybe they will do all of this jazz from the planning application number though, which... would probably make more sense from a UX pov. Good Q

MrSimonEmms: Aaaaah, my mistake

forms-web-app/src/routes/validators/decision-date.js
Fenisay: Well I was planning to talk about this, so on my side I will add planning-department and probably split the controllers too.

appeals-service-api/src/schemas/appeals.js
a6software: I don't have an answer to this right now essentially. It's been raised on teams and is under active discussion. For now, the uuid from mongo is the unique reference I'm storing. From a "front end" dev pov it's a simple thing to change. From a back end dev perspective, it's a can of worms that PINS need to open, discuss, and eat like an I'm a celebrity challenge. Sorry, I don't really know where I'm going with this metaphor.

Fenisay: I think I get it... that's deep...

MrSimonEmms: > I don't have an answer to this right now essentially. It's been raised on teams and is under active discussion. For now, the uuid from mongo is the unique reference I'm storing. From a "front end" dev pov it's a simple thing to change. From a back end dev perspective, it's a can of worms that PINS need to open, discuss, and eat like an I'm a celebrity challenge. Sorry, I don't really know where I'm going with this metaphor.

I loved the metaphor. Can I be Biggins in this?

a6software: Always :D haha

forms-web-app/Dockerfile
a6software: OK, changes made: https://github.com/foundry4/appeal-planning-decision/pull/75/commits/9281a716ab82f670be6f56a28341ce9c02e55d68#diff-3c01ce3ff376b8c4349da777b73024a11cb3b3ead3cec6cd97ae22eae0928996R6

I have left `start:dev` using `src/..` 

Dockerfile change: https://github.com/foundry4/appeal-planning-decision/pull/75/commits/9281a716ab82f670be6f56a28341ce9c02e55d68#diff-a1db04c115605ab99fc2908fe6981ed94ef22946dffc7cc860bef97514e7e58fR12

`Add . .` would be copying in both `src` and `dist` afaik, so have been more explicit to just copy in `dist` - lmk if wrong though!

forms-web-app/src/app.js
a6software: OK, does this mean I can get rid of :

`app.set('trust proxy', 1); // trust first proxy` ? 

I have made the other changes as requested. 

One thing to flag: https://github.com/foundry4/appeal-planning-decision/pull/75/commits/9281a716ab82f670be6f56a28341ce9c02e55d68#diff-9d4e2ee2e03d9f5247e0df2662fb1491e6e09dcbc72c6e84c481f0f9dcf1e395R10

Not sure if that's the best way. All env vars are strings afaik, so an explicit check for `"true"` is the best I could come up with.

forms-web-app/src/config.js
a6software: done as suggested

forms-web-app/src/controllers/your-details.js
a6software: I believe I have done this in line with the appeals service api implementation, cheers :+1:

forms-web-app/package.json
a6software: Sorry, missed this Q - Yes, that is correct

appeals-service-api/src/schemas/appeals.js
```diff
@@ -11,11 +11,11 @@ const AppealSchema = new Schema(
         unique: true,
       },
     },
-    appellantName: {
+    'appellant-name': {
       type: String,
       required: true,
     },
-    appellantEmail: {
+    'appellant-email': {
```
a6software, line 18: by keeping the keys in the same format, we don't need to map between camel and kebab for no real reason

forms-web-app/src/lib/logger.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Logger
+ *
+ * Creates a log instance with common configuration that can be
```
Fenisay, line 4: Cool ! We were in lack of logs.

forms-web-app/src/public/stylesheets/main.css
```diff
@@ -4460,3 +4460,121 @@
   @media (min-width: 40.0625em) {
     .govuk-\!-width-one-quarter {
```
Fenisay, line 4461: There is an error here but I won't tell you were ! Good luck

forms-web-app/src/routes/eligibility.js
```diff
@@ -35,7 +33,7 @@ router.get('/listed-building', listedBuildingController.getListedBuilding);
 router.post(
   '/listed-building',
   listedBuildingValidationRules(),
-  listedBuildingValidator,
```
Fenisay, line 38: So validator is the one by defaults with common validation result and each service with his own validator. Are we good ?

Fenisay (approved):

forms-web-app/src/app.js
MrSimonEmms: Yeah, that's exactly the way that works

FYI, if it's a boolean then I either use `{ item: process.env.BOOL === 'true' }` (if you want it to default to false or `{ item: process.env.BOOL !== 'false' }` (if you want it to default to false. I think defaulting this to `false` is the correct way to go about it

forms-web-app/src/routes/validators/decision-date.js
```diff
@@ -0,0 +1,36 @@
...
+    return value;
+  }
+};
+
+const rules = () => {
+  return [
+    body('decision-date-day').notEmpty(),
+    body('decision-date-month').notEmpty(),
+    body('decision-date-year').notEmpty(),
+    // body('decision-date-year').custom(decisionDateCombiner).isDate().custom(deadlineDateValidator),
```
MrSimonEmms, line 28: Can this be removed?

MrSimonEmms (changes requested):
I need to do some work on the Helm chart to get the new envvars

As we don't have automated releases (yet), I'm happy to get this into `master` first and do after, but it won't be able to be released until that's done

forms-web-app/src/routes/eligibility.js
a6software: Yes, I was following the pattern already set - I think by Jon Morgan, but maybe wrong.

The ordering of these 'two' middleware is important. 

The rules are specific to the given endpoint. There are `listedBuildingValidationRules()` and `decisionDateValidationRules()` and `yourDetailsRules()`.

Each function returns one or more express validator validation middlewares - https://express-validator.github.io/docs/check-api.html

All of these are run against the `req` and passed on to the next middleware.

Finally, the `validator` is common - and really not a validator at all, so maybe the name is bad and misleading - but is concerned with taking any validation errors from the previous middlewares (via examining the `req` object) and then adding the errors and error summary as properties on the `req.body` for display in the template.

The (badly named) `validator` is the same for every route so was extracted out to a separate file as previously it was being duplicated in each validation process.

forms-web-app/src/public/stylesheets/main.css
a6software: :D Well, these are built via the govuk includes - https://github.com/foundry4/appeal-planning-decision/blob/master/forms-web-app/src/sass/main.scss#L1 - so maybe we need to raise this with them upstream? 

fwiw idk what the error is :/

forms-web-app/src/routes/validators/decision-date.js
a6software: Yup - removed. 

I had left this file alone as when I started this work, the decision date validation was in progress as a separate ticket and didn't want to cause conflicts. That said, I then renamed the file and extracted out some stuff - so I guess that went out of the window anyway. 

This file needs a refactor - can get rid of moment imo but it's outside the scope of this already overly large ticket

forms-web-app/src/routes/eligibility.js
a6software: `validator` has been renamed to `validation-error-handler` to better represent what is actually happening.

MrSimonEmms:
We'll also likely need to add redis in to allow for sessions to be stored across replicas

MrSimonEmms:
I've added the infrastructure (needs testing) and rebased from `master` (hope that's not going to cause you any issues @a6software - I can help if it does).

forms-web-app/tests/unit/validators/eligibility/decision-date.test.js
```diff
@@ -0,0 +1,140 @@
...
+          JSON.stringify({ msg: 'Decision date expired', deadlineDate: '1920-08-28T22:59:59.999Z' })
+        )
+      );
+    });
+
+    it('should resolve if valid', async () => {
+      const now = new Date();
+      const req = {
+        body: {
+          'decision-date-day': now.getDay(),
```
MrSimonEmms, line 131: @a6software I think this should be `getDate()` - `getDay` gives the day number and is failing on Sunday (day `0`)

MrSimonEmms (changes requested):

MrSimonEmms:
Infrastructure is now in and I'm happy with it - this should create a Redis instance and configure all the secrets for it to work. I've also added Redis into the dev stack

forms-web-app/tests/unit/validators/eligibility/decision-date.test.js
a6software: Fixed, thanks

MrSimonEmms (dismissed):

MrSimonEmms (approved):

a6software:
cheers all :+1:

PR #77, opened 2020-11-12
fix(decision-date): rebrand planning authority into planning department

Fenisay:
Huge PR

### DIFF
--- a/forms-web-app/src/views/eligibility/decision-date.njk
+++ b/forms-web-app/src/views/eligibility/decision-date.njk
@@ -33,7 +33,7 @@
           namePrefix: "decision-date",
           fieldset: {
             legend: {
-              text: "What's the decision date on the letter from the Local Planning Authority​​?",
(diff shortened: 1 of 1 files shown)


a6software (commented):
just checking, It Was Capitalised Before, is it ok now to be lower case ?

MrSimonEmms:
@a6software dunno. Imagine that Ed is probably the chap to check with

MrSimonEmms:
It's lowercase on https://www.gov.uk/planning-permission-england-wales

Fenisay:
> just checking, It Was Capitalised Before, is it ok now to be lower case ?

Yep, that's normal but thanks for asking !

MrSimonEmms (approved):

Fenisay:
> It's lowercase on https://www.gov.uk/planning-permission-england-wales
> 
> image

in fact it changed many times but this US was not updated. From now on I will stick to the prototype (which also changed :p)

Fenisay:
Thanks

PR #88, opened 2020-11-16
fix(forms-web-app): change listed building error wording when no option selected

a6software:
Change to match the error message in the MS Word error doc


MrSimonEmms (approved):

a6software:
ty

PR #99, opened 2020-11-17
feat(api-documentation): added swagger and jsdoc for live api documen…

rcmsoftwaredevelopment:
…tation under /api-docs

## Ticket Number
https://pins-ds.atlassian.net/browse/UCD-963

## Description of change
Added api documentation via swagger and jsdoc

## Checklist

- [ ] New features have tests

## Important


appeals-service-api/src/docs/components.yml
```diff
@@ -0,0 +1,54 @@
...
+        england:
+          type: boolean
+        wales:
+          type: boolean
+      example:
+        id: E60000281
+        name: Adur LPA
+        inTrial: true
+        england: true
+        wales: true
```
MrSimonEmms, line 32: Picky I know, but Adur should be `wales: false`

MrSimonEmms (commented):
LGTM - I need to run it locally which I'll do after demo

MrSimonEmms (approved):
I blimmin' love this. Well done @rcmsoftwaredevelopment 

If you can be bothered to make the change to Adur not being in Wales that would be good, but equally fine for a bit of Welsh terratorial expansion :wales:

MrSimonEmms (changes requested):
The `GET:/appeals` and `GET:/local-planning-authorities` endpoints are wrong. These are pagination objects with an array of objects. The definition should be something like this (doubt this will work as-is, but shows the idea):

```yaml
responses:
  '200':
    description: OK
    content:
      application/json:
        schema:
          type: object
          properties:
            count: number
            total: number
            page: number
            pageCount: number
            data:
              type: array
              items:
                $ref: '#/components/schemas/LPA'
```

If you can get a way of reusing a Pagination model with the data schema defined that would be wonderful

appeals-service-api/src/server.js
```diff
@@ -29,6 +31,8 @@ module.exports = () => {
       })
     )
     .use(bodyParser.json())
+    .use(helmet()) /* set security HTTP headers */
```
MrSimonEmms, line 34: Helmet not required - will be doing all this in Kubernetes. Prob best to remove

appeals-service-api/src/controllers/localPlanningAuthorities.spec.js
```diff
@@ -92,10 +92,10 @@ describe('LPAs controller test', () => {
 
       expect(res.send).toBeCalledWith({
         data,
-        count: data.length,
-        total: data.length,
         page: 1,
-        pageCount: 1,
+        limit: data.length,
+        totalPages: 1,
```
MrSimonEmms, line 97: :heart:

MrSimonEmms (changes requested):
Can we remove Helmet? It's being done in Kubernetes anyway

MrSimonEmms (approved):

PR #110, opened 2020-11-20
Inject Key Vault secrets as Kubernetes secrets

MrSimonEmms:
## Ticket Number

UCD-995

## Description of change

Configures akv2k8s on the cluster. This is to enable loose-coupling of the Key Vault to the application and enabling injection of individual secrets as environment variables so the individual applications don't need any knowledge of Key Vault (or vice versa).

This removes the individual secret configuration in the `/infrastructure/k8s/setup.sh` file as the secrets are now injectable from Key Vault. In future, any secret should be added to Key Vault rather than being put in the application.

## Checklist

- [ ] New features have tests

## Important


MrSimonEmms:
This will bring down the existing Kubernetes cluster so need approval from @karunak21 and @danWatkinson before deploying

Fenisay (approved):

MrSimonEmms:
Cards on the table. I'm not 100% certain that the CI/CD job will run. I'm about 98% certain it will, and I've replicated things I'm doing locally but there's always the chance I've missed something

Fenisay (approved):

PR #117, opened 2020-11-23
feat(grounds-appeal): add grounds of appeal

Fenisay:
Add of appellant submission appeal statement. Previously known as grounds appeal. I let the file by this name as there is already an other appeal-statement page in the eligibility workflow but if you think that it's more confusing than helpful, let me know.

https://pins-ds.atlassian.net/browse/UCD-828
https://pins-ds.atlassian.net/browse/UCD-832


danWatkinson (dismissed):
Looks good to me, man..

I've given it my 'tick' but please get a development approval as well, I'm just validating the testing..

a6software (approved):

PR #125, opened 2020-11-24
Fix/fwa upload temp dir

MrSimonEmms:
## Ticket Number

UCD-984

## Description of change

Update the `express-fileupload` [config](https://www.npmjs.com/package/express-fileupload)
 - update `tmpDir`
 - update `limits.fileSize`
 - set Helm/k8s to debug uploads by default (future piece to investigate changing internal `console.log` to using pino)

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/forms-web-app/src/config.js
+++ b/forms-web-app/src/config.js
@@ -6,8 +6,10 @@ module.exports = {
   fileUpload: {
     abortOnLimit: process.env.FILE_UPLOAD_ABORT_ON_HITTING_FILE_SIZE_LIMIT === 'true',
     debug: process.env.FILE_UPLOAD_DEBUG === 'true',
-    // default: 1024 * 1024 * 1024 = 1gb
-    maxFileSizeBytes: Number(process.env.FILE_UPLOAD_MAX_FILE_SIZE_BYTES || 1024 * 1024 * 1024),
+    limits: {
+      // default: 1024 * 1024 * 1024 = 1gb
+      fileSize: Number(process.env.FILE_UPLOAD_MAX_FILE_SIZE_BYTES || 1024 * 1024 * 1024),
+    },
     tempFileDir: process.env.FILE_UPLOAD_TMP_PATH,
     useTempFiles: process.env.FILE_UPLOAD_USE_TEMP_FILES === 'true',
   },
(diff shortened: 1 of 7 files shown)


forms-web-app/src/validators/appellant-submission/appeal-statement-schema.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
           'Doc is the wrong file type: The file must be a DOC, DOCX, PDF, TIF, JPG or PNG'
         );
 
-        validateFileSize(size, config.fileUpload.maxFileSizeBytes);
```
a6software, line 37: explains where that value was coming from - by using my own var names + intermingling with the express-fileupload config, I really made that hard to understand.

forms-web-app/src/config.js
```diff
@@ -4,11 +4,13 @@ module.exports = {
     url: process.env.APPEALS_SERVICE_API_URL,
   },
   fileUpload: {
-    abortOnLimit: process.env.FILE_UPLOAD_ABORT_ON_HITTING_FILE_SIZE_LIMIT === 'true',
```
a6software, line 7: express-fileupload has a built in way to handle file uploads that are too big.

This works - but we do not have much control over exactly how. The issue being that express-fileupload throws a 413 if the file size is too big. All good. However, we need to display that error on the page that had the issue, which the middleware made impossible (that I could see, anyway). 

Instead, we are going to rely on our own internal validator for file size (which works) - but has the draw back of accepting the full file first, and then determining file size. I'm not a huge fan of this, as I think this is a regression over what we have. But given the time constraints, this does meet the AC still.

MrSimonEmms: Yeah, I agree with this as a first iteration. I agree that uploading the file and handling it is not great, but much better than it sending a 413 error with no control.

We can put a ticket in the backlog to investigate doing this better (personally, I doubt we'll ever get to it but so long as it's captured)

MrSimonEmms (commented):
Looks good, but want to check in a Kubernetes deployment before it gets merged.

Will need someone else to approve as I was a contributer

MrSimonEmms:
Have checked this and it uploads files fine. Obviously, we need to work out what to do with them next, but that's beyond the scope of this ticket

Fenisay (approved):

PR #132, opened 2020-11-25
Feature/ucd 752 your details

Fenisay:
UCD-752
UCD-753


forms-web-app/src/views/appellant-submission/applicant-name.njk
```diff
@@ -0,0 +1,53 @@
...
+          }
+
+        }) }}
+
+        {{ govukButton({
+          text: "Save and continue",
+          type: "submit"
+        }) }}
+
+        <input type="hidden" name="sections-completed" value="1">
```
a6software, line 46: These 3 hidden fields can be deleted - something the prototype uses but we don't

forms-web-app/src/views/appellant-submission/your-details.njk
```diff
@@ -2,16 +2,25 @@
...
+      titleText: "There is a problem",
+      errorList: errorSummary
+    }) }}
+  {% endif %}
+
   <div class="govuk-grid-row">
 
     <div class="govuk-grid-column-two-thirds">
-      <form action="" method="post" novalidate>
+      <form action="/appellant-submission/your-details" method="POST" novalidate>
```
a6software, line 23: not saying it's right or wrong btw, just that leaving `form action=""` means the form will always post back to the current route - which is handy when routes change like they have in this PR / during dev. That's why I've been leaving them as empty. If you'd prefer them explicit, I'm happy to conform.

forms-web-app/src/routes/appellant-submission/who-are-you.js
```diff
@@ -0,0 +1,19 @@
...
+
+const whoAreYouController = require('../../controllers/appellant-submission/who-are-you');
+const { validationErrorHandler } = require('../../validators/validation-error-handler');
+const {
+  rules: whoAreYouValidationRules,
+} = require('../../validators/appellant-submission/who-are-you');
+
+const router = express.Router();
+
+router.get('/who-are-you', whoAreYouController.getWhoAreYou);
```
a6software, line 11: Looking at the matching controller action here, I think this will need the existing appeal:

```
router.get('/who-are-you', [fetchExistingAppealMiddleware], whoAreYouController.getWhoAreYou);
```

forms-web-app/src/validators/appellant-submission/applicant-name.js
```diff
@@ -0,0 +1,19 @@
...
+  return [
+    body('behalf-appellant-name')
+      .notEmpty()
+      .withMessage('Enter the name your appealing for')
+      .bail()
+      .matches(/^[a-z\-' ]+$/i)
+      .withMessage('Name must only include letters a to z, hyphens, spaces and apostrophes')
+      .bail()
+      .isLength({ min: 2, max: 255 })
+      .withMessage('Name must be between 2 and 255 characters'),
```
a6software, line 13: I can't see these 3 `withMessage` entries on the error document or the ticket - not saying they are wrong, but if you haven't already done so, I'd get Marc or Daisy to confirm they are exactly what they want.

a6software (changes requested):
Just a couple of minor things - top work :+1:

forms-web-app/src/views/appellant-submission/your-details.njk
Fenisay: No you're right, better without

forms-web-app/src/validators/appellant-submission/applicant-name.js
Fenisay: As we said I didn't checked the document at the moment, so by default I reused the existing rule for the name, which might already be wrong so I guess we should fix these message once and for all with a specific PR.

a6software (approved):
LGTM

a6software:
Ahh squash commits before merging, or Simon will hunt you down

PR #141, opened 2020-11-27
refactor(forms-web-app): add nesting to VIEW object to avoid naming c…

a6software:
…onflicts

## Ticket Number

UCD-892

## Description of change

Not absolutely required, but avoids long VIEW names to mitigate naming conflicts for forms that share the same name at different stages of the journey.

## Checklist

- [ ] New features have tests

## Important


forms-web-app/src/controllers/submission.js
```diff
@@ -21,5 +21,5 @@ exports.postSubmission = async (req, res) => {
     return;
```
Fenisay, line 21: To be sure that I understood the rule. All the view at root folder should have their own folder with index ?

Fenisay (dismissed):

forms-web-app/src/controllers/submission.js
a6software: It's a good Q. The way it's set up at the moment, this would be `submission/index` which breaks this convention. 

Maybe the view should just be `views/submission.njk` and then it follows the convention. 

What do you think?

a6software: ` res.redirect(VIEWS.SUBMISSION');`

would resolve to

 `res.redirect('/submission/index');`

which wouldn't work (afaik)

a6software: I've pushed up a second commit that extracts this to top level views - no more sub dirs for top level

Fenisay: Well il would be my first choice, feels more natural to me but I didn't knew if there was a reason to this. I already moved stuff from the root to sub folder eligibility or appellant-submission so I didn't want to do a massive refactoring and been blocked in the PR lol.

Fenisay (approved):

Fenisay (approved):

PR #150, opened 2020-11-30
test(e2e-tests): updated tests to use 50MB file size limit for uploads

frank-dodion:
## Ticket Number
UCD-1107

## Description of change
test(e2e-tests): updated tests to use 50MB file size limit for uploads
This affected appeal statement and decision letter file uploads.
Also fixed typo in checking sensitive information checkbox for appeal statement.

### DIFF
--- a/e2e-tests/create-large-test-files.sh
+++ b/e2e-tests/create-large-test-files.sh
@@ -1,4 +1,11 @@
-dd if=/dev/zero of=cypress/fixtures/appeal-statement-valid-max-size.png bs=104857600 count=1
-dd if=/dev/zero of=cypress/fixtures/appeal-statement-invalid-too-big.png bs=104857601 count=1
-dd if=/dev/zero of=cypress/fixtures/upload-file-valid-max-size.png bs=104857600 count=1
-dd if=/dev/zero of=cypress/fixtures/upload-file-invalid-too-big.png bs=104857601 count=1
+if [[ -z "${FILE_UPLOAD_MAX_FILE_SIZE_BYTES}" ]]; then
+  MAX_SIZE=52428800
+else
+  MAX_SIZE=$FILE_UPLOAD_MAX_FILE_SIZE_BYTES
+fi
+TOO_BIG=$((MAX_SIZE+1))
+dd if=/dev/zero of=cypress/fixtures/appeal-statement-valid-max-size.png bs=$MAX_SIZE count=1
+dd if=/dev/zero of=cypress/fixtures/appeal-statement-invalid-too-big.png bs=$TOO_BIG count=1
+dd if=/dev/zero of=cypress/fixtures/upload-file-valid-max-size.png bs=$MAX_SIZE count=1
+dd if=/dev/zero of=cypress/fixtures/upload-file-invalid-too-big.png bs=$TOO_BIG count=1
+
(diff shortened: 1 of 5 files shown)


e2e-tests/create-large-test-files.sh
```diff
@@ -1,4 +1,4 @@
-dd if=/dev/zero of=cypress/fixtures/appeal-statement-valid-max-size.png bs=104857600 count=1
-dd if=/dev/zero of=cypress/fixtures/appeal-statement-invalid-too-big.png bs=104857601 count=1
-dd if=/dev/zero of=cypress/fixtures/upload-file-valid-max-size.png bs=104857600 count=1
-dd if=/dev/zero of=cypress/fixtures/upload-file-invalid-too-big.png bs=104857601 count=1
+dd if=/dev/zero of=cypress/fixtures/appeal-statement-valid-max-size.png bs=52428800 count=1
```
MrSimonEmms, line 1: Can we put `$FILE_UPLOAD_MAX_FILE_SIZE_BYTES` so it's driven by the envvars? The `+1` is slightly more complex, but it means it's config driven

MrSimonEmms (changes requested):

e2e-tests/create-large-test-files.sh
frank-dodion: Good point. I will make it conditional so that if the env var is available then it will be used but if not a default value of 50Mb will be used.

MrSimonEmms: That's seems fair enough. From looking how it's invoked, the `.github/workflows/e2e-tests-on-PR.yml` and `.github/workflows/publish-cucumber-on-merge.yml` files may need updating to add the `FILE_UPLOAD_MAX_FILE_SIZE_BYTES` envvar.

Personally, so long as the test is in and works, the actual size is irrelevant to the configuration.

This PR won't actually change the max file size in production. That's controlled from `charts/app/values.yaml` (the `documentServiceApi.config.upload.maxSize` value).

frank-dodion: All sounds good. I have implemented the change.

MrSimonEmms:
Is this a change that's needed in production too? If so, can you update the Helm file `charts/app/values.yaml` (the documentServiceApi.config.upload.maxSize value)?

frank-dodion:
> Is this a change that's needed in production too? If so, can you update the Helm file `charts/app/values.yaml` (the documentServiceApi.config.upload.maxSize value)?

I don't think so. The script generates large files that are used for testing. It now uses the config value to set the size or 50MB if none is set. It does not modify the app's behaviour.

MrSimonEmms (approved):
Ok, if we're happy with the deployed app having a 100MB upload limit I'll approve

frank-dodion:
> Ok, if we're happy with the deployed app having a 100MB upload limit I'll approve

I am happy about this code change but I am not sure why we would have a limit set in a deployment that does not match the client's requirement.

MrSimonEmms:
> I am happy about this code change but I am not sure why we would have a limit set in a deployment that does not match the client's requirement.

Easy - there were no requirements when I did the deployment config. If you want to change the deployment file limit, can you raise a PR with the change to the Helm file I've mentioned above?

PR #152, opened 2020-11-30
feat: candidate description of planning application number feature

danWatkinson:
## Ticket Number
UCD-1040 ( UCD-892 )

## Description of change
Candidate feature file describing planning application number 

## Checklist

- [ x ] New features have tests

## Important

### DIFF
--- a/e2e-tests/cypress/integration/appellant-submission-planning-application-number.feature
+++ b/e2e-tests/cypress/integration/appellant-submission-planning-application-number.feature
@@ -1,0 +1,22 @@
+@wip
+Feature: Appellant submits a planning application reference number so that the planning inspectorate can refer to their appeal
+
+  Scenario Outline: Prospective appellant provides a valid planning application number
+    When the user provides a planning application number <valid application number>
+    Then the appeal is updated with the provided planning application number
+
+      Examples:
+          | valid application number |
+          | "P20/21076/NMA"          |
+          | "2020/1660/01/TCA"       |
+          | "118924/JO/2018"         |
+          | "P/3423/20"              |
+
+  Scenario Outline: Prospective appellant provides an invalid planning application number
+    When the user provides a planning application number <invalid application number>
+    Then the user is informed that the application number is not valid because <reason>
+    And the appeal is not updated with the provided planning application number
+
+      Examples:
+          | invalid application number   | reason            |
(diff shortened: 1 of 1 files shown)


e2e-tests/cypress/integration/appellant-submission-planning-application-number.feature
```diff
@@ -0,0 +1,22 @@
...
+          | "118924/JO/2018"         |
+          | "P/3423/20"              |
+
+  Scenario Outline: Prospective appellant provides an invalid planning application number
+    When the user provides a planning application number <invalid application number>
+    Then the user is informed that the application number is not valid because <reason>
+
+      Examples:
+          | invalid application number         | reason                                                    |
+          | "12345678901234567890123456789012" | "application numbers cannot have more than 30 characters" |
```
a6software, line 22: Not that I disagree with this, but this is not on the agreed error messages document. 

> Planning application number  
> Not entered: Enter the planning application number 

is all they have re: error messages for this screen. 

Have they agreed this anywhere?

danWatkinson: so the thinking here is - the feature file is describing things logically, and it should be possible to do that without knowing the actual error message.

in the step-definitions this will create the need for a mapping operation of "our logical description of the error" -> "the actual error text".

appreciate this initially looks a bit convoluted, this approach lets us pin down what we think should happen logically, without getting too bogged down in the exact wording of a message. Acknowledged it leaves that dictionary-lookup piece to do..

danWatkinson: NB ^ is not the same as saying "cool, we don't have that message defined anywhere.. who cares?" it's just saying that when we go into planning/estimation sessions, we should have things clearly laid out so that it's obvious that we need to know x/y/z error conditions, and we don't accept a ticket into the sprint until we have that map of "description of why we reject"->"error message" in the ticket.

a6software: It would appear they don't want this in - https://pins-ds.atlassian.net/browse/UCD-892?focusedCommentId=39259

frank-dodion (approved):
This is good to go.

PR #162, opened 2020-12-01
fix(security): allow unsafe-inline on script-src to allow the accessibleAutocomplete

MrSimonEmms:
## Ticket Number

UCD-1139

## Description of change

This drops the SecurityHeaders.com down to A from A+ which is still acceptable. A future piece of work should be done to put the accessibleAutocomplete inside a JS file that can be loaded from a URL or secured in [some other way](https://content-security-policy.com/examples/allow-inline-script/)

## Checklist

- [ ] New features have tests

## Important


karunak21 (approved):
I have checked the planning department page and it works as expected.

PR #164, opened 2020-12-01
feat(appeals-service-api): add openapi.yaml file

frank-dodion:
## Ticket Number
UCD-1108

## Description of change
Add openapi.yaml file


appeals-service-api/api/openapi.yaml
```diff
@@ -0,0 +1,314 @@
...
+        decisionLetter:
+          $ref: '#/components/schemas/DecisionLetter'
+      description: Section that includes required documents and reference number relating to original apllication
+    SectionState:
+      type: string
+      enum:
+        - 'NOT STARTED'
+        - 'IN PROGRESS'
+        - 'COMPLETED'
+      description: Each subsection begins as NOT STARTED and can progress to IN PROGESS and COMPLETED
```
Fenisay, line 219: typo IN PROGESS

Fenisay:
I have 2 remarks 
1. I don't know if we should align the model (sections) with the one of the site because it doesn't seem immutable to me. What's in it for us ?
2. I am torn between storing the states of the sections rather than recalculating them directly from the present data. For the moment, I don't think the performance gains are justified and this would force me to manage the synchronization so that I never find myself in the state where a complete object is in progress. On the other hand, it would suffice for a status to appear that is much more complicated to determine than the others, and this could make the dynamic solution shaky.

appeals-service-api/api/openapi.yaml
frank-dodion: Thanks. I'll fix.

frank-dodion:
> I have 2 remarks
> 
> 1. I don't know if we should align the model (sections) with the one of the site because it doesn't seem immutable to me. What's in it for us ?
> 2. I am torn between storing the states of the sections rather than recalculating them directly from the present data. For the moment, I don't think the performance gains are justified and this would force me to manage the synchronization so that I never find myself in the state where a complete object is in progress. On the other hand, it would suffice for a status to appear that is much more complicated to determine than the others, and this could make the dynamic solution shaky.

In answer to pint 1. I suppose aligning the api model of an Appeal to follow the logical structure of our web app including the current state of each subsection simply makes the response easier to use. We could adopt a different structure provided we have all of the required information but I am not how that helps. Only the lpdCode and decisionDate used in the initial POST to create the Appeal are considered immutable. I am not sure I understand the issue at point 2. Perhaps we can chat - I'll call. I think with the available endpoints we can update or retrieve the state of an Appeal at any time.

MrSimonEmms (approved):
I really like this. I think there's a good argument to create a web service to run this in our local dev stack so that we can test against it. I'd be happy for this to go in and have that as a future piece of work though.

I'm not sure who should make the final decision on when this goes in, but I'm giving this a big thumbs up

frank-dodion:
> I really like this. I think there's a good argument to create a web service to run this in our local dev stack so that we can test against it. I'd be happy for this to go in and have that as a future piece of work though.
> 
> I'm not sure who should make the final decision on when this goes in, but I'm giving this a big thumbs up

I think we will continue to serve the swagger at the /api-docs endpoint of this service so it can be used for basic manual testing. 

I hope to get this in today so it will drive deliver for this sprint.

MrSimonEmms:
That seems reasonable. Are you going to update this PR so it is served from there, or do that as a future task?

MrSimonEmms:
We'll also needs the `GET:/local-planning-authorities` and `GET:/local-planning-authorities/:id` endpoints adding into your Swagger docs

frank-dodion:
> We'll also needs the `GET:/local-planning-authorities` and `GET:/local-planning-authorities/:id` endpoints adding into your Swagger docs

Yes. Small steps.... We can continue to consume the existing local planning endpoint I think.

frank-dodion:
> > I have 2 remarks
> > 
> > 1. I don't know if we should align the model (sections) with the one of the site because it doesn't seem immutable to me. What's in it for us ?
> > 2. I am torn between storing the states of the sections rather than recalculating them directly from the present data. For the moment, I don't think the performance gains are justified and this would force me to manage the synchronization so that I never find myself in the state where a complete object is in progress. On the other hand, it would suffice for a status to appear that is much more complicated to determine than the others, and this could make the dynamic solution shaky.
> 
> In answer to pint 1. I suppose aligning the api model of an Appeal to follow the logical structure of our web app including the current state of each subsection simply makes the response easier to use. We could adopt a different structure provided we have all of the required information but I am not how that helps. Only the lpdCode and decisionDate used in the initial POST to create the Appeal are considered immutable. I am not sure I understand the issue at point 2. Perhaps we can chat - I'll call. I think with the available endpoints we can update or retrieve the state of an Appeal at any time.

I discussed point 2 with @Fenisay and clarified that this api spec makes no binding constraint on how an Appeal might be represented internally in the UI or the DB.

MrSimonEmms:
> > We'll also needs the `GET:/local-planning-authorities` and `GET:/local-planning-authorities/:id` endpoints adding into your Swagger docs
> 
> Yes. Small steps.... We can continue to consume the existing local planning endpoint I think.

Yeah, I agree on small steps. It's a thing that needs doing, but can be done afterwards. Baby/bathwater and all that jazz

PR #167, opened 2020-12-03
feat: candidate Feature File for Task List

karunak21:
## Ticket Number
UCD-766

## Description of change
Candidate Feature file for Your Details section in Task List page

## Checklist

- [ X] New features have tests

## Important

### DIFF
--- a/e2e-tests/cypress/integration/appeal-submission-tasklist.feature
+++ b/e2e-tests/cypress/integration/appeal-submission-tasklist.feature
@@ -1,0 +1,28 @@
+@wip
+Feature: Your Details status
+
+    Scenario Outline: The status of the Your Details section is accurate
+        Given the user has provided their name: <name> and email: <email>
+        And the user has stated that they "are" the original appellant
+        When the user checks the status of their appeal
+        Then the user should see that the Your Details section is <status>
+
+        Examples:
+            | name        | email            | Status        |
+            | ""          | ""               | "Not Started" |
+            | ""          | "good@email.com" | "Not Started" |
+            | "Good Name" | ""               | "Not Started" |
+            | "Good Name" | "good@email.com" | "Completed"   |
+
+
+      Scenario Outline: The status of the Your Details section is accurate
+          Given the user has provided their name: <name> and email: <email>
+          And the user has stated that they "are not" the original appellant
+          And the user has provided <on behalf of> as the name of the original appellant
+          When the user checks the status of their appeal
+          Then the user should see that Your Details section is <status>
+
+          Examples:
+              | name                | email            | on behalf of    | Status        |
+              | "Another Good Name" | "good@email.com" | ""              | "In Progress" |
+              | "Another Good Name" | "good@email.com" | "Good Name"     | "Completed"   |


karunak21:
We would like to reuse existing Gherkin sentences but those sentences need to be fixed to allow reuse. 
If we are happy with the feature sentences then @danWatkinson  has offered help to work on the existing sentences.

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
```diff
@@ -0,0 +1,22 @@
...
+        # correct this. update who=are=you feature to accept "yes"/"no" or "is"/"isnt"
+        And the user states that they <original appelant> the original appellant
+        # correct this. update who-are-you feature to care about the original apelant name - this sentence should appear in that feature somewhere?
+        And the user provides <on behalf of> as the name of the original appellant
+        Then the user should see that Your Details section is <status>
+        
+        Examples:
+            | name                | email            | original apellant | on behalf of    | Status        |
+            | ""                  | ""               | "doesnt matter"   | "doesnt matter" | "Not Started" |
+            | ""                  | "good@email.com" | "doesnt matter"   | "doesnt matter" | "Not Started" |
```
frank-dodion, line 16: Will it be possible to get into this state where there is no name but there is an email? I think if this is submitted it should fail validation and prevent any update.

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
```diff
@@ -0,0 +1,22 @@
+@wip
+Feature: Your Details status
+
+    Scenario Outline: The status of the Your Details section is accurate
+        # correct this, and correct your-details feature; 'valid' should not feature here; we want to be able to re-use for invalid cases too
```
frank-dodion, line 5: I am not sure we should have any comments in the feature file.

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
danWatkinson: that's the situation we're trying to communicate here - so if we give no name + an email we should fail validation so the status should be 'not started'.
totally aware there may be nicer ways to word things but i was trying to show that in a situation where we've given something invalid:  it doesn't matter what we do for the other fields (maybe "" is clearer there?)

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
danWatkinson: we 100% shouldn't
however -  we're proposing that we'd like to re-use some existing sentences but the existing sentences aren't quite the right shape... if we're happy with the proposed syntax I'll pick this up and go re-work the existing sentences+glue so we get the re-use.. i just can't be bothered to do that until we know the new structure is 👍

frank-dodion: Np. thanks

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
frank-dodion: Ok.

frank-dodion (approved):

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
```diff
@@ -0,0 +1,28 @@
...
+
+        Examples:
+            | name        | email            | Status        |
+            | ""          | ""               | "Not Started" |
+            | ""          | "good@email.com" | "Not Started" |
+            | "Good Name" | ""               | "Not Started" |
+            | "Good Name" | "good@email.com" | "Completed"   |
+
+
+      Scenario Outline: The status of the Your Details section is accurate
```
frank-dodion, line 18: I think we could use a different wording in the title to show it is different from the scenario above i.e. in this case "... when the user is not the original applicant"

frank-dodion (approved):
We could refactor scenario title later but getting this in now will allow work on the steps to continue.

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
karunak21: Will offer tidy up in glue code

karunak21: This will be addressed while implementing the glue code

PR #173, opened 2020-12-04
Configure Pre Prod environment in PINS Azure subscription

MrSimonEmms:
## Ticket Number

UCD-1174

## Description of change

Configure the new PINS Azure subscription and create the PreProd environment in there.
 - Upgrade Terraform to v0.14.0.
 - improve the Cosmos DB connection. Remove the creation of databases inside Terraform and let the individual services create their own databases (as is the Mongo way) - Azure confirmed this would replicate correctly in my call with them last week.
 - ~Apply partial-fix for [issue](https://github.com/mongodb-js/connect-mongodb-session/issues/90) in Connect MongoDB Session. Fixes error in Cosmos when creating a time-based expiry index. This removes the expiry index - I will raise a PR and fix independently (and apply patch to the repo when done if not applied in a timely manner).~ Applied the fix that I've [raised a PR for](https://github.com/mongodb-js/connect-mongodb-session/issues/91).
 - remove empty nginx-ingress config files - `staging` and `prod` are empty so not required in the repo.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/.github/workflows/infra-common.yml
+++ b/.github/workflows/infra-common.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@v1
         with:
-          terraform_version: 0.13.4
+          terraform_version: 0.14.0
           terraform_wrapper: false
 
       - name: Terraform Init
@@ -63,7 +63,7 @@ jobs:
         run: bash ../scripts/lockResourceGroups.sh
           unlock
           "$(terraform workspace show)"
-          "$(terraform output app-name)"
+          "$(terraform output -json app-name | jq -r .)"
 
       - run: terraform apply tfplan
 
@@ -72,4 +72,4 @@ jobs:
         run: bash ../scripts/lockResourceGroups.sh
           lock
           "$(terraform workspace show)"
-          "$(terraform output app-name)"
+          "$(terraform output -json app-name | jq -r .)"
--- a/charts/app/values.yaml
+++ b/charts/app/values.yaml
@@ -113,7 +113,5 @@ keyVault:
   secrets:
     - docs-blob-storage-connection-string
     - fwa-session-key
-  multiValueSecrets:
-    - mongodb-appeals-store
-    - mongodb-documents-store
-    - mongodb-sessions-store
+    - mongodb-connection-url
+  multiValueSecrets: []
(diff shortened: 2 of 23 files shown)


MrSimonEmms:
**IMPORTANT** Before this is merged in, the secrets in GitHub Actions must be updated with the values for the new subscription

Fenisay (approved):

forms-web-app/patches/connect-mongodb-session+2.4.1.patch
```diff
@@ -0,0 +1,48 @@
+diff --git a/node_modules/connect-mongodb-session/index.js b/node_modules/connect-mongodb-session/index.js
```
MrSimonEmms, line 1: FYI - this is an auto-generated file

charts/app/values.yaml
```diff
@@ -34,7 +34,7 @@ formsWebApp:
     name: forms-web-app
     port: 9000
   image:
-    repository: pinscommonukscontainersdefault.azurecr.io/forms-web-app
+    repository: pinscommonukscontainers3887default.azurecr.io/forms-web-app
```
a6software, line 37: curious - why is this numbered now? I see something below:

` name = replace(format(local.name_format, "containers-${random_integer.containers.result}"), "-", "")`

Is it ever possible these get out of sync? 

idk, I may be barking up the wrong bush - curiosity only here

.github/workflows/infra-common.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         run: bash ../scripts/lockResourceGroups.sh
           unlock
           "$(terraform workspace show)"
-          "$(terraform output app-name)"
+          "$(terraform output -json app-name | jq -r .)"
```
a6software, line 66: I'm working on the assumption that `jq` is a guaranteed pre-installed dependency on the github image that runs these commands?

a6software (approved):

charts/app/values.yaml
MrSimonEmms: Splendid question. The container registry name must be globally unique. As `pinscommonukscontainersdefault` is the name in the F4 subscription, I had to change it to create a separate one in the PINS subscription (I wanted to get it up and running beforehand so I could test it). Even though this will be unique once I drop the F4 one, I've used a random integer in everything that requires a globally unique name.

I don't believe there's any chance of it getting out of sync as the registry is designed to stay up forever - if we drop it then we lose all the container images in the registry.

.github/workflows/infra-common.yml
MrSimonEmms: Yes, it appears to be in the container images we're using.

PR #179, opened 2020-12-07
ci: we should run some percentage of our e2e tests as a smoke test wh…

danWatkinson:
…en we deploy

## Ticket Number
UCD-1191

## Description of change

`BASEURL=https://appeals-dev.planninginspectorate.gov.uk/ npm run test:e2e:smoke`

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "test:e2e:demo": "cd ./e2e-tests && npm run test:e2e:demo",
     "test:e2e:postprocess": "cd ./e2e-tests && npm run test:e2e:postprocess",
     "test:e2e:files": "cd ./e2e-tests && ./create-large-test-files.sh",
+    "test:e2e:smoke": "cd ./e2e-tests && BASEURL=$BASEURL npm run test:e2e:smoke",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "repository": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
     "test:e2e:demo": "cd ./e2e-tests && npm run test:e2e:demo",
     "test:e2e:postprocess": "cd ./e2e-tests && npm run test:e2e:postprocess",
     "test:e2e:files": "cd ./e2e-tests && ./create-large-test-files.sh",
+    "test:e2e:smoke": "cd ./e2e-tests && BASEURL=$BASEURL npm run test:e2e:smoke",
```
MrSimonEmms, line 16: Is `BASEURL=$BASEURL` necessary to get it to run on Windows? Shouldn't be necessary on Mac and isn't needed on Linux

MrSimonEmms:
I'm going to create a spike off this PR to investigate how to get this running in CI/CD and how to get it running in k8s as part of the deployment process

package.json
danWatkinson: that's the variable i'm proposing we use to point this at different environments.
the "OOB" behaviour of running the e2e tests is that it has the baseurl set to localhost:9000; this is exposing some variable that we can use to point everything at 'something else'.

the assumption is that you'll have a step that does a deploy, then a step that does a `BASEURL={url we just deployed to} npm run blah`

danWatkinson: sorry, realised the confusion.
this is with the assumption that you're going to pass that value in when you call things, so in root:
`BASEURL=https://appeals-dev.planninginspectorate.gov.uk/ npm run test:e2e:smoke`
and the entry here is just passing that config up the chain

MrSimonEmms:
UCD-1192

MrSimonEmms (approved):

PR #185, opened 2020-12-09
feat(appeals-service-api): implements Appeals API

frank-dodion:
BREAKING CHANGE: Replaces previous endpoints.

## Ticket Number
UCD-985

## Description of change
Implements Appeals API


danWatkinson (approved):
assuming we have a ticket in this story that gets the feature files un-@wip'd all good.. assuming we'd rather get this change in than haggle about a short-term step back in 'provable functionality'

frank-dodion:
> assuming we have a ticket in this story that gets the feature files un-@wip'd all good.. assuming we'd rather get this change in than haggle about a short-term step back in 'provable functionality'

I think it will be for whoever claims to have implemented the feature to prove this by removing the @wip tag.

MrSimonEmms:
Having the `BREAKING CHANGE` will mean this goes to v2. Not that that's a bad thing, but please let me know when this has gone out so I can update the release manifest

danWatkinson:
> Having the `BREAKING CHANGE` will mean this goes to v2. Not that that's a bad thing, but please let me know when this has gone out so I can update the release manifest

Assuming I understand the change, the 'breaking change' is reflecting the fact that tests that used to pass will now not pass.. that's kinda handled by flagging the tests accordingly. Whether we actually want/need to bump version numbers to reflect this? I think you could call it either way..

frank-dodion:
> Having the `BREAKING CHANGE` will mean this goes to v2. Not that that's a bad thing, but please let me know when this has gone out so I can update the release manifest

We have not been using v1 yet so I think this will be the first 'v1'.

frank-dodion:
> > Having the `BREAKING CHANGE` will mean this goes to v2. Not that that's a bad thing, but please let me know when this has gone out so I can update the release manifest
> 
> Assuming I understand the change, the 'breaking change' is reflecting the fact that tests that used to pass will now not pass.. that's kinda handled by flagging the tests accordingly. Whether we actually want/need to bump version numbers to reflect this? I think you could call it either way..



> > Having the `BREAKING CHANGE` will mean this goes to v2. Not that that's a bad thing, but please let me know when this has gone out so I can update the release manifest
> 
> Assuming I understand the change, the 'breaking change' is reflecting the fact that tests that used to pass will now not pass.. that's kinda handled by flagging the tests accordingly. Whether we actually want/need to bump version numbers to reflect this? I think you could call it either way..

I understand that in this context it is a breaking change because previously available/described endpoints have been removed.

MrSimonEmms:
@frank-dodion the v2 is on the release number (which is manifest as the Docker image tag) not on the URL endpoint. So this will be released as `2.0.0`

frank-dodion:
> @frank-dodion the v2 is on the release number (which is manifest as the Docker image tag) not on the URL endpoint. So this will be released as `2.0.0`
> 
> When we move this over to the PINS ownership, I want to make changes to the tag numbering which will reset everything back to `1.0.0` - that's on me and this won't affect anything

Ah. I have no opinion about docker tags.

Fenisay (approved):

rcmsoftwaredevelopment (approved):

Fenisay:
FYI there is also the status CANNOT BE STARTED, even if it's currently not used with the current screens

frank-dodion:
> FYI there is also the status CANNOT BE STARTED, even if it's currently not used with the current screens

Is that for the state of a section? I think that can be added later if we don't use it now.

appeals-service-api/src/__tests__/appeals.spec.js
```diff
@@ -0,0 +1,220 @@
...
+
+  afterAll(async () => {
+    await connection.close();
+    await db.close();
+  });
+
+  test('POST /api/v1/appeals - It responds with a newly created appeal', async () => {
+    const appeal = appealDocument;
+    const response = await request(app).post('/api/v1/appeals').send({
+      // lpaCode: 'E60000281',
```
frank-dodion, line 43: Oh dear. I missed this unwanted comment. It does not affect behaviour so I will remove on next PR.

PR #192, opened 2020-12-10
Fix/ucd 754 navigation

Fenisay:
Put official css file, as I finally managed to only use mix of existing css (a lot of fun) we don't need this one.
This one has been excluded in production so it's simpler to do this way especially that I hope the original gov component should be updated at some point.

UCD-754 refactoring


appeals-service-api/src/__tests__/appeals.spec.js
```diff
@@ -61,7 +58,7 @@ describe('Appeals API', () => {
     expect(response.statusCode).toBe(404);
   });
 
-  test('PUT /api/v1/appeals/{id} - It responds with an updated appeal', async () => {
+  test.skip('PUT /api/v1/appeals/{id} - It responds with an updated appeal', async () => {
```
Fenisay, line 61: @frank-dodion this test is failing but it seems that it's doing what it's supposed to do. Update based with the change made by the service. Can you confirm please ?

Capture d’écran 2020-12-10 à 14 57 08

frank-dodion: Let me see..

frank-dodion: It seems to pass for me locally. I typically use npm run test:tdd from the appeals-service-api folder to get rapid feeddback.

frank-dodion: Is it failing in the ci checks?

appeals-service-api/src/__tests__/appeals.spec.js
```diff
@@ -39,10 +39,7 @@ describe('Appeals API', () => {
 
   test('POST /api/v1/appeals - It responds with a newly created appeal', async () => {
     const appeal = appealDocument;
-    const response = await request(app).post('/api/v1/appeals').send({
-      // lpaCode: 'E60000281',
-      // decisionDate,
-    });
+    const response = await request(app).post('/api/v1/appeals').send(appealDocument);
```
frank-dodion, line 42: I think the current appeals api spec does not expect a payload.

appeals-service-api/src/controllers/appeals.js
```diff
@@ -2,12 +2,11 @@ const uuid = require('uuid');
 const logger = require('../lib/logger');
 
 const mongodb = require('../db/db');
-const { appealDocument } = require('../models/appeal');
 const { validateAppeal } = require('../middleware/validateAppeal');
 
 module.exports = {
   async create(req, res) {
-    const appeal = appealDocument;
+    const appeal = req.body;
```
frank-dodion, line 9: I think the current appeals api spec does not expect a payload. Here we use a known to be 'safe' state for the newly created appeal.

appeals-service-api/src/__tests__/appeals.spec.js
Fenisay: Not sure as I fixed something else, I push without the skip to see. Thanks

Capture d’écran 2020-12-10 à 15 25 16

It's this part which somehow change the date on my laptop and top yours

appeals-service-api/src/controllers/appeals.js
Fenisay: The thing is that on the front I need to initialise a new appeal before, the only difference is that he has no id.
And it's weird to have to ask to create an empty skeleton and then update it. All the first task you do will be replaced with the skeleton without this change.

appeals-service-api/src/__tests__/appeals.spec.js
Fenisay: The ci passed, that's weird.

frank-dodion: For most of the tests we generate a new appeal and set the decision date to yesterday. If could be any past date but I wanted it to stay in the valid period so it could not be a fixed date. I suppose it could be a time zone thing because the essentially we strip out the first ten characters to get a string like this YYYY-MM-DD from an ISO date string.

appeals-service-api/src/controllers/appeals.js
frank-dodion: I think the way to initialise a new appeal at any point is to call the POST (create) endpoint with no payload. The response will be a safe appeal json representation of an appeal with a new id. You can build and submit any PUT updates using that as a template payload.

Fenisay: It was done like this because the previous model was just flat and updated on the fly. Now we have a new data model, why not change ? 

Is there a reason why we would not want to offer the capacity to insert an appeal with the body on the payload ? It seems to be the standard for POST method in REST architecture. I don't really like the 'transactional' way that you describing and the safety you mentioned should not be the issue as we are supposed to restrict the API soon and if the body in the payload is not safe, he will not be stored. 

Did I understood correctly ?

frank-dodion (changes requested):
The proposed changes will mean that calling the POST (create) endpoint does not comply with the api spec. More importantly, the payload accepted is not subject to validation so it could result in a new appeal that is invalid.

Fenisay:
> The proposed changes will mean that calling the POST (create) endpoint does not comply with the api spec. More importantly, the payload accepted is not subject to validation so it could result in a new appeal that is invalid.

I answered you on the other post.


It was done like this because the previous model was just flat and updated on the fly. Now we have a new data model, why not change the API ?

Is there a reason why we would not want to offer the capacity to insert an appeal with the body on the payload ? It seems to be the standard for POST method in REST architecture. I don't really like the 'transactional' way that you describing and the safety you mentioned should not be the issue as we are supposed to restrict the API soon and if the body in the payload is not safe, he will not be stored. 

Did I understood correctly ?

frank-dodion:
> > The proposed changes will mean that calling the POST (create) endpoint does not comply with the api spec. More importantly, the payload accepted is not subject to validation so it could result in a new appeal that is invalid.
> 
> I answered you on the other post.
> 
> It was done like this because the previous model was just flat and updated on the fly. Now we have a new data model, why not change the API ?
> 
> Is there a reason why we would not want to offer the capacity to insert an appeal with the body on the payload ? It seems to be the standard for POST method in REST architecture. I don't really like the 'transactional' way that you describing and the safety you mentioned should not be the issue as we are supposed to restrict the API soon and if the body in the payload is not safe, he will not be stored.
> 
> Did I understood correctly ?

It is always possible to propose a change to the API and provide an implementation but that is not in this PR.

Fenisay:
> > > The proposed changes will mean that calling the POST (create) endpoint does not comply with the api spec. More importantly, the payload accepted is not subject to validation so it could result in a new appeal that is invalid.
> > 
> > 
> > I answered you on the other post.
> > It was done like this because the previous model was just flat and updated on the fly. Now we have a new data model, why not change the API ?
> > Is there a reason why we would not want to offer the capacity to insert an appeal with the body on the payload ? It seems to be the standard for POST method in REST architecture. I don't really like the 'transactional' way that you describing and the safety you mentioned should not be the issue as we are supposed to restrict the API soon and if the body in the payload is not safe, he will not be stored.
> > Did I understood correctly ?
> 
> It is always possible to propose a change to the API and provide an implementation but that is not in this PR.

You can tell me if you think that I want and said makes sense to you first please, then I won't need to change the API for no reason ?

PR #201, opened 2020-12-11
Feature/ucd 1079

MrSimonEmms:
## Ticket Number

UCD-1079

## Description of change

Create the Document Service upload endpoints and get a file uploaded.

## Checklist

- [x] New features have tests

## Important


docker-compose.yml
```diff
@@ -49,6 +49,8 @@ services:
     image: node:14-alpine
     environment:
       BLOB_STORAGE_CONNECTION_STRING: DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://blob-storage:10000/devstoreaccount1;QueueEndpoint=http://blob-storage:10001/devstoreaccount1;
+      FILE_MAX_SIZE_IN_BYTES: 50000000
+      FILE_UPLOAD_PATH: /tmp/upload
```
GoEagles999, line 53: Does this folder need to be created?

document-service-api/src/lib/config.js
```diff
@@ -19,6 +21,23 @@ module.exports = {
...
+      path: process.env.DOCS_API_PATH || path.join(__dirname, '..', 'api'),
+    },
+  },
+  fileUpload: {
+    maxSizeInBytes: Number(process.env.FILE_MAX_SIZE_IN_BYTES || 1000),
+    mimeTypes: [
+      'application/pdf', // pdf
+      'application/msword', // doc
+      'application/vnd.openxmlformats-officedocument.wordprocessingml.document', // docx
+      'image/tiff', // tiff
```
GoEagles999, line 35: could be image/tif as well

document-service-api/src/lib/config.js
```diff
@@ -19,6 +21,23 @@ module.exports = {
...
+      'image/jpeg', // jpeg
```
GoEagles999, line 36: could be image/jpg as well

MrSimonEmms: Seemingly not https://stackoverflow.com/questions/33692835/is-the-mime-type-image-jpg-the-same-as-image-jpeg

GoEagles999 (dismissed):

document-service-api/src/lib/config.js
MrSimonEmms: As with jpg, seemingly not https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

MrSimonEmms:
Need to create a sidecar to do the blob upload

docker-compose.yml
```diff
@@ -45,27 +45,40 @@ services:
       - ./appeals-service-api:/opt/app
     command: npm run start:dev
 
-  document-service-api:
+  document-service-api: &document-service
     image: node:14-alpine
     environment:
       BLOB_STORAGE_CONNECTION_STRING: DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://blob-storage:10000/devstoreaccount1;QueueEndpoint=http://blob-storage:10001/devstoreaccount1;
+      STORAGE_CONTAINER_NAME: document-service-uploads
+      FILE_MAX_SIZE_IN_BYTES: 50000000
```
a6software, line 53: Have I made a balls up then - https://github.com/foundry4/appeal-planning-decision/blob/master/docker-compose.yml#L75

think I need to change this to `50000000` also ?

charts/app/templates/documentServiceApi.yaml
```diff
@@ -1,3 +1,5 @@
+---
+# Deplyment - this is the main API application
```
a6software, line 2: nit: typo on Deplyment

a6software (approved):
Of what I understand, all LGTM :+1:

document-service-api/src/api/openapi.yaml
```diff
@@ -0,0 +1,205 @@
...
+          type: string
+          description: The mime type
+          example: application/pdf
+        size:
+          type: integer
+          description: The length of the file in bytes
+        location:
+          type: string
+          format: uri
+          description: The absolutely URI to the file
```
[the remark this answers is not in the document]
MrSimonEmms: In the original doc I received

charts/app/templates/documentServiceApi.yaml
MrSimonEmms: Nit indeed

rcmsoftwaredevelopment (dismissed):

docker-compose.yml
MrSimonEmms: Yes, and no. In dev, we don't care what the actual size should be. It's a development experience and has no bearing on what it is on the deployed application (and these could conceivably be different in each each environment.

The number you've done there is actually 50 mebibytes - data was decimalised about 10 years ago so a megabyte is 1000^2, but a mebibyte is 1024^2 https://en.wikipedia.org/wiki/Mebibyte

document-service-api/src/api/openapi.yaml
```diff
@@ -0,0 +1,205 @@
...
+openapi: 3.0.0
+info:
+  title: Appeals Document Service API Documentation
+  version: 1.0.0
+  license:
+    name: www.foundry4.com
+servers:
+  - url: http://localhost:3001/
+paths:
+  "/api/v1/{applicationId}":
```
frank-dodion, line 11: Can the applicationId contain '/' characters i.e. forward slashes?

MrSimonEmms: It should be a UUID. I guess is could contain slashes if it was URL encoded. This application just passes in what it receives for logical separation in the blob storage - there's no validation on the appId at all

GoEagles999:
> Need to create a sidecar to do the blob upload

Done?

GoEagles999:
Very solid piece of work mate

GoEagles999 (approved):
Production-grade code mate

MrSimonEmms:
> > Need to create a sidecar to do the blob upload
> 
> Done?

Yes. This is in the `document-service-api/src/tasks/uploadFilesToBlobStorage.js`

document-service-api/src/api/openapi.yaml
frank-dodion: So is this the Appeal Id rather than the original application id? I am not sure we will have access to the original application id as a uuid. I think if we have a value that could include forward slashes as a path variable then e may have problems.
Screenshot 2020-12-14 at 14 27 27

MrSimonEmms: My expectation is that it's the UUID that generated by the Forms Web App and stored in the session to link everything together.

frank-dodion: It seems to work in Postman so it might work from the client app that does not attempt to escape characters. 
Screenshot 2020-12-14 at 14 42 18

docker-compose.yml
MrSimonEmms: I've just read that article. I'm now upset that 1998 wasn't "about 10 years ago".

[I'm old](https://giphy.com/gifs/jamie-lee-curtis-boho-freaky-friday-TfseOfhwd6BJC)

document-service-api/src/api/openapi.yaml
MrSimonEmms: This is a non-issue. This should be the UUID generated by the application. That's just alphanumeric

frank-dodion: I suppose a deeper question is, "Does the Document Service need to know about the planning application id or appeal id at all?" If that information is provided as a linking key then we can retrieve all related docs together but in the context of our current project the list of document ids for a particular appeal is available in the appeal entity. We could acheive the same result by calling Document Service with a collection of document ids.

rcmsoftwaredevelopment (approved):
All fine

a6software (approved):

docker-compose.yml
```diff
@@ -45,27 +45,40 @@ services:
       - ./appeals-service-api:/opt/app
     command: npm run start:dev
 
-  document-service-api:
+  document-service-api: &document-service
```
Fenisay, line 48: Didn't knew about this feature :o

document-service-api/src/api/openapi.yaml
frank-dodion: Thanks for the clarification. Perhaps we could update the wording on the description and enforce that the path variable must be a uuid.

docker-compose.yml
MrSimonEmms: Called "YAML Anchors". Very useful https://blog.daemonl.com/2016/02/yaml.html

document-service-api/src/api/openapi.yaml
MrSimonEmms: That's exactly the point I made above. It **DOESN'T** have to be anything. There's zero validation on this, it's just used as a logical separator and used to group files. But the Doc Service has no knowledge or interest in what that value is.

This may require further work in the future on validation, but the purposes of MVP it's met the good-enough-is-good-enough stage

frank-dodion: If the applicationId is a uuid generated by FWA and stored in the session but not the appeal id (that is generated by the Appeals Service) then I don't think it will be persisted so may not be available when the session closes.

MrSimonEmms: It should be persisted to the Appeals Service document - I don't know about how you changed it when you did #185, but that was the purposes of the `uuid` in the original model. In any case, any documents added as part of an appeal will store the document ID (generated here as `id`, but also a UUID) and put those in the Appeals Service document.

frank-dodion: The appeal id is persisted as the primary key.

frank-dodion (approved):

document-service-api/src/api/openapi.yaml
MrSimonEmms: Yeah, I don't care how it's stored so long as there's a unique ID in the Appeals Service. This is what should be passed in here

PR #203, opened 2020-12-11
feat(appeals-service-app): appeal statement business validation

frank-dodion:
## Ticket Number
UCD-1213

## Description of change
implemented business logic validation for appeal statement file upload

### DIFF
--- a/appeals-service-api/src/__tests__/appeals.spec.js
+++ b/appeals-service-api/src/__tests__/appeals.spec.js
@@ -8,7 +8,7 @@ const { appealDocument } = require('../models/appeal');
 jest.mock('../db/db');
 
 async function createAppeal() {
-  const appeal = appealDocument;
+  const appeal = JSON.parse(JSON.stringify(appealDocument));
   appeal.id = uuid.v4();
   await mongodb.get().collection('appeals').insertOne({ _id: appeal.id, uuid: appeal.id, appeal });
   return appeal;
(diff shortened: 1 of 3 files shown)


rcmsoftwaredevelopment (approved):

appeals-service-api/src/__tests__/appeals.spec.js
```diff
@@ -8,7 +8,7 @@ const { appealDocument } = require('../models/appeal');
 jest.mock('../db/db');
 
 async function createAppeal() {
-  const appeal = appealDocument;
+  const appeal = JSON.parse(JSON.stringify(appealDocument));
```
MrSimonEmms, line 11: What's the story here? Surely this is the same as `appealDocument`

appeals-service-api/src/__tests__/appeals.spec.js
```diff
@@ -34,11 +34,8 @@ describe('Appeals API', () => {
   });
 
   test('POST /api/v1/appeals - It responds with a newly created appeal', async () => {
-    const appeal = appealDocument;
-    const response = await request(app).post('/api/v1/appeals').send({
-      // lpaCode: 'E60000281',
-      // decisionDate,
-    });
+    const appeal = JSON.parse(JSON.stringify(appealDocument));
```
MrSimonEmms, line 37: As above

appeals-service-api/src/controllers/appeals.js
```diff
@@ -7,7 +7,7 @@ const { validateAppeal } = require('../middleware/validateAppeal');
 
 module.exports = {
   async create(req, res) {
-    const appeal = appealDocument;
+    const appeal = JSON.parse(JSON.stringify(appealDocument));
```
MrSimonEmms, line 10: +1

MrSimonEmms (commented):
Confused by the `JSON.parse(JSON.stringify(obj))`

appeals-service-api/src/__tests__/appeals.spec.js
frank-dodion: I think we need a deep copy to avoid issues when tests etc are run in parallel.

MrSimonEmms (approved):
Ok. May be better ways of doing it, but can revisit it in the future (we probably never will)

PR #208, opened 2020-12-14
feat(appeals-service-api): add a service to insert already valued appeal

Fenisay:
UCD-797

**Add of a service to insert valued appeal.**
The first change of the appeal is currently erased as replaced by the empty appeal of appeal-service-api. So I could workaround the front or add a new service to insert a valued appeal. I chose the latter as that I'm pretty sure that the need to insert a valued appeal will be eventually necessary and it's more close to the standard REST architecture where POST is meant to insert and PUT to replace.

**Change from post to get the service to get a new a appeal.**
GET are not supposed to change the db. However this service is not used at the moment so I can remove it at the moment or put it back in post and add appeals/new endpoint.

I think I implemented the less polemic version but my personal choice would have been to test if the body is valued, if so insert it, otherwise create and return an empty appeal to follow the REST guidelines.

I'm waiting for any point of view


e2e-tests/cypress/integration/appellant-submission-planning-application-number.feature
```diff
@@ -9,7 +9,7 @@ Feature: Appellant submits a planning application reference number so that the p
       Examples:
           | valid application number |
           | "P20/21076/NMA"          |
-          | "2020/1660/01/TCA"       |
+#          | "2020/1660/01/TCA"   UCD-1227    |
           | "118924/JO/2018"         |
```
Fenisay, line 13: Discovered a preexisting bug thanks to backend checks
https://pins-ds.atlassian.net/browse/UCD-1227

danWatkinson: is this not saying "this test ran green, then we changed the plumbing, now it runs red"?

if so- on a normal day i'd say we should be fixing that as part of whatever work found it.. Given we're on sprint close day and need things over the line I'm happy enough to defend the idea of "and there is this one outstanding case we need to look at"...

Fenisay: I agree, the thing is that the test should be failing ages ago, it's not part of a recent change and even if the test was green the outcome was actually not better than now. I did try to fix it at the moment but the priority was more to focus to fix the master and finish 830. 

This being said, we went on a workaround for the issue that this PR was meant to fixed so it's less urgent and this particular bug has been covered since then so I'll rebase and this will be forgotten.

PR #199, opened 2020-12-11
docs: create the openapi spec for the document service

MrSimonEmms:
## Ticket Number

UCD-1079

## Description of change

Create the OpenAPI spec. Published on /api-docs and /api-docs/swagger.json (for the raw JSON)

## Checklist

- [ ] Requires infrastructure changes
- [ ] I have updated the documentation accordingly
- [ ] I have merged commits to avoid fixing things in this PR
- [ ] New features have tests
- [ ] Breaking change (team conversation required)

## Important

Please do not merge from `master` (please only [rebase](https://github.com/foundry4/appeal-planning-decision/wiki/An-intro-to-Git-Rebase)). This keeps the history linear and easier to debug.

PR #212, opened 2020-12-14
docs: create the openapi spec for the document service

MrSimonEmms:
## Ticket Number

UCD-1079

## Description of change

Create the OpenAPI spec. Published on /api-docs and /api-docs/swagger.json (for the raw JSON) (supersedes #199)

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/document-service-api/src/api/openapi.yaml
+++ b/document-service-api/src/api/openapi.yaml
@@ -1,0 +1,205 @@
+---
+openapi: 3.0.0
+info:
+  title: Appeals Document Service API Documentation
+  version: 1.0.0
+  license:
+    name: www.foundry4.com
+servers:
+  - url: http://localhost:3001/
+paths:
+  "/api/v1/{applicationId}":
+    description: Resources related to application specific documents
+    get:
+      tags:
+        - Document
+      operationId: getApplicationDocuments
+      description: Obtains all document data related to a given application ID
+      parameters:
+        - "$ref": "#/components/parameters/applicationId"
+      responses:
+        '200':
+          description: Successfully retrieved documents for the given Id
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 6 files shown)


document-service-api/src/api/openapi.yaml
```diff
@@ -0,0 +1,179 @@
...
+        content:
+          multipart/form-data:
+            schema:
+              type: object
+              properties:
+                file:
+                  type: string
+                  format: binary
+      responses:
+        '202':
```
frank-dodion, line 48: I think we need a content: attribute for the responses that do give json responses e.g.
Screenshot 2020-12-14 at 12 20 05

MrSimonEmms: Ok - speak to Lee as author of document

document-service-api/src/api/openapi.yaml
```diff
@@ -0,0 +1,205 @@
...
+openapi: 3.0.0
+info:
+  title: Appeals Document Service API Documentation
+  version: 1.0.0
+  license:
+    name: www.foundry4.com
+servers:
+  - url: http://localhost:3001/
+paths:
+  "/api/v1/{applicationId}":
```
frank-dodion, line 11: Should the path be something like /api/v1/documents...?

frank-dodion, line 11: Perhaps the applicationId could be provided(required) in the header because each calling app will have their own fixed value.

MrSimonEmms: No. This is the document service, so that's the root level

MrSimonEmms: I don't get it?

a6software (approved):
The Node stuff I can :+1: 

The swagger stuff seems good to me, but I'm not king of swagger

PR #228, opened 2020-12-16
fix(forms-web-app): throw if created document is missing id

a6software:
## Ticket Number

UCD-830

## Description of change

Throw an error if the document API service returns an object missing an `id`

The only way I can recreate the error raised by Karuna during testing is if the `id` is either `null` or `undefined`. The only way that could happen is when the document API call returns an object that is missing an `id`. The call is wrapped in a try / catch which throws anyway.

I'm confident the unit tests cover all the outcomes here. The situation as best I can see was a hazard of having the document service in a transitional state during dev, and should not be possible moving forwards. 

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/forms-web-app/src/lib/documents-api-wrapper.js
+++ b/forms-web-app/src/lib/documents-api-wrapper.js
@@ -44,5 +44,13 @@ exports.createDocument = async (appeal, formData) => {
     throw new Error(apiResponse.statusText);
   }
 
-  return apiResponse.json();
+  const response = await apiResponse.json();
+
+  if (!response.id) {
+    const msg = 'Document had no ID';
+    logger.warn({ response }, msg);
+    throw new Error(msg);
+  }
+
+  return response;
 };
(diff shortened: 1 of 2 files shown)


forms-web-app/src/lib/documents-api-wrapper.js
```diff
@@ -44,5 +44,11 @@ exports.createDocument = async (appeal, formData) => {
     throw new Error(apiResponse.statusText);
   }
 
-  return apiResponse.json();
+  const response = await apiResponse.json();
+
+  if (!response.id) {
```
MrSimonEmms, line 49: Is it worth adding some logging? I believe the `apiResponse` will log it anyway, but might make debugging easier

Something like `req.log.warn({ response }, 'Document had no ID')`

a6software: sure 👍

MrSimonEmms (approved):
Bellissimo

Fenisay (approved):

a6software:
ty

PR #231, opened 2020-12-16
test(e2e-tests): extend appeal statement feature file

frank-dodion:
## Ticket Number
UCD-991

## Description of change
Extend appeal statement feature file to include scenarios that cover when user has uploaded a valid file then had an invalid file rejected and finally proceeded with or without another valid file.


danWatkinson (approved):
👍 loving that re-use of the cy. commands there

PR #234, opened 2020-12-17
feat(e2e-tests): add of step definitions for appeal site address feature

Fenisay:
UCD-886

### DIFF
--- a/e2e-tests/cypress/integration/appeal-submission-appeal-site-address.feature
+++ b/e2e-tests/cypress/integration/appeal-submission-appeal-site-address.feature
@@ -1,4 +1,3 @@
-@wip
 Feature: Appellant provides the Appeal Site Address
     I need to provide the appeal site address, so that all parties know which site the appeal is against.
 
@@ -6,25 +5,24 @@ Feature: Appellant provides the Appeal Site Address
         Given the user is prompted for the site address
         When the user provides their appeal site address as <Address Line 1> and <Address Line 2> and <Town or City> and <County> and <Postcode>
         Then the user is able to continue with the provided address
-        #And the user can see that their appeal has been updated with the provided address
+        And the user can see that their appeal has been updated with the provided site address as <Address Line 1> and <Address Line 2> and <Town or City> and <County> and <Postcode>
         Examples:
             | Address Line 1  | Address Line 2 | Town or City | County       | Postcode  |
             | "1 Taylor Road" | "Clifton"      | "Bristol"    | "South Glos" | "BS8 1TG" |
-            | "2 Taylor Road" | ""             | ""           | "South Glos" | "BS8 1TG" |
-            | "4 Taylor Road" | ""             | "Bristol"    | "South Glos" | "BS8 1TG" |
-            | "5 Taylor Road" | "Clifton"      | ""           | "South Glos" | "BS8 1TG" |
+            | "2 Taylor Road" | ""             | ""           | "South Glos" | "M1 1AA" |
+            | "4 Taylor Road" | ""             | "Bristol"    | "South Glos" | "M60 1NW" |
+            | "5 Taylor Road" | "Clifton"      | ""           | "South Glos" | "DN55 1PT" |
 
     Scenario Outline: Prospective Appellant submits an appeal site address without mandatory information
         Given the user is prompted for the site address
         When the user provides their appeal site address as <Address Line 1> and <Address Line 2> and <Town or City> and <County> and <Postcode>
         Then the user is informed that they cannot continue with the provided address because <reason>
-        #And the user can see that their appeal has NOT been updated with the provided address
+        And the user can see that their appeal has not been updated with the provided address
         Examples:
             | Address Line 1 | Address Line 2 | Town or City | County       | Postcode  | reason                       |
-            | ""             | ""             | ""           | "South Glos" | "BS8 1TG" | "Address Line 1 is required" |
-            | "aaa"          | ""             | ""           | ""           | "BS8 1TG" | "County is required"         |
+            | ""             | ""             | ""           | "South Glos" | "W1A 1HQ" | "Address Line 1 is required" |
+            | "aaa"          | ""             | ""           | ""           | "EC1A 1BB" | "County is required"         |
             | "aaa"          | ""             | ""           | "South Glos" | ""        | "Postcode is required"       |
-
 
     Scenario: Prospective appellant fails to provide any address information
         Given the user is prompted for the site address
@@ -53,9 +51,9 @@ Feature: Appellant provides the Appeal Site Address
 
     Scenario Outline: Prospective appellant provides address data that exceeds the maximum length constraint for each field
         Given the user is prompted for the site address
-        When the user provides a value which is too long (<component>: <count>)
+        When the user provides a value which is too long - <component> : <count>
         Then the user is informed that they cannot continue with the provided address because <reason>
-        #And the user can see that their appeal has NOT been updated with the provided address
+        And the user can see that their appeal has not been updated with the provided address
         Examples:
             | component        | count | reason                                        |
             | "Address Line 1" | 61    | "Address Line 1 has a limit of 60 characters" |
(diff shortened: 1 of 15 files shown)


e2e-tests/cypress/integration/appeal-submission-appeal-site-address/appeal-submission-appeal-site-address.js
```diff
@@ -0,0 +1,122 @@
...
+    cy.provideAddressLine2(addressLine2);
+    cy.provideTownOrCity(townCity);
+    cy.provideCounty(county);
+    cy.providePostcode(postcode);
+    cy.clickSaveAndContinue();
+  },
+);
+
+When('the user provides a value which is too long - {string} : {int}', (component, count) => {
+  const value = Array(count + 1).join('x');
```
a6software, line 20: Is this the same as `'x'.repeat(count + 1)`

e2e-tests/cypress/integration/appeal-submission-appeal-site-address/appeal-submission-appeal-site-address.js
```diff
@@ -0,0 +1,122 @@
...
+        break;
+      case 'Postcode has a limit of 8 characters':
+        cy.confirmSiteAddressWasRejectedBecause('Postcode must be 8 characters or fewer');
+        break;
+    }
+  },
+);
+
+Then('the user is informed that {string}', (reason) => {
+  switch (reason) {
```
a6software, line 111: is it worth having a `default` case that `throw`s if the given reason isn't one of the expected reasons? 

idk if it is, just a thought

forms-web-app/tests/unit/validators/appellant-submission/site-location.test.js
```diff
@@ -368,6 +368,38 @@ describe('validators/appellant-submission/site-location', () => {
...
+            'site-address-line-one': '1 Taylor Road',
+            'site-address-line-two': 'Clifton',
+            'site-town-city': 'Bristol',
+            'site-county': 'South Glos',
+            'site-postcode': 'aaaaaaa',
+          },
+        }),
+        expected: (result) => {
+          expect(result.errors).toHaveLength(1);
+          expect(result.errors[0].msg).toEqual("Postcodes can't be all letters");
```
a6software, line 400: are these error messages approved by Marc & Daisy?

e2e-tests/cypress/integration/appeal-submission-appeal-site-address.feature
```diff
@@ -53,7 +51,7 @@ Feature: Appellant provides the Appeal Site Address
 
     Scenario Outline: Prospective appellant provides address data that exceeds the maximum length constraint for each field
         Given the user is prompted for the site address
-        When the user provides a value which is too long (<component>: <count>)
+        When the user provides a value which is too long - <component> : <count>
         Then the user is informed that they cannot continue with the provided address because <reason>
         #And the user can see that their appeal has NOT been updated with the provided address
```
frank-dodion, line 56: This looks like an important behaviour. Should it be commented out?

e2e-tests/cypress/support/appeal-submission-appeal-site-address/confirmSiteAddressWasAccepted.js
```diff
@@ -0,0 +1,7 @@
+module.exports = () => {
```
frank-dodion, line 1: Should we also navigate back to the address input page and see that the recently added input values are displayed?

e2e-tests/cypress/integration/appeal-submission-appeal-site-address/appeal-submission-appeal-site-address.js
Fenisay: Maybe lol

e2e-tests/cypress/integration/appeal-submission-appeal-site-address.feature
Fenisay: I assumed that it was while we were waiting for the check your answer page. So I just took the feature file as it is

e2e-tests/cypress/support/appeal-submission-appeal-site-address/confirmSiteAddressWasAccepted.js
Fenisay: Well it's apparently the hole point of the new integration stories, but the back button will not prove that the data is stored. I'll update this PR later as I'm on other tickets now

frank-dodion: I am not suggesting use of the back button. I would expect navigation by url and to be able to see the newly added data to confirm that it has not been changed in the case of invalid inputs.

e2e-tests/cypress/integration/appeal-submission-appeal-site-address.feature
frank-dodion: I think we should be able to confirm the updated state without the check your answers page.

forms-web-app/tests/unit/validators/appellant-submission/site-location.test.js
Fenisay: It was on the feature file so I guess. And I hope because I had to add this behavior...

e2e-tests/cypress/integration/appeal-submission-appeal-site-address/appeal-submission-appeal-site-address.js
Fenisay: Hmm... maybe to force a failure

Fenisay: Really nice catch, I found an untested test !

forms-web-app/tests/unit/validators/appellant-submission/site-location.test.js
danWatkinson: yeah we need to be a bit careful here.

The feature file is supposed to be 'logical'. If there's a reason a thing should fail, we want to state what that reason is.
It is -not- stating the text the user should see

forms-web-app/src/validators/appellant-submission/site-location.js
```diff
@@ -1,5 +1,20 @@
 const { body } = require('express-validator');
 
+function validatePostcode(postcode) {
+  const firstChar = postcode.charAt(0);
+
+  if (!Number.isNaN(parseInt(firstChar, 10))) {
```
danWatkinson, line 6: I'm happy with the implementation of the stepdefs above,
I'm not sure about these as validation steps though?

We only put 2 example 'bad postcodes' in as an example of things we thought should fail, it's not supposed to be an exhaustive list (as things stand.. maybe it should be?)

My assumption at the time was: there will be a postcode validator, it will reject bad postcodes - what i'm doing here is proving it with some trivial examples.

If we're going to use the feature file to literally test-drive out the validation then we need more examples than those offered.

Additionally - I'd talk to the API guys about whether this validation should be done inside forms web app or inside the API... remember: if there's an API you can post data to it, and if you can post data to it you might post invalid data... so protecting the 'client' is less useful than protecting the API.
(That last bit isn't my business, incidentally - so long as the tests are green i'm not fussed  -where- the validation is.. but I suspect the API guys might be..)

forms-web-app/tests/unit/validators/appellant-submission/site-location.test.js
```diff
@@ -368,6 +368,38 @@ describe('validators/appellant-submission/site-location', () => {
           expect(result.errors[0].msg).toEqual('Postcode must be 8 characters or fewer');
         },
       },
+      {
+        title: 'invalid - site-postcode shoudl begin with a letter',
+        given: () => ({
```
danWatkinson, line 373: shoudl->should

forms-web-app/src/validators/appellant-submission/site-location.js
Fenisay: I changed the regex and added more examples

Fenisay: ([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})

I added different valid examples throughout the feature file

forms-web-app/tests/unit/validators/appellant-submission/site-location.test.js
Fenisay: fixed

e2e-tests/cypress/support/appeal-submission-appeal-site-address/confirmSiteAddressWasAccepted.js
Fenisay: I did this in the uncommented scenario line as it's the new "template".

forms-web-app/src/validators/appellant-submission/site-location.js
Fenisay: Concerning the point of the API, all the validation feature per feature is not yet done in the backend. We are just dealing with trivial inconsistencies at the moment. Maybe it will change but it will require its own US and it will not be trivial one.

frank-dodion (approved):

PR #235, opened 2020-12-17
feat(e2e-tests): update applicant name feature

Fenisay:
UCD-1159

### DIFF
--- a/e2e-tests/cypress/integration/appellant-submission-applicant-name.feature
+++ b/e2e-tests/cypress/integration/appellant-submission-applicant-name.feature
@@ -2,15 +2,31 @@
 Feature: The user gives the original appellant name
   The users needs to give the original appellant name to save and continue.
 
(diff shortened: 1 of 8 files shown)


MrSimonEmms (commented):
Based on the fact that this is just changing a test file (and not the published source code), arguably this should be a `chore` or `test` commit, not a `feat`.

danWatkinson:
i would say that if we're hoping to accumulate the whole story in this PR rather than just the test-changes.. the statement re: feature/chore may or may not be valid..
in this instance i think i agree: the net result should be parity, so this is really a refactoring excercise/chore that just happens to include some test-tweaking to make sure that we -would- catch any problems we created in the rest of the change.

other opinions are available.

a6software (approved):
Approving on the basis the code looks fine.

danWatkinson (approved):
looks good

PR #244, opened 2020-12-17
test(e2e-tests): updated Planning Application Number submission feature

frank-dodion:
## Ticket Number
UCD-1037

## Description of change
Updated feature file to include max character constraint example


Fenisay (approved):

Fenisay:
Simon told me that as we are just changing the test code we should put 

chore(e2e-tests): message

instead of feat or fix

frank-dodion:
> Simon told me that as we are just changing the test code we should put
> 
> chore(e2e-tests): message
> 
> instead of feat or fix

Thanks for approval. Interesting point about chore or test. In this case it is looks to me like a test commit.
Screenshot 2020-12-17 at 13 39 13

PR #251, opened 2020-12-17
feat(forms-web-app): stop setting appeal statement sensitive info cb

frank-dodion:
## Ticket Number
UCD-830

## Description of change
Updated behaviour of checkbox on appeal statement page so that it does not automatically follow the state of the appeal.

### DIFF
--- a/forms-web-app/src/views/appellant-submission/appeal-statement.njk
+++ b/forms-web-app/src/views/appellant-submission/appeal-statement.njk
@@ -147,7 +147,6 @@
           name: "does-not-include-sensitive-information",
           items: [
             {
-              checked: (appeal.yourAppealSection.appealStatement.hasSensitiveInformation === false),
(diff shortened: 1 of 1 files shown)


Fenisay:
What's the expected behavior ?

Fenisay (approved):

frank-dodion:
> What's the expected behavior ?

The small PR about the appeal statement is to stop setting the state of the sensitive information checkbox based on the state of the appeal. This will mean that the checkbox will be a UI-only component that will be unchecked each time the page is visited. The user must then explicitly click each time they want to proceed with save and continue. Apparently this is what the client asked for when reviewing 830.

frank-dodion:
Thanks for the approval Yassine. I will hold off merging this for now because we will want the formal requirements and JIRA tickets to catch up :-)

PR #268, opened 2020-12-21
feat(appeals-service-api): extend validation for original appellant

GoEagles999:
## Ticket Number

UCD-1279

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/appeals-service-api/src/__tests__/appeals.spec.js
+++ b/appeals-service-api/src/__tests__/appeals.spec.js
@@ -415,7 +415,7 @@ describe('Appeals API', () => {
   test('PUT /api/v1/appeals/{id} - It responds with an error - original appellant cannot be unspecified', async () => {
     const appeal = await createAppeal();
 
-    appeal.sectionStates.aboutYouSection.isOriginalApplicant = null;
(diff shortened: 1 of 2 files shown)


Fenisay:
You should remove all the sectionStates mentions as it's not the appeal but the status (COMPLETED, IN PROGRESS etc) of the tasks

Fenisay (changes requested):
You should remove all the sectionStates mentions as it's not the appeal but the status (COMPLETED, IN PROGRESS etc) of the tasks

And I'm pretty sure that this rule won't allow you to go on any screen. This method is really about to verify that there is no inconsistency in the model. For example, uploaded file id existing without the name. It's not a state that should be possible.  

Otherwise, having this field not valued happens every time you create a new appeal, so basically having this rule will prevent us to start any screen which is not original appellant and we do not want that.

Fenisay:
What I would do at your place will be to ensure that  :

if isOriginalAppellant is true then the behalfAppellantNameOf should not be valued

If isOriginalAppellant is false then the behalfAppellantNameOf should be valued

But you should double check that it's not already existing

Fenisay (approved):

PR #273, opened 2020-12-21
feat(appeals-service-api): extend validation in appeals-service api f…

Fenisay:
UCD-888

### DIFF
--- a/appeals-service-api/src/__tests__/appeals.spec.js
+++ b/appeals-service-api/src/__tests__/appeals.spec.js
@@ -346,6 +346,32 @@ describe('Appeals API', () => {
     expect(response.statusCode).toBe(400);
   });
 
+  test('PUT /api/v1/appeals/{id} - It responds with an error - the appellant owning the whole site but still told other owners should fail', async () => {
+    const appeal = await createAppeal();
+    appeal.appealSiteSection.siteOwnership.ownsWholeSite = true;
+    appeal.appealSiteSection.siteOwnership.haveOtherOwnersBeenTold = true;
+
+    const response = await request(app).put(`/api/v1/appeals/${appeal.id}`).send(appeal);
+    expect(response.body.code).toEqual(400);
+    expect(response.body.errors).toContain(
+      'If the appellant owns the whole appeal site there can be no other owners'
+    );
+    expect(response.statusCode).toBe(400);
+  });
+
+  test('PUT /api/v1/appeals/{id} - It responds with an error - other owner told without knowing if there is any should fail', async () => {
+    const appeal = await createAppeal();
+    appeal.appealSiteSection.siteOwnership.haveOtherOwnersBeenTold = true;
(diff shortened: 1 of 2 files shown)


appeals-service-api/src/__tests__/appeals.spec.js
```diff
@@ -346,6 +346,32 @@ describe('Appeals API', () => {
...
 
+  test('PUT /api/v1/appeals/{id} - It responds with an error - the appellant owning the whole site but still told other owners should fail', async () => {
+    const appeal = await createAppeal();
+    appeal.appealSiteSection.siteOwnership.ownsWholeSite = true;
+    appeal.appealSiteSection.siteOwnership.haveOtherOwnersBeenTold = true;
+
+    const response = await request(app).put(`/api/v1/appeals/${appeal.id}`).send(appeal);
+    expect(response.body.code).toEqual(400);
+    expect(response.body.errors).toContain(
+      'If the appellant own the hole appeal site there is any other owners'
```
a6software, line 357: "If the appellant owns the whole appeal site there can be no other owners"

appeals-service-api/src/__tests__/appeals.spec.js
```diff
@@ -346,6 +346,32 @@ describe('Appeals API', () => {
...
+
+  test('PUT /api/v1/appeals/{id} - It responds with an error - other owner told without knowing if there is any should fail', async () => {
+    const appeal = await createAppeal();
+    appeal.appealSiteSection.siteOwnership.haveOtherOwnersBeenTold = true;
+    appeal.appealSiteSection.siteOwnership.ownsWholeSite = null;
+
+    const response = await request(app).put(`/api/v1/appeals/${appeal.id}`).send(appeal);
+    expect(response.body.code).toEqual(400);
+    expect(response.body.errors).toContain(
+      'We should know if there is another owners before knowing if they were told'
```
a6software, line 370: Not sure how to word this. It raises a point around how technical or non-technical these error messages should be. 

The other messages I have seen are non technical. However, as these messages are not intended for public consumption (afaik), imo it would make more sense to be terse and explicit:

`haveOtherOwnersBeenTold cannot be true when `ownsWholeSite is null`

a6software (dismissed):
I'm approving as the technical details of this are fine. 

Not sure of the error message wording - not really sure who is deciding this.

appeals-service-api/src/__tests__/appeals.spec.js
Fenisay: It was my french brain who decided so I'll go with your version thanks

appeals-service-api/src/__tests__/appeals.spec.js
Fenisay: I completely agree, I followed the existing mood but we definitely should have this task

a6software (approved):

PR #302, opened 2021-01-06
test(e2e-tests): feature file to check decision date eligibility for mvp

frank-dodion:
## Ticket Number
UCD-1352

## Description of change
Feature file to check decision date eligibility for mvp

### DIFF
--- a/e2e-tests/cypress/integration/mvp-eligibility-check-decision-date.feature
+++ b/e2e-tests/cypress/integration/mvp-eligibility-check-decision-date.feature
@@ -1,0 +1,32 @@
+@wip
+Feature: Check that decision date is eligible to commence an appeal
+  Note - This feature is intended to support redirection to ACP Service as part of initial MVP
+
+  Scenario: Eligibility decision dates less than 12 weeks old can proceed
+    Given the application decision date is requested
+    When a decision date is provided that is less than 12 weeks old
+    Then the appeal can be commenced
+
+  Scenario: Eligibility decision dates older than 12 weeks cannot proceed
+    Given the application decision date is requested
+    When a decision date is provided that is more than 12 weeks old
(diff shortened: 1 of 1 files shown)


danWatkinson (commented):
can we put this somewhere other than /integration? given that we're putting out an 'alternative description of the world' i was planning on just putting it in a different root-folder so it never got run accidentally etc

frank-dodion:
> can we put this somewhere other than /integration? given that we're putting out an 'alternative description of the world' i was planning on just putting it in a different root-folder so it never got run accidentally etc

Will do. I'll hold off until UCD-1371 is available and I will update to comply with that.

e2e-tests/cypress/integration/mvp-eligibility-check-decision-date.feature
```diff
@@ -0,0 +1,32 @@
...
+  Scenario: Eligibility decision dates less than 12 weeks old can proceed
+    Given the application decision date is requested
+    When a decision date is provided that is less than 12 weeks old
+    Then the appeal can be commenced
+
+  Scenario: Eligibility decision dates older than 12 weeks cannot proceed
+    Given the application decision date is requested
+    When a decision date is provided that is more than 12 weeks old
+    Then the decision date is highlighted as beyond the deadline for appeal
+
```
Fenisay, line 14: I'm fine using highlighted in both scenario but idk if it was what you meant or not. When the decision date expired you are redirected to a specific page. This is not an error message like it would be if the date is invalid.

frank-dodion: I am not sure what you mean by 'both' here. This highlighted scenario is a key behaviour and I don't expect it to be handled as an error. Instead it will provide an end of journey page similar to this.
Screenshot 2021-01-07 at 08 35 52

PR #315, opened 2021-01-07
feat(forms-web-app): Google Analytics feature

Fenisay:
UCD-1178

### DIFF
--- a/e2e-tests/cypress/support/appeal-header-footer-links/confirmGoogleAnalyticsLinkIsPresent.js
+++ b/e2e-tests/cypress/support/appeal-header-footer-links/confirmGoogleAnalyticsLinkIsPresent.js
@@ -1,0 +1,8 @@
+module.exports = () => {
+  cy.get('head script[data-cy="Google Analytics"]').should(
+    'have.attr',
+    'src',
+    'https://www.googletagmanager.com/gtag/js?id=G-FFM8G1FWCZ',
+  );
+  cy.wait(Cypress.env('demoDelay'));
+};
--- a/packages/forms-web-app/src/config.js
+++ b/packages/forms-web-app/src/config.js
@@ -45,5 +45,6 @@ module.exports = {
     port: Number(process.env.PORT || 3000),
     sessionSecret: process.env.SESSION_KEY,
     useSecureSessionCookie: process.env.USE_SECURE_SESSION_COOKIES === 'true',
+    googleAnalyticsId: 'G-FFM8G1FWCZ',
   },
 };
--- a/packages/forms-web-app/src/views/includes/head.njk
+++ b/packages/forms-web-app/src/views/includes/head.njk
@@ -3,14 +3,15 @@
   <!--[if lte IE 8]><link href="/public/stylesheets/main-ie8.css" rel="stylesheet" type="text/css" /><![endif]-->
   <!--[if gt IE 8]><!--><link href="/public/stylesheets/main.css" media="all" rel="stylesheet" type="text/css" /><!--<![endif]-->
 
+
   <!-- Global site tag (gtag.js) - Google Analytics -->
-  <script data-cy="Google Analytics" async src="https://www.googletagmanager.com/gtag/js?id=G-FFM8G1FWCZ"></script>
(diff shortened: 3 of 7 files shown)


rcmsoftwaredevelopment (dismissed):

packages/forms-web-app/src/views/includes/head.njk
```diff
@@ -3,4 +3,14 @@
...
   <!--[if gt IE 8]><!--><link href="/public/stylesheets/main.css" media="all" rel="stylesheet" type="text/css" /><!--<![endif]-->
 
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script data-cy="Google Analytics" async src="https://www.googletagmanager.com/gtag/js?id=G-FFM8G1FWCZ"></script>
+  <script>
+    window.dataLayer = window.dataLayer || [];
+    function gtag(){dataLayer.push(arguments);}
+    gtag('js', new Date());
+
+    gtag('config', 'G-FFM8G1FWCZ');
```
MrSimonEmms, line 13: Can we put this in `config.js`?

packages/forms-web-app/src/views/includes/head.njk
```diff
@@ -3,4 +3,14 @@
   <!--[if lte IE 8]><link href="/public/stylesheets/main-ie8.css" rel="stylesheet" type="text/css" /><![endif]-->
   <!--[if gt IE 8]><!--><link href="/public/stylesheets/main.css" media="all" rel="stylesheet" type="text/css" /><!--<![endif]-->
 
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script data-cy="Google Analytics" async src="https://www.googletagmanager.com/gtag/js?id=G-FFM8G1FWCZ"></script>
```
MrSimonEmms, line 7: Can we put the id in `config.js`?

MrSimonEmms (changes requested):

packages/forms-web-app/src/views/includes/head.njk
Fenisay: We could but from what I understood the vault is descoped and it will be one key for all the environments. So I'm not sure we need to make this configurable. Am I missing something ?

MrSimonEmms: Yes, it's been descoped as putting as an envvar, but I still think this should be stored in the `config.js` so it's a single source of truth. Something like:
```
module.exports = {
  server: {
    googleAnalytics: 'G-FFM8G1FWCZ'  
  }
}
```

At the very least, as this value is used twice in the template, I think it would be better putting to a variable - to my mind, getting that variable from `config.js` is as good a place as any to store it

MrSimonEmms (approved):

rcmsoftwaredevelopment (approved):

packages/forms-web-app/src/config.js
```diff
@@ -45,5 +45,6 @@ module.exports = {
     port: Number(process.env.PORT || 3000),
     sessionSecret: process.env.SESSION_KEY,
     useSecureSessionCookie: process.env.USE_SECURE_SESSION_COOKIES === 'true',
+    googleAnalyticsId: 'G-FFM8G1FWCZ',
```
a6software, line 48: if this isn't going to be an environment variable it seems simpler just to keep it hardcoded in the script file. 

I don't really mind either way, just a comment.

e2e-tests/cypress/support/appeal-header-footer-links/confirmGoogleAnalyticsLinkIsPresent.js
```diff
@@ -0,0 +1,8 @@
+module.exports = () => {
+  cy.get('head script[data-cy="Google Analytics"]').should(
+    'have.attr',
+    'src',
+    'https://www.googletagmanager.com/gtag/js?id=G-FFM8G1FWCZ',
+  );
+  cy.wait(Cypress.env('demoDelay'));
```
a6software, line 7: be fun watching the cypress videos for this one :D

a6software (approved):

MrSimonEmms (approved):

PR #316, opened 2021-01-07
feat(forms-web-app): update submission page title and error message

Fenisay:
UCD-791


danWatkinson (commented):
Question:
If this runs green in cypress, doesn't that mean we're missing a cypress test that cares about the title + a test that cares about the error message?

I appreciate that unit-testing provides the same security, but it doesn't provide the same -proof-. I can't hand over a video of this being proved.

What i -can- do is grab a video from one of our scenarios and say "look, this is excercising the functionality" but i have no proof that we're asserting anything and i'm reliant on either executing the tests myself locally or scrolling backwards+forwards over a video trying to read some text, which defeats the point of the automation.

I'm open to other views, but my opening gambit is that we should have cypress tests that prove this, whether or not we have unit-tests.

Fenisay:
> Question:
> If this runs green in cypress, doesn't that mean we're missing a cypress test that cares about the title + a test that cares about the error message?
> 
> I appreciate that unit-testing provides the same security, but it doesn't provide the same -proof-. I can't hand over a video of this being proved.
> 
> What i -can- do is grab a video from one of our scenarios and say "look, this is excercising the functionality" but i have no proof that we're asserting anything and i'm reliant on either executing the tests myself locally or scrolling backwards+forwards over a video trying to read some text, which defeats the point of the automation.
> 
> I'm open to other views, but my opening gambit is that we should have cypress tests that prove this, whether or not we have unit-tests.
> 
> (and i've just written this and seen it go back into draft so this may all be irrelevant, but i've typed it now..) ;)

+ the cypress was not green as expected, sorry I miss click the tag :P

e2e-tests/cypress/integration/appellant-confirms-declaration.feature
```diff
@@ -0,0 +1,12 @@
+@UI-ONLY
+Feature: Declaration must be agreed before a submission can be made
+
+  Scenario: Prospective appellant does not agree to the declaration
+    Given an appeal wants to be submitted
```
danWatkinson, line 5: phrasing?
it could be ok, but that feels like quite a jump- starting to give wants+needs to the appeal itself.
given an appeal has been prepared? given a prospective appellant has completed their appeal details?

(not a problem in terms of it's ability to prove what's going on, just not sure that's the phrasing we'd have chosen?)

a6software (approved):
LGTM

PR #330, opened 2021-01-08
feat: Update forms-web-app, feature file, glue code to check if the local planning department is stored

Fenisay:
## Ticket Number

UCD-974

## Description of change

Update forms-web-app, feature file, glue code to check if the local planning department is stored
## Checklist

- [x] New features have tests

## Important


e2e-tests/cypress/integration/eligibility-local-planning-department.feature
```diff
@@ -3,14 +3,23 @@ Feature: A prospective appellant supplies a Local Planning Department for the ca
...
+    Given the list of Local Planning Department is presented
+    When the user provides a Local Planning Department not in the provided list
+    Then the user is informed that a Local Planning Department in the provided list is required
+    And appeal is not updated with the unknown Local Planning Department
 
   Scenario: Some Local Planning Departments are not participating in this service
+    Given the list of Local Planning Department is presented
     When the user provides a Local Planning Department that is not participating in this service
     Then the user is informed that the selected Local Planning Department is not participating in the service
-    And the user is directed to the Appeal a Planning Decision service
```
danWatkinson, line 12: are we deliberately losing the confirmation that we're pointing the user to the other service?

we -could- just wrap that confirmation up  in 'the user is informed that the selected LPD is not participating' step maybe if we're specifically trying to reduce our 'and' count? but i don't see that check happening anywhere in the PR, so feels like we either keep the existing line or wire its check into an existing step..

Fenisay: Discussed in pm

danWatkinson: cool, happy we're not losing any coverage 👍 

cool from a QA perspective, assume we still need a review of the code, so not specifically ticking this as 'approved' as i'm not reading anything other than the e2e-tests section..

packages/forms-web-app/src/controllers/eligibility/planning-department.js
```diff
@@ -1,31 +1,67 @@
...
+    res.render(VIEW.ELIGIBILITY.PLANNING_DEPARTMENT, {
+      appeal,
+      errors,
+      errorSummary: [{ text: e.toString(), href: '#' }],
+    });
     return;
   }
 
-  res.redirect(`/${VIEW.ELIGIBILITY.LISTED_BUILDING}`);
+  if (!errors['local-planning-department']) {
```
a6software, line 62: We have the the check for ` if (errors['local-planning-department'])` on line 33.

If we get to line 62, is it possible we might not trigger this conditional? 

If we don't trigger the conditional, the controller wouldn't return a response (?) so we'd be in a bad state. Can we remove this conditional check, or add in another final fall back?

packages/forms-web-app/src/controllers/eligibility/planning-department.js
```diff
@@ -1,31 +1,67 @@
...
   const { departments } = await getDepartmentData();
-
+  const { appeal } = req.session;
+  let appealLPD = '';
+  if (appeal.lpaCode) {
+    const lpd = await getDepartmentFromId(appeal.lpaCode);
+    if (lpd) {
+      appealLPD = lpd.name;
+    }
+  }
```
a6software, line 17: seems fair enough that because this code is needed in two controllers, we have some repetition here. If we need this in a third place, I think we should consider extracting to a middleware like we do for the `appeal`.

packages/forms-web-app/tests/unit/controllers/check-answers.test.js
```diff
@@ -1,17 +1,54 @@
...
+        appeal,
+      });
+    });
+    it('should call the correct template with empty local planning department', async () => {
+      await getDepartmentFromId.mockResolvedValue(undefined);
+
+      appeal.lpaCode = 'lpdCode';
+      await checkAnswersController.getCheckAnswers(req, res);
+
+      checkAnswersController.getCheckAnswers(req, res);
```
a6software, line 37: is this intentionally called twice - line 35 / 37 ? One without the `await`?

packages/forms-web-app/src/controllers/eligibility/planning-department.js
Fenisay: Yep I agree and I wondered too but I think it will be the only times

packages/forms-web-app/src/controllers/eligibility/planning-department.js
Fenisay: The final one is because we could be in this section if the department is ineligible as we want still want to store the lpaCode in the appeal but at this point the render is already triggered. All the errors are caught except this one.

packages/forms-web-app/tests/unit/controllers/check-answers.test.js
Fenisay: I don't think so, I fix this !

a6software (approved):

PR #331, opened 2021-01-08
chore(coverage): update jest config to achieve 100 coverage

MrSimonEmms:
## Ticket Number

N/A

## Description of change

Simplify the coverage configuration. Everything set to 100% and then ignored things we had excluded. It saves having to update the numbers if we add code that then changes the coverage calculations

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/jest.config.js
+++ b/packages/forms-web-app/jest.config.js
@@ -7,12 +7,13 @@ module.exports = {
   collectCoverage: true,
   collectCoverageFrom: ['src/**/*.js'],
   coverageReporters: ['json', 'html', 'text', 'text-summary'],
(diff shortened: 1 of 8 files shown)


packages/forms-web-app/jest.config.js
```diff
@@ -7,12 +7,13 @@ module.exports = {
...
-      functions: 94,
-      lines: 93,
-      statements: 93,
+      branches: 100,
+      functions: 100,
+      lines: 100,
+      statements: 100,
     },
   },
 };
```
Fenisay, line 19: I know someone that will be happy @a6software

MrSimonEmms: For the record, this doesn't actually increase what we've got covered. It just simplifies this so, when adding new middleware for instance, we don't have to change the coverage calculations

a6software (approved):

PR #336, opened 2021-01-11
test(e2e-tests): update tests and docs for ACP integration

frank-dodion:
## Ticket Number
UCD-1449

## Description of change
Update tests and docs for ACP integration

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,3 +256,11 @@ Kubernetes cluster.
 ```shell script
 kubectl logs -f -n app-prod app-form-web-app-1111111111-aaaaa
 ```
+### ACP integration
+
+The current target MVP checks only the decision date eligibility before passing the user to an existing external service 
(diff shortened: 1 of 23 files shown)


Fenisay (approved):
LGTM
Just a note to say that now we could be on the right url and don't see the cypress test failing if we do not receive a 2XX code. So it's a thing to add on the pros for "Should we test some contents on top of the url"

frank-dodion:
> LGTM
> Just a note to say that now we could be on the right url and don't see the cypress test failing if we do not receive a 2XX code. So it's a thing to add on the pros for "Should we test some contents on top of the url"

Thanks for review/approval. The updated to cypress functions to suppress errors for non 2xx responses applies only to navigation and not assertions so it should not undermine our tests. We have 'confirm navigation' functions that check for the correct url. These do not rely on the response code.

PR #347, opened 2021-01-12
test(e2e-tests): update section status icon feature file

Fenisay:
## Ticket Number

UCD-766

## Description of change

update section status icon feature file

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/e2e-tests/cypress/integration/appeal-submission-tasklist.feature
+++ b/e2e-tests/cypress/integration/appeal-submission-tasklist.feature
@@ -1,33 +1,51 @@
-@wip
-Feature: Your Details task status
+Feature: Task lists - Section status
+  As a prospective appellant I want to clearly see which sections of the
+  appeal submission I have completed, so that I do not waste my time checking
 
-  Scenario: The status of the Your Details section is accurate
-    Given the user has not been in "Your Details" section
-    When the user checks the status of their appeal
-    Then the user should see that the "Your Details" task is "NOT STARTED"
+  Initially all the tasks are "NOT STARTED", once the appellant has provided required information they are "COMPLETED"
+  Some tasks could be multi steps and if not all steps are completed, then the task is considered "IN PROGRESS"
 
-  Scenario Outline: The status of the Your Details section is accurate
-    Given the user "is" the original appellant
-    And the user has provided their name: <name> and email: <email>
-    When the user checks the status of their appeal
-    Then the user should see that the "Your Details" task is <status>
-
+  Scenario Outline: All the tasks are in "NOT STARTED" state
+    Given the <task> part of the appeal is not started
+    When the appeal tasks are presented
+    Then the state for <task> is displayed to be "NOT STARTED"
     Examples:
-      | name        | email            | status        |
-      | ""          | ""               | "IN PROGRESS" |
-      | ""          | "good@email.com" | "IN PROGRESS" |
-      | "Good Name" | ""               | "IN PROGRESS" |
-      | "Good Name" | "good@email.com" | "COMPLETED"   |
+      | task                                            |
+      | "About you - Your details"                      |
+      | "Planning application - Application number"     |
+      | "Planning application - Upload application"     |
+      | "Planning application - Upload decision letter" |
+      | "Your appeal - Appeal statement"                |
+      | "Your appeal - Supporting documents"            |
+      | "Appeal site - Site location"                   |
+      | "Appeal site - Site ownership"                  |
+      | "Appeal site - Site access"                     |
+      | "Appeal site - Site safety"                     |
 
-
-  Scenario Outline: The status of the Your Details section is accurate
-    Given the user "is not" the original appellant
-    And the user has provided their name: <name> and email: <email>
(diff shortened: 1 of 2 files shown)


e2e-tests/cypress/integration/appeal-submission-tasklist.feature
```diff
@@ -1,33 +1,52 @@
 @wip
-Feature: Your Details task status
+Feature: Task lists - Section status
+  As a prospective appellant I want to clearly see which sections of the
+  appeal submission I have completed, so that i do not waste my time checking
```
frank-dodion, line 4: I instead of i

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
```diff
@@ -1,33 +1,52 @@
...
 
-  Scenario: The status of the Your Details section is accurate
-    Given the user has not been in "Your Details" section
-    When the user checks the status of their appeal
-    Then the user should see that the "Your Details" task is "NOT STARTED"
+  Initially all the tasks are "NOT STARTED", once the appellant provided required information they are "COMPLETED"
```
frank-dodion, line 6: ...has provided...

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
Fenisay: fixed

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
Fenisay: fixed

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
```diff
@@ -1,33 +1,52 @@
...
+    Then the status of the "About you - Your details" is "IN PROGRESS"
 
-    Examples:
-      | name                | email            | on behalf of | status        |
-      | "Another Good Name" | "good@email.com" | ""           | "IN PROGRESS" |
-      | "Another Good Name" | "good@email.com" | "Good Name"  | "COMPLETED"   |
+  Scenario: The status of the "Site ownership" task is IN PROGRESS
+    Given the site ownership is presented for the first time
+    When it is confirmed that the whole site is not owned
+    Then the status of the "Appeal site - Site ownership" is "IN PROGRESS"
```
frank-dodion, line 52: I think this may need additional scenarios to cover at least the two examples from the previous examples table but maybe even more scenarios if we want to cover all possibilities.

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
```diff
@@ -1,33 +1,52 @@
...
-  Scenario Outline: The status of the Your Details section is accurate
-    Given the user "is not" the original appellant
-    And the user has provided their name: <name> and email: <email>
-    And the user has provided <on behalf of> as the name of the original appellant
-    When the user checks the status of their appeal
-    Then the user should see that the "Your Details" task is <status>
+  Scenario: The status of the "Your details" task is IN PROGRESS
+    Given confirmation of whether appellant is original applicant is requested
+    When confirmation provided that appellant is original applicant
+    Then the status of the "About you - Your details" is "IN PROGRESS"
```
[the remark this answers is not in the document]
Fenisay: Dan asked a question like this and Ian said we should not cover every combination. The purpose is to ensure that these sections could be in progress and the business is never considering the scenario of a wild user to be realistic even if I agree with you...

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
Fenisay: > I think this may need additional scenarios to cover at least the two examples from the previous examples table but maybe even more scenarios if we want to cover all possibilities.

It's not my understanding to cover all the possibilities as for example the original appellant name is not supposed to be answered if the appellant did not answered not being the original appellant before. If we try to do this, it will trigger an appeal-service-api inconsistency error so it's really a bigger problem/topic. 

But we could raise this point at the daily or wait for an other view on the question.

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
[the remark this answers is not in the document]
frank-dodion: OK. I am moving to the view that this should be a feature file that is concerned only with displaying the correct status badge rather than exercising the rules to create a particular state. I think that should be on other feature files related to specific sections/features. In that case we could use a general scenario outline like Given <section> is <state>, When the appeal tasks are presented, Then the state for <section> is displayed to be <state>. If we take this approach then the various section feature files should be concerned about how we get to be in each of the possible states i.e. the business rules.

e2e-tests/cypress/integration/appeal-submission-tasklist.feature
frank-dodion: I agree that it is best for this feature file to focus only on displaying correct status badge so we could remove the scenarios that consider business rules to get into a particular state.

frank-dodion (approved):

PR #357, opened 2021-01-12
test(e2e-tests): glue codes for task status feature

Fenisay:
## Ticket Number

UCD-766
UCD-935
## Description of change

glue codes for task status feature file
## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/services/task.service.js
+++ b/packages/forms-web-app/src/services/task.service.js
@@ -9,9 +9,9 @@ function statusAppealStatement(appeal) {
   return task.uploadedFile.id ? TASK_STATUS.COMPLETED : TASK_STATUS.NOT_STARTED;
 }
 
-// eslint-disable-next-line no-unused-vars
-function statusOtherDocuments(appeal) {
-  return TASK_STATUS.NOT_STARTED;
+function statusSupportingDocuments(appeal) {
+  const task = appeal.yourAppealSection.otherDocuments;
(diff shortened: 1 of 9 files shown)


packages/forms-web-app/src/services/task.service.js
```diff
@@ -9,9 +9,9 @@ function statusAppealStatement(appeal) {
   return task.uploadedFile.id ? TASK_STATUS.COMPLETED : TASK_STATUS.NOT_STARTED;
 }
 
-// eslint-disable-next-line no-unused-vars
-function statusOtherDocuments(appeal) {
-  return TASK_STATUS.NOT_STARTED;
+function statusSupportingDocuments(appeal) {
+  const task = appeal.yourAppealSection.otherDocuments;
+  return task.uploadedFiles.length > 0 ? TASK_STATUS.COMPLETED : TASK_STATUS.NOT_STARTED;
```
a6software, line 14: Supporting docs is optional - 0 can be a valid number for the task to be completed, so long as the user has submitted the form. 

I'm not sure of the "fix" for this though, I need more time to think about it.

Fenisay: I kept the same logic than other pages and the status are updated only if the appeal is modified. Otherwise, other screens should be IN PROGRESS at the moment they first accessed.

The difference here is that this task is not mandatory for the check your answers part

PR #410, opened 2021-01-26
feat(forms-web-app): ensure LPD data is always refreshed

frank-dodion:
## Ticket Number
UCD-1563

## Description of change
Ensure LPD data is always refreshed in LPD controller

### DIFF
--- a/packages/forms-web-app/tests/unit/services/department.service.test.js
+++ b/packages/forms-web-app/tests/unit/services/department.service.test.js
@@ -2,6 +2,7 @@ const { resetDepartments } = require('../../../src/services/department.service')
 const { getDepartmentFromName } = require('../../../src/services/department.service');
 const { getDepartmentFromId } = require('../../../src/services/department.service');
 const { getDepartmentData } = require('../../../src/services/department.service');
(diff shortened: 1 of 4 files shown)


rcmsoftwaredevelopment (approved):

packages/forms-web-app/tests/unit/services/department.service.test.js
```diff
@@ -12,14 +13,69 @@ describe('services/department.service', () => {
     jest.resetAllMocks();
   });
 
+  describe('getRefreshedDepartmentData', () => {
+    it('should  give empty result ', async () => {
+      await getLPAList.mockResolvedValue({ data: [] });
```
a6software, line 18: minor - think you can drop the `await` before any call to `.mockResolvedValue`.

a6software (approved):

packages/forms-web-app/tests/unit/services/department.service.test.js
frank-dodion: Oh. Yes. I think I copied from another example but it does no harm so I think I will leave it for refactoring :-)

PR #505, opened 2021-02-15
build: bump semantic-release to 17.3.9 to address dependabot alert

a6software:
## Ticket Number
UCD-1697

## Description of change
bump semantic-release to 17.3.9 to address dependabot alert - https://github.com/foundry4/appeal-planning-decision/security/dependabot/package-lock.json/marked/open

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,12752 @@
 {
   "name": "appeal-planning-decision",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.0.0",
+      "license": "ISC",
+      "devDependencies": {
+        "@commitlint/cli": "^11.0.0",
+        "@commitlint/config-conventional": "^11.0.0",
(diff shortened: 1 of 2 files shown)


Fenisay (approved):

a6software:
close, unwanted

marcfielding1:
I'm not sure this is unwanted, an outdated package shouldn't ever be ignored I don't think regardless of what it's *currently* used for. Can you reopen?

MrSimonEmms:
@marcfielding1 this isn't really used in the source code. Semantic Release is used to generate the version numbers from the commit messages, so it never goes anywhere near the production code - it's a dev dependency only

marcfielding1:
@MrSimonEmms Yeah I do understand that, I guess my point would be if we're delivering software to the client and dependency that's way of date sends the wrong message - also whilst *this* dep likely won't cause us any issues if we adopt that as a policy then eventually it could happen that we get outdated deps creeping in.

I do agree with you, but I disagree that we shouldn't fix it because it's outdated.

PR #510, opened 2021-02-15
fix(ucd-1698): fixed content on planning permission kick out page

rcmsoftwaredevelopment:
fixed content on planning permission kick out page

## Ticket Number
UCD-1698

## Description of change
fixed content on planning permission kick out page


Fenisay (approved):
LGTM

PR #521, opened 2021-02-16
Subtask/ucd 831 improve tests

Fenisay:
## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/lib/parse-date-string.js
+++ b/packages/appeals-service-api/src/lib/parse-date-string.js
@@ -1,8 +1,19 @@
-const { isDate, parseISO } = require('date-fns');
+const { isDate, isValid, parseISO } = require('date-fns');
+
+function isISODate(date) {
+  if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(date)) return false;
+  const d = new Date(date);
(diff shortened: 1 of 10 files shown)


MrSimonEmms (approved):

rcmsoftwaredevelopment (approved):

packages/appeals-service-api/src/lib/parse-date-string.js
```diff
@@ -1,8 +1,19 @@
...
+  if (originalValue === null || (isDate(originalValue) && isValid(originalValue))) {
+    return originalValue;
+  }
+
+  if (isISODate(originalValue)) {
+    return parseISO(originalValue);
   }
-  return null;
+
+  return new Error('Invalid Date or string not ISO format');
```
a6software, line 18: should this be `throw` instead of `return`?

Fenisay: Throw is breaking the whole validation stuff, what I could have done is just return the string message but it was weird. In the end it will be embedded in an error.

a6software: fair enough - maybe add this as a comment into the code to explain why - otherwise all good

a6software (approved):

PR #524, opened 2021-02-17
feat(appeals-service-api): add appeal to publisher queue on submit

frank-dodion:
## Ticket Number
UCD-1690

## Description of change
Add appeal to publisher queue on submission and prevent resubmission.


Fenisay (approved):
LGTM

PR #532, opened 2021-02-17
Subtask/ucd 831 refactor controllers

Fenisay:
## Ticket Number

UCD-831

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/src/controllers/appeals.js
```diff
@@ -18,199 +17,71 @@ module.exports = {
...
     logger.debug(`Creating appeal ${appeal.id} ...`);
-    try {
-      await mongodb
-        .get()
-        .collection('appeals')
-        .insertOne({ _id: appeal.id, uuid: appeal.id, appeal });
 
+    const document = await insertAppeal(appeal);
+
+    if (document.result.ok) {
```
a6software, line 23: ` if (document.result && document.result.ok) {`

// or 

` if (document.result?.ok) {`

think we can use optional chaining now

? idk maybe im being anal

packages/appeals-service-api/src/controllers/appeals.js
```diff
@@ -18,199 +17,71 @@ module.exports = {
...
-          res.status(200).send(doc.appeal);
-        })
-        .catch((err) => {
-          logger.warn(`Could not find appeal ${idParam}\n${err}`);
-          res.status(404).send(null);
-        });
-    } catch (err) {
-      logger.error(`Problem retrieving appeal ${idParam}\n${err}`);
-      res.status(500).send(`Problem finding appeal ${idParam}`);
+      const document = await getAppeal(idParam);
```
a6software, line 38: that's confusing - fn name of `getAppeal` shadows the fn on line 33

maybe

```
const { getAppeal: getAppealFromAppealApiService, updateAppeal, insertAppeal } = require('../services/appeal.service');

...

 const document = await getAppealFromAppealApiService(idParam);
```

idk - naming is probably wrong, but on initial read it looks like you're calling this again recursively

MrSimonEmms (approved):

packages/appeals-service-api/src/validators/appeals/appeals.validator.js
```diff
@@ -1,10 +1,21 @@
...
 const appealUpdateValidationRules = async (req, res, next) => {
   try {
     req.body = await updateAppeal.validate(req.body, { abortEarly: false });
+    logger.debug('Valid input format');
+
+    const appealId = req.body.id;
+    const pathId = req.params.id;
+
+    if (appealId && pathId !== appealId) {
+      next(ApiError.notSameId());
```
a6software, line 16: I may be wrong here but I would perhaps call `return;` after this line, or `return next(ApiError.notSameId());`

Reason being, and this shouldn't cause a bug here, but I believe it could lead to some unexpected results...

What will happen here (again, afaik) is that the `next();` on line 19 will *also* be called, but at a much later point in the middleware stack. I'm guessing that would be unexpected / unwanted, as the way I read this, the flow is that this middleware is expected to end here (?) 

Whilst this should cause no bugs here today, it might be confusing in the future if logic was added after this `if` statement, as that logic would still get run but later in the chain. That might cause some really weird behaviour.

Of course, I may be entirely wrong. 

https://stackoverflow.com/a/43404213 explains it better than I have.

packages/appeals-service-api/src/validators/appeals/appeals.validator.js
```diff
@@ -14,8 +25,22 @@ const appealUpdateValidationRules = async (req, res, next) => {
 const appealInsertValidationRules = async (req, res, next) => {
   try {
     req.body = await insertAppeal.validate(req.body, { abortEarly: false });
+    logger.debug('Valid input format');
+
+    const appealId = req.body.id;
+    const pathId = req.params.id;
+    if (await isAppealSubmitted(appealId)) {
+      logger.debug('Appeal is already submitted so end processing request with 409 response');
+      next(ApiError.appealAlreadySubmitted());
```
a6software, line 34: same issue as above, and potentially more serious this time (?)

frank-dodion:
I was able to create a new Appeal (POST), update it to SUBMITTED (PUT) and change it back to DRAFT (PATCH). I don't think this is expected behaviour.
Screenshot 2021-02-18 at 09 22 13
Screenshot 2021-02-18 at 09 21 44
Screenshot 2021-02-18 at 09 21 04

Fenisay:
> I was able to create a new Appeal (POST), update it to SUBMITTED (PUT) and change it back to DRAFT (PATCH). I don't think this is expected behaviour.
> Screenshot 2021-02-18 at 09 22 13
> Screenshot 2021-02-18 at 09 21 44
> Screenshot 2021-02-18 at 09 21 04

That's what I wanted to do at least, we could need to change the api after being submitted which is what @MrSimonEmms asked yesterday. The patch endpoint is not used by the forms.

packages/appeals-service-api/src/validators/appeals/appeals.validator.js
Fenisay: Made the change

a6software (approved):

frank-dodion:
> > I was able to create a new Appeal (POST), update it to SUBMITTED (PUT) and change it back to DRAFT (PATCH). I don't think this is expected behaviour.
> 
> That's what I wanted to do at least, we could need to change the api after being submitted which is what @MrSimonEmms asked yesterday. The patch endpoint is not used by the forms.
> 
> However you can see that the appeal is pushed on the queue only the first time we submit, which is the main reason I assumed that you added this rule.
> 
> ```
>       const isFirstSubmission = oldAppeal.state === 'DRAFT' && newAppeal.state === 'SUBMITTED';
> 
>       const updatedDocument = await updateAppeal(_.merge(oldAppeal, newAppeal), isFirstSubmission);
> ```

My understanding of the requirements was that two rules/behaviours were required i.e 1. Prevent modification of Appeal after it is SUBMITTED and 2. Place copy of Appeal on queue on submission. With the introduction of PATCH endpoint it became possible to 'Submit' and Appeal by either PUT or PATCH so both rules had to be applied to both endpoints.

PR #553, opened 2021-02-22
feat(pdf-service-api): initial implementation of the pdf service

frank-dodion:
## Ticket Number
AS-1038

## Description of change
This is an initial solution for the POST endpoint only. 
IT IS A WIP BUT IS REQUIRED TO PRACTICE DEPLOYMENTS.


packages/pdf-service-api/src/lib/healthchecks.js
```diff
@@ -0,0 +1,13 @@
+const { healthcheck } = require('@pins/common');
+const config = require('./config');
+const logger = require('./logger');
+
+const tasks = [{}];
```
MrSimonEmms, line 5: The object needs to be removed, so reads `[]` - I'm going to fix this as part of the Helm stuff and then you can rebase

MrSimonEmms (changes requested):

docker-compose.yml
```diff
@@ -216,6 +216,22 @@ services:
...
+    image: node:14-alpine
+    environment:
+      GOTENBERG_URL: http://gotenberg:4000
+      LOGGER_LEVEL: "debug"
+      SERVER_SHOW_ERRORS: "true"
+    ports:
+      - 3003:3000
+    working_dir: /opt/app
+    links:
+      - appeals-service-api
```
MrSimonEmms, line 229: Does this need a connection to the `appeals-service-api`? If so, we'll need an envvar for it.

Also, this will need `gotenberg` adding else it won't be able to connect via the internal DNS

MrSimonEmms (changes requested):

docker-compose.yml
```diff
@@ -249,4 +265,3 @@ services:
       DISABLE_UNOCONV: '0'
```
MrSimonEmms, line 265: This needs changing to `1`

MrSimonEmms (changes requested):

docker-compose.yml
frank-dodion: The pdf service does not need to call out to the appeal api service. i have replaced that line with '- gotenberg' as you suggested but everything seemed to be working ok in local docker previously.

docker-compose.yml
frank-dodion: Done.

packages/pdf-service-api/src/lib/healthchecks.js
frank-dodion: I can see you have fixed this. Thanks.

MrSimonEmms (approved):
Looks ok and I've confirmed it works as part of my ticket on #558 - happy for you to wait for additional review if you want, or blame me :)

frank-dodion:
> Looks ok and I've confirmed it works as part of my ticket on #558 - happy for you to wait for additional review if you want, or blame me :)

Any port in a storm will do :-)

PR #574, opened 2021-02-24
Subtask/as 1030 submission accuracy validation

davehlewis:
## Ticket Number

AS-1030

## Description of change
Added validator to question as per AS-2 instructions

## Checklist

- [X] New features have tests

## Important

### DIFF
--- a/packages/lpa-questionnaire-web-app/src/controllers/accuracy-submission.js
+++ b/packages/lpa-questionnaire-web-app/src/controllers/accuracy-submission.js
@@ -7,3 +7,26 @@ exports.getAccuracySubmission = (req, res) => {
     backLink: req.session.backLink || `/${req.params.id}/${VIEW.TASK_LIST}`,
   });
 };
+
+exports.postAccuracySubmission = async (req, res) => {
+  const { body } = req;
+  const { errors = {}, errorSummary = [] } = body;
+
+  const values = {
+    'accurate-submission': body['accurate-submission'],
+    'inaccuracy-reason': body['inaccuracy-reason'],
+  };
+
+  if (Object.keys(errors).length > 0) {
+    res.render(VIEW.ACCURACY_SUBMISSION, {
+      appeal: getAppealSideBarDetails(req.session.appeal),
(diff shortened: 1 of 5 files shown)


packages/lpa-questionnaire-web-app/src/controllers/accuracy-submission.js
```diff
@@ -7,3 +7,26 @@ exports.getAccuracySubmission = (req, res) => {
     backLink: req.session.backLink || `/${req.params.id}/${VIEW.TASK_LIST}`,
   });
 };
+
+exports.postAccuracySubmission = async (req, res) => {
+  const { body } = req;
+  const { errors = {}, errorSummary = [] } = body;
```
swdpcomputing, line 13: errors and errorSummary used once. Do we need a variable?

packages/lpa-questionnaire-web-app/tests/unit/validators/accuracy-submission.test.js
```diff
@@ -0,0 +1,113 @@
...
+      expect(rule.stack[1].validator.name).toEqual('isEmpty');
+      expect(rule.stack[1].negated).toBeTruthy();
+      expect(rule.stack[1].message).toEqual(
+        'Enter details of why this does not accurately reflect the planning application'
+      );
+    });
+  });
+
+  describe('validator', () => {
+    [
```
swdpcomputing, line 46: This test could be clearer if large examples are in separate files

davehlewis: Given the shared modular nature of this test, I think it makes sense. As we are doing single questions this structure is ok, if pushing it a little. Plus it keeps us consistent with the other team.

packages/lpa-questionnaire-web-app/src/controllers/accuracy-submission.js
davehlewis: They get used more later, so it's sort of planning ahead, knowing what the final version looks like

swdpcomputing (approved):

PR #580, opened 2021-02-24
Subtask/as 1206 fwa cookie consent

a6software:
## Ticket Number

AS-

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/views/includes/cookie-banner.njk
```diff
@@ -1,5 +1,20 @@
-<div class="app-cookie-banner">
-  <p class="app-cookie-banner__message">
-    {{cookieText | safe }}
-  </p>
-</div>
+{% from "../macros/cookie-banner.njk" import cookieBanner %}
+{% from "../macros/cookie-banner-accepted.njk" import cookieBannerAccepted %}
+{% from "../macros/cookie-banner-rejected.njk" import cookieBannerRejected %}
+
+{% set didNotAcceptOrRejectCookiePolicyOnPreviousRequest = res_locals.acceptedCookiePolicy !== true and res_locals.acceptedCookiePolicy !== false %}
```
a6software, line 5: am I having a brain fart here, or is there no way to check an object is defined and explicitly `false` in a nunjucks template? I'm sure I have solved this problem before.

The issue I face(d) is that nunjucks can't seem to reliably determine if an object is undefined or false. It sees them as one and the same, which feels wrong and weird, and like I'm probably doing something wrong somehow.

PR #613, opened 2021-03-02
feat(forms-web-app): generate pdf and store it document store

Fenisay:
## Ticket Number

AS-1040

## Checklist

- [x] If adding or remove environment variables (e.g. in `docker-compose.yaml`) then I have updated the appropriate Helm chart
- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/config.js
+++ b/packages/forms-web-app/src/config.js
@@ -27,6 +27,9 @@ module.exports = {
     timeout: Number(process.env.DOCUMENTS_SERVICE_API_TIMEOUT || 10000),
     url: process.env.DOCUMENTS_SERVICE_API_URL,
   },
+  pdf: {
+    url: process.env.PDF_SERVICE_API_URL,
+  },
   fileUpload: {
     debug: process.env.FILE_UPLOAD_DEBUG === 'true',
     pins: {
--- a/packages/forms-web-app/src/controllers/appellant-submission/submission.js
+++ b/packages/forms-web-app/src/controllers/appellant-submission/submission.js
@@ -1,4 +1,5 @@
-const logger = require('../../lib/logger');
+const { storePdfAppeal } = require('../../services/pdf.service');
+
 const { VIEW } = require('../../lib/views');
 const { createOrUpdateAppeal } = require('../../lib/appeals-api-wrapper');
 
@@ -23,8 +24,9 @@ exports.postSubmission = async (req, res) => {
       const { appeal } = req.session;
       appeal.state = 'SUBMITTED';
       req.session.appeal = await createOrUpdateAppeal(appeal);
+      await storePdfAppeal(appeal);
     } catch (e) {
-      logger.error(e);
+      req.log.error(e);
       res.render(VIEW.APPELLANT_SUBMISSION.SUBMISSION, {
         errors,
         errorSummary: [{ text: e.toString(), href: '#' }],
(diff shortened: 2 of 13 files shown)


Fenisay:
The e2e-tests are failing because the lpa is missing in the appeal. I think it would be better to see AS 102 to be merged to use the common method to fulfill an appeal. However the rest is ready to be reviewed and any feedback is welcome.

packages/forms-web-app/src/config.js
```diff
@@ -27,6 +27,10 @@ module.exports = {
     timeout: Number(process.env.DOCUMENTS_SERVICE_API_TIMEOUT || 10000),
     url: process.env.DOCUMENTS_SERVICE_API_URL,
   },
+  pdf: {
+    timeout: Number(process.env.DOCUMENTS_SERVICE_API_TIMEOUT || 10000),
```
a6software, line 31: Maybe duplicate this and rename so it can be controlled separately? 

`PDF_SERVICE_API_TIMEOUT` ?

idk, maybe not needed.

packages/forms-web-app/src/controllers/appellant-submission/submission.js
```diff
@@ -23,6 +25,7 @@ exports.postSubmission = async (req, res) => {
       const { appeal } = req.session;
       appeal.state = 'SUBMITTED';
       req.session.appeal = await createOrUpdateAppeal(appeal);
+      await storePdfAppeal(appeal);
     } catch (e) {
       logger.error(e);
```
a6software, line 30: Simon's meeting on Friday afternoon was about moving to `req.logger.{method}`, as that way we can track an error from start to finish using the correlation id.

packages/forms-web-app/src/lib/documents-api-wrapper.js
```diff
@@ -20,7 +20,12 @@ exports.createDocument = async (appeal, formData) => {
   let apiResponse;
   try {
     const fd = new FormData();
-    fd.append('file', fs.createReadStream(formData.tempFilePath), formData.name);
+
+    if (typeof data === 'object') {
+      fd.append('file', fs.createReadStream(data.tempFilePath), data.name);
+    } else {
```
a6software, line 26: One thing I hit upon when doing that original line was that sometimes I'd get an 0 byte PDF created.

I'd imagine:

```
const data = undefined;
fd.append('file', data, `${appeal.id}.pdf`);
```

would create a 0 byte PDF. 

Maybe worth being a bit more defensive here? 

```
else if (typeof data === 'string' || data instanceof String) {
//...
} else {
throw?
}

File: packages/forms-web-app/src/services/pdf.service.js [hunk 4]
```diff
@@ -0,0 +1,33 @@
+const fetch = require('node-fetch');
+const { createDocument } = require('../lib/documents-api-wrapper');
+const { generatePDF } = require('../lib/pdf-api-wrapper');
+const { VIEW } = require('../lib/views');
+
+const getHtmlAppeal = async (appeal) => {
+  const response = await fetch(
+    `http://forms-web-app:3000/${VIEW.APPELLANT_SUBMISSION.SUBMISSION_INFORMATION}/${appeal.id}`
```
a6software, line 8: hardcoded to `http://forms-web-app:3000` - will that work on k8s? (I have no idea btw)

Fenisay: I think so as there is internal services

PR #619, opened 2021-03-02
feat(as-1222): cookie view preferences glue code

rcmsoftwaredevelopment:
cookie view preferences glue code

## Ticket Number
AS-100

## Description of change
glue code for view cookie preferences web page, and feature file as I messed up original branch

### DIFF
--- a/e2e-tests/cypress/integration/cookie-consent-view-preferences/cookie-consent-view-preferences.js
+++ b/e2e-tests/cypress/integration/cookie-consent-view-preferences/cookie-consent-view-preferences.js
@@ -1,0 +1,14 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps';
+import { getTask } from '../common/task';
+
+Given('a user has elected to manage their cookie preference', () => {
+  cy.goToHouseholderQuestionPage();
+});
+
+When('the user views the Cookie Preferences service', () => {
+  cy.goToCookiePreferencesPage()
+});
+
+Then('the user is provided information of what task each cookie is performing', () => {
+  cy.userIsNavigatedToPage('cookie');
+});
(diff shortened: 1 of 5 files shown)


MrSimonEmms (approved):

a6software (approved):

e2e-tests/cypress/integration/cookie-consent-view-preferences/cookie-consent-view-preferences.js
```diff
@@ -0,0 +1,14 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps';
+import { getTask } from '../common/task';
+
+Given('a user has elected to manage their cookie preference', () => {
+  cy.goToHouseholderQuestionPage();
+});
+
+When('the user views the Cookie Preferences service', () => {
+  cy.goToCookiePreferencesPage()
```
danWatkinson, line 9: rough edge, but i get why it's here.
In theory the AC is wanting this to be driven by the cookie banner, so this step should really be some `cy.clickManageCookiePreferences()` that makes the user click on the banner.
however- the banner doesn't exist yet; this story exists as is to get you into the swing of front-end stuff, and i don't think it's very useful to get hung up on that in this PR

what i would say is- there are stories coming up that involve finishing the plumbing on this page once chris has the main cookie story into master. As part of that story i would like to see this step corrected to use the cookie banner rather than a direct 'visit'.

a6software: It's a bit of a catch 22 as I've added the test in one of my other PR's - https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/597/files#diff-92c4d67f61512fa26e8f78fef4775d1c57525e7cb055ce4ccfa3da3ff4364a79R13

but I can't uncomment it without this one, as if I do so that PR build fails.

PR #621, opened 2021-03-02
test(e2e-tests): add step defs

frank-dodion:
## Ticket Number
AS-1191

## Description of change
Add step defs


danWatkinson (approved):
nice- like the moving of things into fixtures; i still have designs on turning all these json's into a builder object that lets us fluidly build this stuff.. but 100% prefer the well named files > ticket-related stuff

PR #645, opened 2021-03-05
feat(lpa-questionnaire-web-app): add view for development plan as-19

swdpcomputing:
## Ticket Number
AS-1301

## Description of change
add view for development plan as-19

## Checklist

- [x] New features have tests

## Important


packages/lpa-questionnaire-web-app/src/views/development-plan.njk
```diff
@@ -0,0 +1,118 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <form action="" method="POST" novalidate>
+        <span class="govuk-caption-l">
+          <span class="govuk-visually-hidden">Section. </span>
+          Optional supporting documents
+        </span>
+
+        {% call govukFieldset({
+          legend: {
+            text: "Development Plan Document or Neighbourhood Plan",
```
davehlewis, line 46: This is non-standard. The heading ideally would come from the Radios, with the text in the paragraph as a hint. This would change the design slightly but be more in line with GDS specs. Check with @pippalou on this I think?

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
```diff
@@ -0,0 +1,118 @@
...
+            classes: "govuk-fieldset__legend--l",
+            isPageHeading: true
+          }
+        }) %}
+        <p>Has any Development Plan Document or Neighbourhood Plan relevant to this appeal been submitted for examination (or in the case of a Neighbourhood Plan, been examined and is awaiting a referendum)?</p>
+
+        {% set developmentPlanTextArea %}
+          {{ govukTextarea({
+            hint: {
```
davehlewis, line 55: Looking at the design this should be the label, not the hint. It is missing a label, and hint results in different styling

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
```diff
@@ -0,0 +1,118 @@
...
+            errorMessage: errors['plan-changes-text'] and {
+                text: errors['plan-changes-text'].msg,
+                attributes: {
+                  'data-cy': 'plan-changes-error'
+                }
+            }
+          }) }}
+        {% endset %}
+
+        {{ govukRadios({
```
davehlewis, line 73: This is missing a label - think we will have GDS issues without one

davehlewis (changes requested):

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
swdpcomputing: In the preview there's only labels for the yes and the no in the component. Do we have to look into changing the component appearance from the spec?

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
swdpcomputing: Makes sense. Edited.

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
swdpcomputing: Removed and put code under the radio buttons object as a hint.

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
swdpcomputing: Changed with previous fix

PR #649, opened 2021-03-05
feat(forms-web-app): setup routing for guidance pages

davehlewis:
## Ticket Number

AS-1332

## Description of change
- Setup routing for guidance pages and update index route to go to Before Appeals page
- Created new template for guidance pages
- Added content for Before Appeals page

## Checklist

- [X] New features have tests

## Important


packages/forms-web-app/src/controllers/guidance-pages.js
```diff
@@ -0,0 +1,13 @@
...
+
+exports.getBeforeAppeal = (_, res) => {
+  res.render(VIEW.GUIDANCE_PAGES.BEFORE_APPEAL, {
+    currentUrl: '/before-you-appeal',
+    heading: 'Appeal a householder planning decision',
+    nextPage: {
+      text: 'When you can appeal',
+      url: '/when-you-can-appeal',
+    },
+    title: 'Before you appeal - Appeal a householder planning decision - GOV.UK',
```
a6software, line 11: are these things being set this way due to us also needing this page in the questionnaire project / somewhere else?

davehlewis: The guidance pages are used 4 or 5 times in the project, so made it this way to make it more easily reusable

packages/forms-web-app/src/routes/index.js
```diff
@@ -6,8 +6,10 @@ const appellantSubmissionRouter = require('./appellant-submission');
 const eligibilityRouter = require('./eligibility');
 const homeRouter = require('./home');
 const cookieRouter = require('./cookies');
+const guidancePagesRouter = require('./guidance-pages');
 
 router.use('/', homeRouter);
+router.use('/', guidancePagesRouter);
```
a6software, line 12: Personal preference - could this be changed to:

`router.use(guidancePagesRouter);`

To me (and again, appreciate personal opinion), but this reads like there's two routers handling the same path, which feels like the second router shouldn't work. Appreciate it does in express, so feel free to ignore this. Some frameworks wouldn't like this ... and those are the ones I come from :D 

Again, feel free to ignore.

davehlewis: I mainly went for consistency but I can see the benefit of it. Will update.

rcmsoftwaredevelopment (approved):

PR #658, opened 2021-03-08
feat(lpa-questionnaire-web-app): add view for development page

swdpcomputing:
## Ticket Number

AS-1301

## Description of change

add view for development plan

## Checklist

- [x] New features have tests

## Important


packages/lpa-questionnaire-web-app/src/controllers/development-plan.js
```diff
@@ -0,0 +1,24 @@
+const { VIEW } = require('../lib/views');
+const getAppealSideBarDetails = require('../lib/appeal-sidebar-details');
+
+exports.getDevelopmentPlan = (req, res) => {
+  const developmentPlanSection =
+    req.session.appealReply.optionalDocumentsSection.developmentOrNeighbourhood;
```
marcfielding1, line 6: What happens if this doesn't exist? (By the way I'm just typing comments, they're not meant to be terse, just wizzing through it.)

marcfielding1: ie that's a very deep object reference so it relies on 5 objects existing.

marcfielding1: It may be the API just handles it and returns 500 if we're happy with that, but ideally, if the params are wrong then 400 I guess maybe?

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
```diff
@@ -0,0 +1,121 @@
...
+{% from "govuk/components/fieldset/macro.njk" import govukFieldset %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% from "components/appeal-details/macro.njk" import appealDetails %}
+{% from "govuk/components/details/macro.njk" import govukDetails %}
+{% from "govuk/components/textarea/macro.njk" import govukTextarea %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    text: "Back",
+    href: backLink,
```
marcfielding1, line 16: @davehlewis Excuse my ignorance, is this the right way to still do a backlink?

packages/lpa-questionnaire-web-app/src/controllers/development-plan.js
swdpcomputing: Passes the tests, and that object is set explicitly

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
swdpcomputing: Based from Dave's backlink.

packages/lpa-questionnaire-web-app/src/controllers/development-plan.js
marcfielding1: I suppose my point is we're assuming this route has only the web application as the consumer, which is MVP I guess.

marcfielding1 (approved):

packages/lpa-questionnaire-web-app/src/controllers/development-plan.js
davehlewis: I think Marc has a point here. It is something I regret not adding in the earlier section. Simple optional chaining would help, with some defaults when destructuring. If the model changes mid review then it will cause errors with how the data is structured and we don't want that

packages/lpa-questionnaire-web-app/src/views/development-plan.njk
davehlewis: Yep that's the right way. backLink is a session variable set by the appropriate page. It falls back to task list page. While not necessarily the "right" way, it is "a" way that works right now

PR #744, opened 2021-03-15
Subtask/as 1488

MrSimonEmms:
## Ticket Number

AS-1488

## Description of change

Configure the envvars for the Notify service

## Checklist

- [ ] New features have tests

## Important


charts/app/values.yaml
```diff
@@ -27,6 +27,11 @@ appealsServiceApi:
   config:
     port: 3000
     logLevel: trace
+    notify:
+      baseUrl: https://api.notifications.service.gov.uk
+      secretName: ""
+      serviceId: dummy-service-id-for-notify
```
MrSimonEmms, line 33: @frank-dodion I assume I need to ask PINS for the correct value for this?

charts/app/values.yaml
```diff
@@ -27,6 +27,11 @@ appealsServiceApi:
...
+      templateId: 15ed37a9-506c-4845-88ea-95502282a863
```
MrSimonEmms, line 34: @frank-dodion is this the correct value for dev/preprod/prod?

MrSimonEmms (commented):
@MrSimonEmms add in the Docs Service config

Fenisay (approved):
Do you confirm that there is no dev key and same with dev and preprod ?

MrSimonEmms:
> Do you confirm that there is no dev key and same with dev and preprod ?

I hereby confirm and solemnly swear that "yeah, think so."

It does seem that the PINS strategy for dealing with external services is to have a prod and a preprod one. This is about their first project with a dev stack, so we're defaulting to using the preprod one in dev

charts/app/values.yaml
MrSimonEmms: Oh, so the `serviceId` is also a sensible default that we can override. I'll make that change. Thanks m'colleague

rcmsoftwaredevelopment (approved):

charts/app/values.yaml
frank-dodion: Yes.

charts/app/values.yaml
frank-dodion: I think we do not need serviceId because it can be derived from a valid apiKey.

frank-dodion (approved):

PR #784, opened 2021-03-19
Fix/notify service error logging

MrSimonEmms:
## Ticket Number

AS-

## Description of change

- Improve logging of Notify service in line with [Axios recommendations](https://github.com/axios/axios#handling-errors)
- Deploy to prod

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/lib/notify.js
+++ b/packages/appeals-service-api/src/lib/notify.js
@@ -78,7 +78,28 @@ async function sendEmail(appeal) {
     logger.debug({ options }, 'options');
     await notifyClient.sendEmail(templateId, destinationEmail, options);
   } catch (err) {
-    logger.error({ err }, 'Problem sending email');
+    const message = 'Problem sending email';
+    if (err.response) {
+      logger.error(
+        {
+          message: err.message,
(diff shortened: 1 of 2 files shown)


Fenisay (approved):

packages/appeals-service-api/src/lib/notify.js
```diff
@@ -78,7 +78,28 @@ async function sendEmail(appeal) {
...
-    logger.error({ err }, 'Problem sending email');
+    const message = 'Problem sending email';
+    if (err.response) {
+      logger.error(
+        {
+          message: err.message,
+          data: err.response.data,
+          status: err.response.status,
+          headers: err.response.headers,
+        },
```
a6software, line 89: any use to spreading the rest of `err` in the `if` / `else if` flows?

MrSimonEmms: Possibly. It's the same as in the all the Horizon functions so I have confidence that this works - I don't know what else is in that `response` object that shouldn't be in the logs

frank-dodion (approved):

PR #787, opened 2021-03-19
Disable ACP in Prod

MrSimonEmms:
Reverts Planning-Inspectorate/appeal-planning-decision#786


Simeon-techbss:
We will wait until after Stand up on Monday 22nd before proceeding with this.

Simeon-techbss:
Based on confirmation from both Ed and Paul M - I am agreeing to remove the ACP flag and return production to a full end to end service - we are aware that at this time there is still an issue with Horizon.

Simeon-techbss (approved):

PR #790, opened 2021-03-22
Subtask/as 1537 non js upload

davehlewis:
## Ticket Number

AS-1537

## Description of change
- Added support for non-js version of file upload
- Added post function with support for upload, delete, and submit. Upload and delete are in event of JS being disabled
- Refactor file upload to template as reused a lot
- move to add non-js solution added some changes to the file endpoints. Added missing tests for these also
- All util functions for files now in a single file for ease

Delete handling for saved files and file upload to doc store are to do in another subtask

Note: commits will be squashed in feature file

## Checklist

- [X] New features have tests

## Important


packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
```diff
@@ -8,3 +9,80 @@ exports.getUploadPlans = (req, res) => {
...
+
+exports.postUploadPlans = async (req, res) => {
+  const documents = req.body?.files?.documents || [];
+
+  const { delete: deleteFile = '', errors = {}, submit = '' } = req.body;
+
+  // Chance for delete to be triggered due to non-JS solution. delete will be set to value of filename if button clicked
+  if (deleteFile) {
+    // TODO: add handling for deletion from DB and doc store as part of AS-1538
+    const file = req.session.uploadedFiles?.find((upload) => upload.name === deleteFile);
```
a6software, line 21: I'm curious on this, can't seem to find an immediate answer.

As best I understand it, with the questionnaire the user can return to the site at a later date and resume - using an id in the URL. 

What happens with previously uploaded files in this instance?

I'm guessing it works because the session is long lived(?)

However, I'm also thinking we should be deleting the session when the user finally submits at the end. Now, weirdly, the code I added in to do this for FWA has been removed, which I'm guessing is because during dev / debug, it's a PITA, but that's a different issue. 

Just curious as to how that might play out? Maybe it's not even an issue.

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
```diff
@@ -8,3 +9,80 @@ exports.getUploadPlans = (req, res) => {
...
+
+    if (file) {
+      req.session.uploadedFiles = req.session.uploadedFiles.filter(
+        (upload) => upload.name !== deleteFile
+      );
+    }
+  } else if (Array.isArray(documents) && documents.length) {
```
a6software, line 28: Looks like `Array.isArray(documents)` would always be true due to line 14?

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
```diff
@@ -8,3 +9,80 @@ exports.getUploadPlans = (req, res) => {
...
+      ? req.session.uploadedFiles.reduce((errorsOutput, file) => {
+          if (file.error)
+            errorsOutput.push({
+              href: `#${file.name}`,
+              text: file.error,
+            });
+          return errorsOutput;
+        }, [])
+      : []),
+  ];
```
a6software, line 61: wondering if this could be moved off to a separate function, as this controller is > 80 lines at this point, and seemingly going to get bigger due to line 63

a6software (approved):
Only 1 optional suggestion from my pov, happy to merge as is either way.

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
```diff
@@ -8,3 +9,80 @@ exports.getUploadPlans = (req, res) => {
...
+  if (!submit || constructedErrorSummary.length) {
+    res.render(VIEW.UPLOAD_PLANS, {
+      appeal: getAppealSideBarDetails(req.session.appeal),
+      backLink: req.session.backLink || `/${req.params.id}/${VIEW.TASK_LIST}`,
+      errorMessage,
+      errorSummary: constructedErrorSummary,
+      uploadedFiles: req.session.uploadedFiles?.map((doc) => ({
+        deleteButton: {
+          text: 'Delete',
+        },
```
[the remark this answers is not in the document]
davehlewis: I agree, however sadly this is how the MOJ template expects the object. Strange really but works

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
davehlewis: I'm with you, missing some context. The session is used far less than that - as soon as someone hits save and continue it writes to the doc store. If they leave the question and reopen, it clears unsaved docs by way of middleware. In this way it neatly avoids the issue.

The next ticket handles the upload and saving. Deletion of a document from that point will delete from the doc store.

Hope that makes sense?

swdpcomputing (approved):

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
davehlewis: Good catch, line 14 change happened later than this line

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
davehlewis: I'm considering modularising the whole function as the next task, given it is generic and will be reused about a dozen times. Happy for this to be part of a different task or this PR?

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
a6software: Yes, thanks for the explanation - much appreciated.

packages/lpa-questionnaire-web-app/src/controllers/upload-plans.js
a6software: Sounds good - surprised we haven't had to do the JS variant yet on upload. Wonder if that's been forgotten actually. I'll raise it.

PR #791, opened 2021-03-22
feat(forms-web-app): added non-js fallback for LPD selection

frank-dodion:
## Ticket Number
AS-1620

## Description of change
Added non-js fallback for LPD selection


Fenisay (approved):
LGTM

PR #832, opened 2021-03-25
feat(forms-web-app): update page title for error state

frank-dodion:
AS-1706 and AS-1707

## Description of change
Update page title for error state for householder planning permission eligibility question. Step def updated to check page title on error condition scenario.


Fenisay (approved):
LGTM

PR #919, opened 2021-04-06
Fix/as 115 backlink without js

a6software:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-115

## Description of change
Back link should work when Javascript is disabled

PR is done, but a bug has been spotted during testing with an indirect dependency. AS-1854

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/src/views/layouts/main.njk
```diff
@@ -41,7 +41,7 @@
       {% block backButton %}
         {{ govukBackLink({
           text: "Back",
-          href: "javascript:history.back()",
+          href: navigation[1],
```
a6software, line 44: the middleware ensures there will always be at least two items in the navigation history - the current page, and the fallback page.

There is an edge case where this would break, and that would be if we showed the back button on the fall back page. This should never happen, though if it ever does, it should be caught by this test:

https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/919/files#diff-2f8b09958305a61fd225779a562fa8dcd065b8572ae31fe6f072e36282c4fcf1R74

e2e-tests/cypress/integration/eligibility-listed-building-status.feature
```diff
@@ -2,15 +2,16 @@
 Feature: A prospective appellant states whether or not their appeal covers a listed building
     Our service doesn't cover listed buildings.
 
-  @as-1670
-  Scenario: Listed Building statement is required
-    When the user does not provide a Listed Building statement
-    Then the user is informed that a Listed Building statement is required
+#  @as-1670
```
a6software, line 5: this should not be commented out

rcmsoftwaredevelopment (approved):

a6software:
ty

PR #929, opened 2021-04-08
fix(forms-web-app): controller, validator, route + unit tests for eligibility listed building

a6software:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-1858

## Description of change
Fixes issue whereby given answer was not being saved or redisplayed on the front end

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/eligibility/listed-building.js
+++ b/packages/forms-web-app/src/controllers/eligibility/listed-building.js
@@ -1,46 +1,61 @@
 const { VIEW } = require('../../lib/views');
-
-const FORM_FIELD = {
-  'is-your-appeal-about-a-listed-building': {
-    id: 'is-your-appeal-about-a-listed-building',
-    items: [
-      {
-        value: 'yes',
-        text: 'Yes',
-      },
-      {
-        value: 'no',
-        text: 'No',
-      },
-    ],
-  },
-};
-
-exports.FORM_FIELD = FORM_FIELD;
+const { validIsListedBuildingOptions } = require('../../validators/eligibility/listed-building');
+const { createOrUpdateAppeal } = require('../../lib/appeals-api-wrapper');
 
 exports.getServiceNotAvailableForListedBuildings = (req, res) => {
   res.render(VIEW.ELIGIBILITY.LISTED_OUT);
 };
 
 exports.getListedBuilding = (req, res) => {
-  res.render(VIEW.ELIGIBILITY.LISTED_BUILDING, { FORM_FIELD });
+  res.render(VIEW.ELIGIBILITY.LISTED_BUILDING, { appeal: req.session.appeal });
 };
 
-exports.postListedBuilding = (req, res) => {
+exports.postListedBuilding = async (req, res) => {
   const { body } = req;
-
   const { errors = {}, errorSummary = [] } = body;
+
+  const { appeal } = req.session;
+
+  let isAppealAboutAListedBuilding = null;
+  if (validIsListedBuildingOptions.includes(req.body['is-your-appeal-about-a-listed-building'])) {
+    isAppealAboutAListedBuilding = req.body['is-your-appeal-about-a-listed-building'] === 'yes';
+  }
 
   if (Object.keys(errors).length > 0) {
     res.render(VIEW.ELIGIBILITY.LISTED_BUILDING, {
+      appeal: {
+        ...appeal,
+        eligibility: {
+          ...appeal.eligibility,
+          isListedBuilding: isAppealAboutAListedBuilding,
+        },
+      },
       errors,
       errorSummary,
-      FORM_FIELD,
     });
     return;
   }
 
-  if (body['is-your-appeal-about-a-listed-building'] === 'yes') {
+  try {
+    req.session.appeal = await createOrUpdateAppeal({
+      ...appeal,
+      eligibility: {
+        ...appeal.eligibility,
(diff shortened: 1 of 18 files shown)


a6software:
I think this might fail until the appeal service api change is in (?)

packages/forms-web-app/src/controllers/eligibility/listed-building.js
```diff
@@ -1,46 +1,62 @@
...
     });
     return;
   }
 
-  if (body['is-your-appeal-about-a-listed-building'] === 'yes') {
+  try {
+    req.session.appeal = await createOrUpdateAppeal({
+      ...appeal,
+      eligibility: {
+        ...appeal.eligibility,
```
swdpcomputing, line 44: Does this duplicate code entered in line 42? Not duplicating is probably more inconvenient...

packages/forms-web-app/src/controllers/eligibility/listed-building.js
```diff
@@ -1,46 +1,62 @@
...
+        isListedBuilding: isAppealAboutAListedBuilding,
+      },
+    });
+  } catch (e) {
+    logger.error(e);
```
swdpcomputing, line 49: Is this in keeping with the new logging conventions recently updated in Confluence?

packages/forms-web-app/tests/unit/controllers/eligibility/listed-building.test.js
```diff
@@ -19,54 +35,100 @@ describe('controllers/eligibility/listed-building', () => {
...
+          'is-your-appeal-about-a-listed-building': 'bad value',
+          errors: { a: 'b' },
+          errorSummary: [{ text: 'There were errors here', href: '#' }],
+        },
       };
-      const mockResponse = {
-        redirect: jest.fn(),
+      await listedBuildingController.postListedBuilding(mockRequest, res);
+
+      expect(createOrUpdateAppeal).not.toHaveBeenCalled();
```
swdpcomputing, line 56: This is not testing re-rendering, should this be in it's own explicitly named test?

packages/forms-web-app/tests/unit/controllers/eligibility/listed-building.test.js
```diff
@@ -19,54 +35,100 @@ describe('controllers/eligibility/listed-building', () => {
...
+
+      expect(res.redirect).not.toHaveBeenCalled();
```
swdpcomputing, line 58: As last comment

packages/forms-web-app/tests/unit/controllers/eligibility/listed-building.test.js
```diff
@@ -19,54 +35,100 @@ describe('controllers/eligibility/listed-building', () => {
...
+        body: {},
       };
 
-      listedBuildingController.postListedBuilding(mockRequest, mockResponse);
+      const error = new Error('Cheers');
+      createOrUpdateAppeal.mockImplementation(() => Promise.reject(error));
+
+      await listedBuildingController.postListedBuilding(mockRequest, res);
+
+      expect(res.redirect).not.toHaveBeenCalled();
```
swdpcomputing, line 83: As last comment

rcmsoftwaredevelopment (approved):

packages/forms-web-app/src/controllers/eligibility/listed-building.js
a6software: Line 44 is required or the existing key / values inside `eligibility` would be lost.

Perhaps better illustrated with a screenshot:

Screenshot from 2021-04-12 08-56-28

`a` is the original object

`x` is a wholesale copy

`y` is a copy that redefines its own `b`, but in doing so we lose the contents of `a.b`.

`p` demonstrates the additional spread (like line 44) which retains everything that exists in `a.b` and adds in new stuff, too.

packages/forms-web-app/src/controllers/eligibility/listed-building.js
a6software: I believe I have now resolved this. Thanks for flagging!

packages/forms-web-app/tests/unit/controllers/eligibility/listed-building.test.js
a6software: These steps (this and the other two comments that relate) are testing the function calls that happen before the re-rendering. If I split them out I would have something like this:

(assume we have five steps in the controller)

test 1:

1. controller call
2. sub func 1 is called

test 2:

1. controller call
2. sub func 1 is called
3. sub func 2 is called

test 2:

1. controller call
2. sub func 1 is called
3. sub func 2 is called
4. sub func 3 is called

test 2:

1. controller call
2. sub func 1 is called
3. sub func 2 is called
4. sub func 3 is called
5. sub func 4 is called

test 2:

1. controller call
2. sub func 1 is called
3. sub func 2 is called
4. sub func 3 is called
5. sub func 4 is called
6. sub func 5 is called

Seems like a lot of repeated tests to get to the same outcome. 

But equally, I feel like I have misunderstood?

rcmsoftwaredevelopment (approved):

a6software:
I am merging this as it's blocking two tickets and I am happy this approach follows the existing approach re: the three test comments.

@swdpcomputing I will address any feedback under a new PR as needed.

swdpcomputing:
LGTM

a6software:
thanks!

PR #984, opened 2021-04-20
feat: as-1903; session management for LPA questionnaire

danWatkinson:
## Ticket Number
https://pins-ds.atlassian.net/browse/AS-1903

## Description of change
pinning down the session behaviour for LPA questionnaire

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/lpa-submissions-e2e-tests/cypress.json
+++ b/lpa-submissions-e2e-tests/cypress.json
@@ -7,8 +7,8 @@
   "env": {
     "demoDelay": 0,
     "DOCUMENT_SERVICE_BASE_URL": "http://localhost:3001/api/v1",
-    "ASSUME_LIMITED_ACCESS": false,
-    "APPEAL_REPLY_SERVICE_BASE_URL": "http://localhost:3002"
+    "APPEAL_REPLY_SERVICE_BASE_URL": "http://localhost:3002/api/v1/reply",
+    "APPEAL_SERVICE_BASE_URL": "http://localhost:3000/api/v1/appeals"
   },
   "retries": {
     "runMode": 1,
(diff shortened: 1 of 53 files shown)


lpa-submissions-e2e-tests/cypress.json
```diff
@@ -8,7 +8,8 @@
     "demoDelay": 0,
     "DOCUMENT_SERVICE_BASE_URL": "http://localhost:3001/api/v1",
     "ASSUME_LIMITED_ACCESS": false,
-    "APPEAL_REPLY_SERVICE_BASE_URL": "http://localhost:3002"
+    "APPEAL_REPLY_SERVICE_BASE_URL": "http://localhost:3002",
```
danWatkinson, line 11: mental note - agree these should have the /api/v1 bit baked in at config //todo

lpa-submissions-e2e-tests/cypress/fixtures/anAppeal.json
```diff
@@ -0,0 +1,93 @@
+{
```
danWatkinson, line 1: stole an arbitrary appeal from team1's e2e tests; there would likely be an advantage to testing more than one appeal type at some stage...

davehlewis: for consistency this the data that is seeded: https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/feature/as-1903-lpa-sessions/data/data/appeals-service-api/001-appeals.json

danWatkinson: in all honesty we should have a bunch of these that we feed in to different tests to triangulate with.. if you check the assertions somewhere down below against the 'appeal sidebar' we're checking that the value displayed is the value we put in, so if we can't just swap appeals in and out we're presumably finding bugs?

danWatkinson:
binned this off for a new branch to 1) rebase from master, and 2) bring my PR up towards the top of the pile because it was 2 weeks old and looked like it had been forgotten
->https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1053

PR #1030, opened 2021-04-27
Feature/AS-1606

pbradley-f4:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-1606

## Description of change
Addition of new "your planning appeal" page allowing the appellant to view their submitted appeal

## Checklist
- [x] New features have tests

## Important


packages/forms-web-app/src/controllers/your-planning-appeal.js
```diff
@@ -0,0 +1,21 @@
+const logger = require('../lib/logger');
+const { VIEW } = require('../lib/views');
+const appealModel = require('../models/appealModel');
+
+exports.getYourPlanningAppeal = async (req, res, next) => {
+  const { appealId } = req.params;
+  let appeal;
+
+  try {
+    appeal = await appealModel.getAppeal(appealId);
```
a6software, line 10: There is existing code to do this via middleware. 

https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/forms-web-app/src/middleware/fetch-appeal-by-url-param.js

An example of how it's used:

https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/forms-web-app/src/routes/appellant-submission/submission-information.js#L11

If you prefer your approach, please can you refactor the above & any other usages to use your code, so we keep one standard?

packages/forms-web-app/src/controllers/your-planning-appeal.js
```diff
@@ -0,0 +1,21 @@
...
+    if (error.response.status === 404) return next();
+
+    logger.error(error);
+    return next(error);
+  }
+
+  res.render(VIEW.YOUR_PLANNING_APPEAL, {
+    appeal,
+  });
+};
```
a6software, line 21: Please could the path to this file be:

`packages/forms-web-app/src/controllers/your-planning-appeal/index.js`

This will help with https://pins-ds.atlassian.net/browse/AS-1607 where there needs to be sub pages under this route. 

Thanks

packages/forms-web-app/src/apis/appealsApi.js
```diff
@@ -0,0 +1,58 @@
...
+      url,
+      method,
+      headers: {
+        'Content-Type': 'application/json',
+        'X-Correlation-ID': correlationId,
+        ...params.headers,
+      },
+    });
+
+    logger.debug('Api successfully responded with', apiResponse);
```
a6software, line 27: If given, the logger statements should log the object as the first param.

https://github.com/pinojs/pino/blob/master/docs/api.md#errors

`logger.debug(apiResponse, 'Api successfully responded with');`

or

`logger.debug({ apiResponse, otherInterestingObjects }, 'Api successfully responded with');`

packages/forms-web-app/src/controllers/your-planning-appeal.js
pbradley-f4: Cheers Chris.. .I saw this but the functioanlity/behaviour wanst quite what was needed I think. Will have a chat and go from there.

packages/forms-web-app/src/apis/appealsApi.js
pbradley-f4: nice catch ... cheers

packages/forms-web-app/src/controllers/your-planning-appeal/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+    logger.error(error);
+    return next(error);
+  }
+
+  return res.render(VIEW.YOUR_PLANNING_APPEAL, {
+    appeal,
+    lpd,
+  });
+};
```
a6software, line 25: FWIW I am happy to move to axios - no strong opinions on the library used.

As I understand it, your concern here is that by using the two middleware:

* https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/forms-web-app/src/middleware/fetch-appeal-by-url-param.js
* https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/forms-web-app/src/middleware/fetch-appeal-lpd-by-appeal-lpa-code.js

is that they are coupled tightly to the session?

It's a good call - and a bad decision by me to do that. 

However, I would suggest we address the existing middlewares, rather than create a new way to solve the problem.

I have updated my experimental PR to use `res.locals.xxx` instead:

* https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1043/files#diff-019963eade9afb2e756006ba5343a8ca4ee1a6ad34581edf34c71201fa86cc55
* https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1043/files#diff-9cb622bf8bdd390e6e67facca5694b49ba25d47bf3d2405bfb8c142f5b224b6a

This *should* address the root cause issue, if the problem is solely the use of `req.session`. 

We do not need (afaik) to use `req.session` for the submission-information display, we can rely on `res.locals` as above. That change is also in the experimental PR. 

If we did need to put the retrieved appeal into the session again for any reason, we could add a middleware to handle that specific use case, and call that after, e.g. :

```
module.exports = (req, res, next) => {
  const { appeal } = res.locals;

  if (!appeal) {
    delete req.session.appeal;
    return next();
  }

  try {
    req.session.appeal = appeal;
  } catch (err) {
    delete req.session.appeal;
  }

  return next();
};
``` 

This isn't needed at the moment, afaik, so I have not included it.

To be completely clear, I'm not knocking what you have done technically - no issues there - my desire is simply not to have two ways to solve the same problem. If we can adapt what we already have, I believe we should, in the first instance.

pbradley-f4: I totally understand where you are coming from with not wanting 2 approaches, however sometimes 2 approaches work for different use cases. While the session could be the right way to acheive what we need for the stepped submission phase, for me this kind of view should be always using up to date data and should be clean and simple. Im not sure if I like the idea of middleware populating a session or res.locals with this data, if not simply because having that interstitial "data retreive and store" phase adds no value for this as far as I can see?

a6software: From my pov the value add comes from when we want to retrieve the appeal again on many separate pages. By placing that process in middleware, we can "one line" it in where ever it's needed, rather than having the additional try / catch per controller along with all the associated unit testing code. 

This use case is already in place for the submission information page: 

https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/forms-web-app/src/controllers/appellant-submission/submission-information.js#L13

As such this wouldn't be two different approaches for different use cases. If we choose this approach, imho we absolutely need to refactor submission-information to be the same.

packages/forms-web-app/src/routes/index.js
```diff
@@ -7,11 +7,13 @@ const eligibilityRouter = require('./eligibility');
...
 const cookieRouter = require('./cookies');
 const guidancePagesRouter = require('./guidance-pages');
+const yourPlanningAppealRouter = require('./your-planning-appeal');
 
 router.use('/', homeRouter);
 router.use(guidancePagesRouter);
 router.use('/cookies', cookieRouter);
 router.use('/appellant-submission', appellantSubmissionRouter);
 router.use('/eligibility', eligibilityRouter);
+router.use('/your-planning-appeal', yourPlanningAppealRouter);
```
AnthonyJWalsh, line 17: Pedantic. But  should this be "my-planning-appeal" ???

a6software: I have asked internally - I got told what URL to use here as part of the story.

a6software: Anthony Hope is stating it is correct as `/your-planning-appeal`. 

Also - sorry, I thought this was for AS-1607, sorry to jump in on your ticket Paul!!

packages/forms-web-app/tests/unit/routes/index.test.js
```diff
@@ -21,6 +21,7 @@ describe('routes/index', () => {
     expect(use).toHaveBeenCalledWith('/cookies', cookieRouter);
     expect(use).toHaveBeenCalledWith('/appellant-submission', appellantSubmissionRouter);
     expect(use).toHaveBeenCalledWith('/eligibility', eligibilityRouter);
-    expect(use.mock.calls.length).toBe(5);
+    expect(use).toHaveBeenCalledWith('/your-planning-appeal', eligibilityRouter);
```
a6software, line 24: highlights the tests here are less than stellar here:

`expect(use).toHaveBeenCalledWith('/your-planning-appeal', yourPlanniningAppealRouter);`

a6software (dismissed):
Seems odd to me that nothing failed in the `submission-information` journey - will take a look at that :/

a6software (approved):

a6software:
approved again after squash / force + merging now as I need this for 1607

PR #1033, opened 2021-04-28
feat: unable to submit appeals with expired deadline date

dimodimovplanning:
stop users from submitting appeals with expired decision date

## Ticket Number
https://pins-ds.atlassian.net/browse/AS-2076

## Description of change
users will be unable to submit appeals with expired decision date

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/controllers/appellant-submission/confirmation.js
```diff
@@ -1,12 +1,18 @@
...
+  const decisionDatePassed = isDecisionDatePassed(req.session.appeal);
 
-  res.render(VIEW.APPELLANT_SUBMISSION.CONFIRMATION, {
-    appellantEmail,
-    appealId,
-  });
+  if (decisionDatePassed) {
+    res.redirect(`/${VIEW.ELIGIBILITY.DECISION_DATE_PASSED}`);
+  } else {
+    req.session.appeal = null;
```
swdpcomputing, line 12: Is this simply to clean up? 

Also, I can't see a test for the path where it becomes nullified.

packages/forms-web-app/src/controllers/eligibility/decision-date.js
```diff
@@ -8,6 +8,14 @@ const getDeadlineDate = (decisionDate) => {
   return addWeeks(endOfDay(decisionDate), 12);
 };
 
+const isDecisionDatePassed = (appeal) => {
```
swdpcomputing, line 11: I can't see a test for this function added....

a6software, line 11: please extract this, `require` back in, and add unit tests to cover

thanks

swdpcomputing (commented):
Apologies if these comments / questions are irrelevant / obvious as I'm not up to date with what you're doing. I had some free time so thought I'd have a look and save someone else the time if possible.

a6software:
Hi Dimo, 

I'm looking at https://pins-ds.atlassian.net/browse/AS-2002 and see AS-2077 is also in 'waiting for review' - please can you share that change also? From my point of view, that is the more important piece of this ticket as without it, we still leave open the hole of accepting invalid appeals directly to the API. This ticket addresses the forms web app / 'front end' stuff, but if e.g. the appeals questionaire modifies the appeal directly, we don't validate the decision date as per your new logic in that case.

Thanks

PR #1043, opened 2021-04-29
feat: adds /your-planning-appeal/:appealid

a6software:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-1606

## Checklist

- [ ] New features have tests

## Important


a6software:
Heads up - the CI will fail for this, for various reasons - it's an experiment, so not that concerned. 

@pau1brad1ey I was intrigued as to what issue you are facing with the existing middleware available for your PR - https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1030

This is an experiment with using the middleware - please can you help me understand the issue you're facing? This approach seems to work, as best I can see. 

**However**

@dimodimovplanning there is a bug in the change made in this PR - https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1011 

The line at fault here is:

https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1043/files#diff-2d8569bdc6bca4acc1c92b10f7cdaeb3bb0c026f1c85e924e7e5049bb64e6fa1R6

For clarity, run the feature file scenario in this PR through to completion. It will fail - but that's seemingly due to a pending test step. 

1. Open the mongo DB.
2. Whilst running the scenario, one new appeal should be created.
3. This appeal should be filled in during the test
4. The appeal is submitted
5. The confirmation page is displayed
6. With the line commented out, this test gets further - but breaks other tests. This is the expected outcome, however it means we no longer can do multiple appeals.

Repeat the tests with the `req.session.appeal = null` line.

Now the test fails differently.

Check the DB after the test has run, we now have two appeals. One for the appeal that got filled in, and one that gets newly created by the middleware on the visit to the confirmation page. This is a bug for sure. 

Happy to discuss tomorrow if unsure, however this is a show stopper for AS-1607 at this point.

a6software:
closing, was experimental

PR #1164, opened 2021-05-25
chore: as-2527 address clobber tags errors

roryconlan:
## Ticket Number
https://pins-ds.atlassian.net/browse/AS-2527

## Description of change
Workflows regularly fail due to the git fetch error "would clobber existing tag". This seems to happen all the time. I propose to address this by doing a full git fetch at the start of the pipeline, instead of a shallow fetch. This should resolve the issue. It's the flux sync tags that are causing the problem.

## Checklist

- [ ] New features have tests

## Important


swdpcomputing (approved):
LGTM

PR #1136, opened 2021-05-20
Feature/as 1660 send confirmation email

LouiseMcMahon:
## Ticket Number
https://pins-ds.atlassian.net/browse/AS-1660

## Description of change
This ticket is to send a confirmation email to an LPA once a questionnaire has been submitted.  The email will contain a link to the Questionnaire PDF for that particular appeal.

## Checklist

- [ ] Requires infrastructure changes
- [x] If adding or remove environment variables (e.g. in `docker-compose.yaml`) then I have updated the appropriate Helm chart
- [ ] I have updated the documentation accordingly
- [x] My commit history in this PR is linear
- [x] New features have tests
- [ ] Breaking change (team conversation required)

## Important

Please do not merge from `master` (please only [rebase](https://github.com/foundry4/appeal-planning-decision/wiki/An-intro-to-Git-Rebase)). This keeps the history linear and easier to debug.

PR #1174, opened 2021-05-27
Feature/as 1660 send confirmation email cm

a6software:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-1660

## Description of change
This ticket is to send a confirmation email to an LPA once a questionnaire has been submitted. The email will contain a link to the Questionnaire PDF for that particular appeal.

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/charts/app/templates/appealReplyServiceApi.yaml
+++ b/charts/app/templates/appealReplyServiceApi.yaml
@@ -1,4 +1,4 @@
-{{ $queueConfig := index .Values "horizon-householder-appeal-publish" }} #Same as appealsService
+{{ $queueConfig := index .Values "horizon-householder-appeal-publish" }}
 
 apiVersion: apps/v1
 kind: Deployment
@@ -88,6 +88,23 @@ spec:
                   name: {{ $queueConfig.amqp.connection.existingUsernameSecret | quote }}
             - name: HORIZON_HAS_PUBLISHER_QUEUE
               value: {{ $queueConfig.amqp.receiver.source | quote }}
+            - name: DOCUMENTS_SERVICE_API_URL
+              value: "http://{{ include "app.fullname" . }}:{{ .Values.documentServiceApi.service.port }}"
+            - name: APPEALS_SERVICE_API_URL
+              value: "http://{{ include "app.fullname" . }}:{{ .Values.appealsServiceApi.service.port }}"
+            - name: SRV_NOTIFY_BASE_URL
+              value: {{ .Values.appealReplyServiceApi.config.notify.baseUrl | quote }}
+            - name: SRV_NOTIFY_SERVICE_ID
+              value: {{ .Values.appealReplyServiceApi.config.notify.serviceId | quote }}
+            - name: SRV_NOTIFY_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  key: value
+                  name: {{ .Values.appealReplyServiceApi.config.notify.secretName }}
+            - name: SRV_NOTIFY_TEMPLATE_ID_APPEAL_REPLY_SUBMISSION_CONFIRMATION
+              value: {{ .Values.appealReplyServiceApi.config.notify.templates.appealReplyNotificationEmailToLpa | quote }}
+            - name: SRV_NOTIFY_EMAIL_REPLYTO_ID_APPEAL_REPLY_SUBMISSION_CONFIRMATION
+              value: {{ .Values.appealReplyServiceApi.config.notify.emailReplyToId.appealReplyNotificationEmailToLpa | quote }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
       {{- with .Values.nodeSelector }}
--- a/packages/appeal-reply-service-api/.eslintrc
+++ b/packages/appeal-reply-service-api/.eslintrc
@@ -1,6 +1,6 @@
 {
   "parserOptions": {
-    "ecmaVersion": 8,
+    "ecmaVersion": 2020,
     "sourceType": "module"
   },
   "extends": ["airbnb-base", "prettier"],
--- a/packages/appeal-reply-service-api/src/__tests__/notify.spec.js
+++ b/packages/appeal-reply-service-api/src/__tests__/notify.spec.js
@@ -1,0 +1,235 @@
+const mockError = jest.fn();
+const mockInfo = jest.fn();
+
+jest.doMock('../lib/logger', () => ({
+  error: mockError,
+  info: mockInfo,
+}));
+
+jest.mock('axios');
+jest.mock('@pins/common/src/lib/notify/notify-builder', () => {
+  const mock = {
+    setNotifyClient: jest.fn(() => mock),
+    setTemplateId: jest.fn(() => mock),
+    setTemplateVariable: jest.fn(() => mock),
+    setTemplateVariablesFromObject: jest.fn(() => mock),
+    addFileToTemplateVariables: jest.fn(() => mock),
+    setDestinationEmailAddress: jest.fn(() => mock),
+    setEmailReplyToId: jest.fn(() => mock),
+    setReference: jest.fn(() => mock),
+    sendEmail: jest.fn(() => Promise.resolve('done')),
+  };
+  return mock;
+});
+
+const axios = require('axios');
+const NotifyBuilder = require('@pins/common/src/lib/notify/notify-builder');
+const { sendAppealReplySubmissionConfirmationEmailToLpa } = require('../lib/notify.js');
+const config = require('../lib/config');
+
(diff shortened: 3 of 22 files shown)


packages/common/src/lib/notify/notify-builder.js
```diff
@@ -1,184 +1,212 @@
...
+    this.destinationEmail = undefined;
+    this.emailReplyToId = undefined;
+    this.reference = undefined;
+
+    return this;
+  },
 
   /**
-   * Intentionally do not expose the notify client directly.
+   * Gets the set client and set it if its not set sets it
```
a6software, line 23: ?

packages/appeal-reply-service-api/src/__tests__/notify.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    setTemplateId: jest.fn(() => mock),
+    setTemplateVariable: jest.fn(() => mock),
+    setTemplateVariablesFromObject: jest.fn(() => mock),
+    addFileToTemplateVariables: jest.fn(() => mock),
+    setDestinationEmailAddress: jest.fn(() => mock),
+    setEmailReplyToId: jest.fn(() => mock),
+    setReference: jest.fn(() => mock),
+    sendEmail: jest.fn(() => Promise.resolve('done')),
+  };
+  return mock;
```
a6software, line 19: why can't this be short hand?

because it is self referencing - jest fn's return self

packages/common/src/lib/notify/notify-builder.js
```diff
@@ -1,184 +1,212 @@
...
+    logger.info('Resetting the notify client');
+    this.notifyClient = undefined;
+    this.templateId = undefined;
+    this.templatePersonalisation = undefined;
+    this.destinationEmail = undefined;
+    this.emailReplyToId = undefined;
+    this.reference = undefined;
+
+    return this;
+  },
```
a6software, line 20: I'm honestly not sure how I feel about this. 

My personal opinion was that we should not use a single instance of the notify client to send multiple emails as it leads to the risk that this function has been added to avoid.

This now relies on the developer remembering to `reset` between uses, and of greater concern, requires future development to ensure this method is kept updated. 

imho, it was preferable to explicitly build a new instance of the notify wrapper per email dispatch, to avoid the risk of retaining one or more pieces of something else's data.

packages/appeal-reply-service-api/src/__tests__/notify.spec.js
a6software: cc @davehlewis re: https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1136#discussion_r637919650

packages/appeal-reply-service-api/src/__tests__/notify.spec.js
```diff
@@ -0,0 +1,83 @@
...
+        data: {
+          name,
+          email,
+        },
+      })
+      .mockResolvedValueOnce({
+        data: fileData,
+      });
+
+    await notify.sendAppealReplySubmissionConfirmationEmailToLpa({
```
swdpcomputing, line 54: Why are we importing notify rather than just pulling in the function?

rcmsoftwaredevelopment (dismissed):

packages/appeal-reply-service-api/src/__tests__/notify.spec.js
swdpcomputing, line 54: Would be clearer to define the reply object parameter outside the function call.

packages/appeal-reply-service-api/src/lib/notify.js
```diff
@@ -0,0 +1,56 @@
...
+    const lpaRes = await axios.get(lpaUrl);
+    lpa = lpaRes.data;
+
+    const pdfUrl = `${config.docs.api.url}/api/v1/${reply.id}/${reply.submission.pdfStatement.uploadedFile.id}/file`;
+    logger.info({ docsPath: pdfUrl }, 'docs service route');
+
+    const pdfRes = await axios.get(pdfUrl, { responseType: 'blob' });
+    pdf = pdfRes.data;
+  } catch (err) {
+    logger.error({ err }, 'Unable to get data required to send email.');
```
swdpcomputing, line 26: At a later date it might be useful to get some seperation of why the error occurred, i.e. which service / piece of data.

packages/appeal-reply-service-api/src/__tests__/notify.spec.js
```diff
@@ -0,0 +1,83 @@
...
+      'planning application number': appealId,
+    });
+    expect(NotifyBuilder.setReference).toHaveBeenCalledWith(`${appealId}.SubmissionConfirmation`);
+    expect(NotifyBuilder.addFileToTemplateVariables).toHaveBeenCalledWith(
+      'link to appeal submission pdf',
+      fileData
+    );
+    expect(NotifyBuilder.sendEmail).toHaveBeenCalled();
+  });
+});
```
swdpcomputing, line 83: I can't see any unhappy path testing. Is it not appropriate here for some reason?

packages/appeal-reply-service-api/src/controllers/reply.js
```diff
@@ -73,35 +74,40 @@ module.exports = {
       .collection(dbId)
       .findOne({ _id: idParam })
       .then(async (originalDoc) => {
```
swdpcomputing, line 76: Comment out rather than remove the custom logging for Horizon please.

packages/common/src/lib/notify/notify-builder.js
swdpcomputing: I agree with this. There should be a new instance. Does that have an impact upon state?

a6software: Yes, the difference is if a new instance is created per send, we get a clean state by virtue of construction. This was the previous implementation.

In this PR's implementation, we get a clean state on instantiation, but the state is only wiped if the consumer explicitly calls `reset()`. 

I can live with this, but I'm not sold on this being an improvement on what we had. 

That said, the other approach was not perfect by any means.

packages/appeal-reply-service-api/src/__tests__/notify.spec.js
a6software: Updated to pull in the function :+1:

a6software: Extracted as suggested :+1:

packages/appeal-reply-service-api/src/lib/notify.js
a6software: This is now separated as suggested :+1:

packages/appeal-reply-service-api/src/controllers/reply.js
a6software: Re-added, and commented as requested. Sorry about that!

packages/appeal-reply-service-api/.eslintrc
```diff
@@ -1,6 +1,6 @@
 {
   "parserOptions": {
-    "ecmaVersion": 8,
+    "ecmaVersion": 2020,
```
a6software, line 3: for optional chaining - discussed and approved, and in use elsewhere

only impacts linting

packages/appeal-reply-service-api/src/controllers/reply.js
swdpcomputing: It looks like some of the 'STEVE' logs are still out?

a6software: https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1174/files#diff-9b363b6ec9a80d94cff8dffbe57677b8cca5f9f7626f83a4048f593dbc214f76R78

and

https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/1174/files#diff-9b363b6ec9a80d94cff8dffbe57677b8cca5f9f7626f83a4048f593dbc214f76R90

Now back in play, sorry about that!

swdpcomputing (approved):

charts/app/templates/appealReplyServiceApi.yaml
```diff
@@ -1,4 +1,4 @@
-{{ $queueConfig := index .Values "horizon-householder-appeal-publish" }} #Same as appealsService
```
a6software, line 1: `[WARNING] templates/appealReplyServiceApi.yaml: document starts with an illegal indent: " #Same as appealsService", which may cause parsing problem`

swdpcomputing (approved):

AnthonyJWalsh (approved):

a6software:
ty

PR #1260, opened 2021-06-09
Fix/horizon householder appeal publish version bump

swdpcomputing:
## Ticket Number

## Description of change

add getHorizonId

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/horizon-householder-appeal-publish/src/config.js
+++ b/packages/horizon-householder-appeal-publish/src/config.js
@@ -11,4 +11,7 @@ module.exports = {
   openfaas: {
     gatewayUrl: process.env.GATEWAY_URL,
   },
+  logger: {
+    level: process.env.LOGGER_LEVEL || 'info',
+  },
 };
--- a/packages/horizon-householder-appeal-publish/src/lib/logger.js
+++ b/packages/horizon-householder-appeal-publish/src/lib/logger.js
@@ -1,0 +1,14 @@
+/**
+ * Logger
+ *
+ * Creates a log instance with common configuration that can be
+ * used throughout the application
+ */
+
+const pino = require('pino');
+const config = require('../config');
+
+module.exports = pino({
+  level: config.logger.level,
+  redact: config.logger.redact,
+});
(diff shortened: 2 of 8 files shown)


packages/horizon-householder-appeal-publish/src/config.js
```diff
@@ -11,4 +11,7 @@ module.exports = {
   openfaas: {
     gatewayUrl: process.env.GATEWAY_URL,
   },
+  logger: {
+    level: process.env.LOGGER_LEVEL || 'info',
```
a6software, line 15: does a corresponding entry need adding to the helm config ?

packages/horizon-householder-appeal-publish/src/lib/logger.js
```diff
@@ -0,0 +1,14 @@
...
+ * Logger
+ *
+ * Creates a log instance with common configuration that can be
+ * used throughout the application
+ */
+
+const pino = require('pino');
+const config = require('../config');
+
+module.exports = pino({
```
a6software, line 11: this is more of an FYI than a request to change - doing this makes life very difficult for writing unit tests in my experience. 

Not asking you to change - just flagging it as I know you love a good bit of jest ;)

packages/horizon-householder-appeal-publish/src/config.js
swdpcomputing: I don't know, does it?

packages/horizon-householder-appeal-publish/src/lib/logger.js
swdpcomputing: This is just the way it's done in other packages, I have little experience. Very happy to have suggestions for a different method.

packages/horizon-householder-appeal-publish/src/config.js
a6software: As best I understand it, `process.env.LOGGER_LEVEL` will be `undefined` when deployed beyond local dev. This will fall back to `info`, so all good in terms of valid config.

However, if you want or need to have a different log level beyond local dev then you would need to set up that config inside helm. 

I guess it's YAGNI, right?

packages/horizon-householder-appeal-publish/src/lib/logger.js
a6software: To clarify, this module exports an immediate function call to a third party package.

Writing tests for code that does this is very difficult. My comment was to suggest that for your own code, try to avoid doing this as it makes jest harder than it already is.

swdpcomputing: OK

packages/horizon-householder-appeal-publish/src/config.js
swdpcomputing: In that case, leave it. Ta.

PR #1292, opened 2021-06-14
fix(lpa-questionnaire-web-app): fix deletion event for invalid files

davehlewis:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-2726

## Description of change
Fixes incident when user could not delete file when it is invalid

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/lpa-questionnaire-web-app/tests/unit/controllers/files.test.js
+++ b/packages/lpa-questionnaire-web-app/tests/unit/controllers/files.test.js
@@ -151,6 +151,24 @@ describe('controllers/files', () => {
           });
         },
       },
+      {
+        title: 'should return status 200 if deleted file is invalid (id is "undefined" string)',
+        given: () => ({
+          ...mockReq(),
+          body: {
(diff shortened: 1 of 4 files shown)


packages/lpa-questionnaire-web-app/tests/unit/controllers/files.test.js
```diff
@@ -151,6 +151,24 @@ describe('controllers/files', () => {
           });
         },
       },
+      {
+        title: 'should return status 200 if deleted file is invalid (id is "undefined" string)',
```
swdpcomputing, line 155: Is this supposed to return success if the id does not exist? I presume this is considered success if the file does not exist?

davehlewis: If you look at the comment it explains that the specific string of undefined only happens in the use case of an invalid document

swdpcomputing (approved):

PR #1299, opened 2021-06-15
feat: add your-planning-appeal link to post appeal submission email

pbradley-f4:
## Ticket Number
https://pins-ds.atlassian.net/browse/AS-1703

## Description of change
Adds link to `your-planning-appeal` to submission confirmation email instead of pdf

## Checklist
- [x] If adding or remove environment variables (e.g. in `docker-compose.yaml`) then I have updated the appropriate Helm chart
- [x] New features have tests

## Important

### DIFF
--- a/charts/app/templates/appealsServiceApi.yaml
+++ b/charts/app/templates/appealsServiceApi.yaml
@@ -51,6 +51,8 @@ spec:
               path: /health
               port: {{.Values.appealsServiceApi.service.name }}
           env:
+            - name: APP_APPEALS_BASE_URL
+              value: "https://{{ .Values.global.domainNames.appealsWebApp }}"
             - name: DOCUMENTS_SERVICE_API_URL
               value: "http://{{ include "app.fullname" . }}:{{ .Values.documentServiceApi.service.port }}"
             - name: HORIZON_HAS_PUBLISHER_HOST
@@ -107,6 +109,8 @@ spec:
               value: {{ .Values.appealsServiceApi.config.notify.templateId | quote }}
             - name: SRV_NOTIFY_APPEAL_SUBMISSION_RECEIVED_NOTIFICATION_EMAIL_TO_LPA_TEMPLATE_ID
               value: {{ .Values.appealsServiceApi.config.notify.templates.appealNotificationEmailToLpa | quote }}
+            - name: SRV_NOTIFY_APPEAL_SUBMISSION_CONFIRMATION_EMAIL_TO_APPELLANT_TEMPLATE_ID
+              value: {{ .Values.appealsServiceApi.config.notify.templates.appealSubmissionConfirmationEmailToAppellant | quote }}
           volumeMounts:
             - name: lpa-data-volume
               mountPath: /data/lpa
--- a/charts/app/values.yaml
+++ b/charts/app/values.yaml
@@ -2,6 +2,10 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+global:
+  domainNames:
+    appealsWebApp: ""
+
 appealReplyServiceApi:
   replicaCount: 1
   service:
@@ -42,6 +46,7 @@ appealsServiceApi:
       templateId: 15ed37a9-506c-4845-88ea-95502282a863
       templates:
         appealNotificationEmailToLpa: 79488d5d-7efd-4273-a11f-e73f11d19676
+        appealSubmissionConfirmationEmailToAppellant: 72f71441-12bf-4455-afbc-c58f9c72bfbd
 
 documentServiceApi:
   replicaCount: 1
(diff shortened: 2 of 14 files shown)


releases/dev/app.yml
```diff
@@ -15,6 +15,10 @@ spec:
   rollback:
     enable: true
   values:
+    globals:
+      domainNames:
+        appealsWebApp: appeals-dev.planninginspectorate.gov.uk
```
a6software, line 20:
```
➜  appeal-planning-decision git:(feature/AS-1703) ✗ helm lint charts/app
==> Linting charts/app
[ERROR] templates/: template: app/templates/appealsServiceApi.yaml:55:40: executing "app/templates/appealsServiceApi.yaml" at <.Values.globals.domainNames.appealsWebApp>: nil pointer evaluating interface {}.domainNames

Error: 1 chart(s) linted, 1 chart(s) failed
```

So, at a guess, this also needs the config adding:

```
# charts/app/values.yaml

globals:
  domainNames:
    appealsWebApp: http://localhost:9003/
```

That does satisfy helm lint, **but** that doesn't mean it's the correct config. I'm not sure, that's just my best guess.

packages/appeals-service-api/src/lib/notify.js
```diff
@@ -116,6 +116,33 @@ async function sendEmail(appeal) {
   }
 }
 
+async function sendAppealSubmissionConfirmationEmailToAppellant(appeal) {
+  if (!isValidAppealForSubmissionReceivedNotificationEmail(appeal)) {
```
a6software, line 120: should this be

`isValidAppealForSubmissionConfirmationEmailToAppellantEmail`

or have I misunderstood ? (genuinely unsure)

releases/dev/app.yml
roryconlan: You're almost there guys. I've just refreshed my memory on this stuff.
https://helm.sh/docs/chart_template_guide/values_files/

In order for this to work, you would need another dev specific values file. Something like this:
helm lint charts/app --values charts/app/values.yaml --values releases/dev/app-dev-values.yaml

However, this would need a change to the flux install, to instruct it to read the new environment specific values files. I'll write up a user story. For now, if you want to get this out, you will have to update the default values.yaml, as Chris mentioned. Unfortunately, as it's currently configured, this will update all three environments.

charts/app/templates/appealsServiceApi.yaml
```diff
@@ -106,6 +108,8 @@ spec:
             - name: SRV_NOTIFY_TEMPLATE_ID
               value: {{ .Values.appealsServiceApi.config.notify.templateId | quote }}
             - name: SRV_NOTIFY_APPEAL_SUBMISSION_RECEIVED_NOTIFICATION_EMAIL_TO_LPA_TEMPLATE_ID
+              value: {{ .Values.appealsServiceApi.config.notify.templates.appealSubmissionConfirmationEmailToAppellant | quote }}
+            - name: SRV_NOTIFY_APPEAL_SUBMISSION_CONFIRMATION_EMAIL_TO_APPELLANT_TEMPLATE_ID
               value: {{ .Values.appealsServiceApi.config.notify.templates.appealNotificationEmailToLpa | quote }}
           volumeMounts:
```
roryconlan, line 114: These are mixed up.
Should be:
- name: NAME_VAR_1
   value: VALUE_VAR_1

charts/app/values.yaml
```diff
@@ -15,9 +15,9 @@ appealReplyServiceApi:
     port: 3000
     logLevel: trace
     notify:
-      baseUrl: "" # Let Notify package resolve this itself
-      secretName: ""
-      serviceId: ""
+      baseUrl: '' # Let Notify package resolve this itself
+      secretName: ''
+      serviceId: ''
       emailReplyToId:
```
roryconlan, line 21: My YAML formatter doesn't like your changes to this file.  Are you sure about these?

releases/dev/app.yml
pbradley-f4: 👍

charts/app/templates/appealsServiceApi.yaml
pbradley-f4: 👍

charts/app/values.yaml
pbradley-f4: Prettier kicked in..will revert all of that to be sure

packages/appeals-service-api/src/lib/notify.js
pbradley-f4: To be honest, I just resued the same method to check the state of the appeal before sending as in existing `sendAppealSubmissionReceivedNotificationEmailToLpa`. Agree its a poorly named method and maybe the check is a little blunt, but just blindly followed the exising pattern if Im honest.

a6software (approved):
LGTM

roryconlan (approved):

PR #1333, opened 2021-06-29
feat(horizon-householder-appeal-publish): add pdf to horizon submission

swdpcomputing:
## Ticket Number

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/horizon-householder-appeal-publish/handler-reply.js
+++ b/packages/horizon-householder-appeal-publish/handler-reply.js
@@ -44,19 +44,34 @@ const convertDocumentArray = (arrayOfFiles, type) => {
  * @param {object} body contains the reply object
  * @returns {documents} documents object for Horizon
  */
-const populateDocuments = (body) => {
+const populateDocuments = (event) => {
   const documents = [];
 
   Object.keys(documentsSections).forEach((key) => {
-    if (body[documentsSections[key]] !== undefined) {
+    if (event.body[documentsSections[key]] !== undefined) {
       documents.push(
         convertDocumentArray(
-          body[documentsSections[key]][key].uploadedFiles,
+          event.body[documentsSections[key]][key].uploadedFiles,
           sectionTypes[`${key}Type`]
         )
       );
     }
   });
+
+  const id = event.body.submission.pdfStatement?.uploadedFile.id;
+
+  if (id === undefined) {
+    const message = 'PDF not present in submission';
+    event.log.error(
+      {
+        message,
+        data: event.body.submission,
+      },
+      message
+    );
+  } else {
+    documents.push({ id, type: sectionTypes.pdfType });
+  }
 
   return documents.flat();
 };
@@ -102,7 +117,7 @@ const handlerReply = async (event, context) => {
   try {
     const { body } = event;
     const replyId = body.id;
-    await publishDocuments(event.log, populateDocuments(body), replyId, horizonId);
+    await publishDocuments(event.log, populateDocuments(event), replyId, horizonId);
     return {
       id: horizonId,
     };
--- a/packages/horizon-householder-appeal-publish/src/sectionTypes.js
+++ b/packages/horizon-householder-appeal-publish/src/sectionTypes.js
@@ -10,4 +10,5 @@ module.exports = {
   statutoryDevelopmentType: 'Development Plans',
   otherPoliciesType: 'LPA Questionnaire Documents',
   supplementaryPlanningDocumentsType: 'Supplementary Guidance',
+  pdfType: 'LPA Questionnaire Documents',
 };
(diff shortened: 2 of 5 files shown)


packages/horizon-householder-appeal-publish/handler-reply.js
```diff
@@ -58,6 +58,8 @@ const populateDocuments = (body) => {
     }
   });
 
+  documents.push({ id: body.submission.pdfStatement.uploadedFile.id, type: sectionTypes.pdfType });
```
AnthonyJWalsh, line 61: What happens here if there is no pdfStatement in the submission on the body? The FWA stuff will break as it doesnt always have a body.submission.pdfStatement object.....

AnthonyJWalsh (changes requested):
One of two things is happening here. 

Either 

a) the code changes below will cause a breaking change to the existing documents upload functionality used by the forms web app as forms web app documents don't have a body.submission.pdfStatement.uploadedFile.id

or b) the forms web app uses a completely different mechanism to upload files to Horizon, which seems like unnecessary duplication. 

Happy to chat about this.

packages/horizon-householder-appeal-publish/handler-reply.js
swdpcomputing: I wasn't aware that there would ever not be a pdfStatement. Will amend.

swdpcomputing:
> One of two things is happening here.
> 
> Either
> 
> a) the code changes below will cause a breaking change to the existing documents upload functionality used by the forms web app as forms web app documents don't have a body.submission.pdfStatement.uploadedFile.id
> 
> or b) the forms web app uses a completely different mechanism to upload files to Horizon, which seems like unnecessary duplication.
> 
> Happy to chat about this.

This part of the Horizon code has a handler (for the appeal), and a handler-reply (for the questionnaire). There is indeed some similar functionality, however they do a different enough job that I kept them separate. Also, I didn't want to mess around too much with Team 1's work. The appeal has parts that the reply does not, such as publishing contacts. There is indeed shared functionality that is used in both processes, such as:

packages/horizon-householder-appeal-publish/src/catchErrorHandling.js
packages/horizon-householder-appeal-publish/src/publishDocuments.js

...which I separated from the original functionality for the purpose of sharing code.

packages/horizon-householder-appeal-publish/handler-reply.js
AnthonyJWalsh: I mean, i don't know for certain that it is or is not possible, just seems like a safety check that doesnt cost us anything

AnthonyJWalsh:
> > One of two things is happening here.
> > Either
> > a) the code changes below will cause a breaking change to the existing documents upload functionality used by the forms web app as forms web app documents don't have a body.submission.pdfStatement.uploadedFile.id
> > or b) the forms web app uses a completely different mechanism to upload files to Horizon, which seems like unnecessary duplication.
> > Happy to chat about this.
> 
> This part of the Horizon code has a handler (for the appeal), and a handler-reply (for the questionnaire). There is indeed some similar functionality, however they do a different enough job that I kept them separate. Also, I didn't want to mess around too much with Team 1's work. The appeal has parts that the reply does not, such as publishing contacts. There is indeed shared functionality that is used in both processes, such as:
> 
> packages/horizon-householder-appeal-publish/src/catchErrorHandling.js
> packages/horizon-householder-appeal-publish/src/publishDocuments.js
> 
> ...which I separated from the original functionality for the purpose of sharing code.

Seems fair. As this is an interim solution (Horizon integration more generally), im less fussed to be honest. Approved.

packages/horizon-householder-appeal-publish/handler-reply.js
swdpcomputing: If that's the case the process should have failed before now if it was unable to create a pdf and place it in this location. The whole process should stop, and this would be an innappropriate place for it to fail.

AnthonyJWalsh: "should have" - but can this service, in isolation, guarantee that it will never get a body.submission without a pdfStatement on it? And if for some reason it does get sent something without a pdfStatement, will we get a useful error message out that quickly shows us exactly why it is failing, or will it crash out, never returning the documents object that its supposed to, leaving us digging around trying to work out why a set of LPA documents isn't in Horizon.

swdpcomputing: I think that's fair. We probably need to have a conversation about what happens if this occurs. Will add the error messge.

LouiseMcMahon (dismissed):

AnthonyJWalsh (dismissed):

LouiseMcMahon (approved):

PR #1335, opened 2021-06-29
feat(forms-web-app): proxy to documents service api to allow in session downloads

a6software:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-2802

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/e2e-tests/cypress/support/appeal-statement-submission/confirmAppealStatementFileIsUploaded.js
+++ b/e2e-tests/cypress/support/appeal-statement-submission/confirmAppealStatementFileIsUploaded.js
@@ -1,6 +1,6 @@
 module.exports = (filename) => {
   cy.goToAppealStatementSubmission();
-  cy.get('#appeal-upload-file-name')
+  cy.get('[data-cy="appeal-upload-file-link"]')
     .invoke('text')
     .then((text) => {
       expect(text).to.eq(filename);
--- a/packages/document-service-api/src/controllers/documents.js
+++ b/packages/document-service-api/src/controllers/documents.js
@@ -77,10 +77,12 @@ module.exports = {
         data: fileData,
       });
     } else {
+      const fileName = doc.get('name');
       const mimeType = doc.get('mimeType');
       req.log.info({ mimeType }, 'Displaying file');
 
       res.set('content-type', mimeType);
+      res.set('x-original-file-name', fileName);
 
       res.send(fileBuffer);
     }
--- a/packages/forms-web-app/src/controllers/document-service-proxy.js
+++ b/packages/forms-web-app/src/controllers/document-service-proxy.js
@@ -1,0 +1,34 @@
+const { getDocument } = require('../lib/documents-api-wrapper');
+const logger = require('../lib/logger');
+
+exports.getDocument = async (req, res) => {
+  const { appealId, documentId } = req.params;
+
+  if (!appealId || !documentId) {
+    logger.error('Request was missing either an appealId, or a documentId');
+    return res.sendStatus(400);
+  }
+
+  let fileResponse;
+  try {
+    fileResponse = await getDocument(appealId, documentId);
+  } catch (e) {
+    logger.error({ e }, 'Document service proxy lookup failed.');
+    return res.sendStatus(400);
+  }
+
+  try {
+    const fileName = fileResponse.headers.get('x-original-file-name');
+
+    res.set({
+      'content-length': fileResponse.headers.get('content-length'),
+      'content-disposition': `attachment;filename="${fileName}"`,
+      'content-type': fileResponse.headers.get('content-type'),
+    });
+  } catch (e) {
+    logger.error({ e }, 'Document was retrieved but appears to be missing key data.');
+    return res.status(500);
+  }
+
+  return fileResponse.body.pipe(res);
+};
--- a/packages/forms-web-app/src/controllers/document-service-proxy.js
+++ b/packages/forms-web-app/src/controllers/document-service-proxy.js
@@ -27,7 +27,7 @@ exports.getDocument = async (req, res) => {
     });
   } catch (e) {
     logger.error({ e }, 'Document was retrieved but appears to be missing key data.');
-    return res.status(500);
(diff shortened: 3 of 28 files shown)


e2e-tests/cypress/support/appeal-statement-submission/confirmAppealStatementFileIsUploaded.js
```diff
@@ -1,6 +1,6 @@
 module.exports = (filename) => {
   cy.goToAppealStatementSubmission();
-  cy.get('#appeal-upload-file-name')
+  cy.get('[data-cy="appeal-upload-file-link"]')
```
a6software, line 3: all follow the same pattern - swapped from a `span` to an `a` tag nested in the `span`, so now getting the text from the anchor.

packages/document-service-api/src/controllers/documents.js
```diff
@@ -77,10 +77,12 @@ module.exports = {
         data: fileData,
       });
     } else {
+      const fileName = doc.get('name');
       const mimeType = doc.get('mimeType');
       req.log.info({ mimeType }, 'Displaying file');
 
       res.set('content-type', mimeType);
+      res.set('x-original-file-name', fileName);
```
a6software, line 85: pass the original file name through as a header, as this will be useful for 'naming' the file that will be downloaded from FWA

packages/forms-web-app/src/controllers/document-service-proxy.js
```diff
@@ -0,0 +1,34 @@
...
+      'content-disposition': `attachment;filename="${fileName}"`,
+      'content-type': fileResponse.headers.get('content-type'),
+    });
+  } catch (e) {
+    logger.error({ e }, 'Document was retrieved but appears to be missing key data.');
+    return res.status(500);
+  }
+
+  return fileResponse.body.pipe(res);
+};
```
a6software, line 34: This endpoint is deliberately externally vague about what went wrong - just the `400` / `500` response code. 

Internally the logs should be useful in identifying which step failed. If something does go wrong, there are very few reasons this could have happened.

Either the document service API was down, or the appeal doesn't exist (very unlikely) or the document doesn't exist (also highly unlikely). 

One possibility is that the filename is missing - this *should* be highly unlikely going forwards, however it *may* fail on older records. This was the purpose of updating the DB with Paul's script to ensure all the expected fields are on every file upload. 

Finally, the downloaded file should be passed on / piped to the client, so this (afaik) should use as little memory as possible when working with large files <- sanity check please

packages/forms-web-app/src/lib/documents-api-wrapper.js
```diff
@@ -2,10 +2,62 @@ const fs = require('fs');
...
+  const logger = parentLogger.child({
+    correlationId,
+    service: 'Documents Service API',
+  });
+
+  try {
+    logger.debug({ url, method, opts, headers }, 'New call');
+
+    return await utils.promiseTimeout(
+      config.appeals.timeout,
```
a6software, line 23: do we want a specific documents service timeout that differs from the appeals timeout?

packages/forms-web-app/src/middleware/navigation-history.js
```diff
@@ -2,6 +2,7 @@ const config = require('../config');
 
 const defaultOptions = {
   fallbackPath: config.appeals.startingPoint,
+  filterPaths: [/favicon.ico/, /^\/document\//],
```
a6software, line 5: ensure unwanted paths do not get thrown into the 'back' link functionality.

This is a bodge really, as the back link functionality itself is a bodge (imo). 

Perhaps a better system here would be a white list?

packages/forms-web-app/src/middleware/navigation-history.js
```diff
@@ -22,6 +23,11 @@ module.exports = (options = defaultOptions) => (req, res, next) => {
 
   const currentPage = req.baseUrl + req.path;
 
+  if (activeOptions.filterPaths.some((regexFilter) => regexFilter.test(currentPage))) {
+    next();
+    return;
+  }
```
a6software, line 29: don't add unwanted paths to the back link stack.

packages/forms-web-app/src/views/appellant-submission/appeal-statement.njk
```diff
@@ -123,7 +123,9 @@
                 data-file-id="{{ appeal.yourAppealSection.appealStatement.uploadedFile.id }}"
                 data-file-location="{{ appeal.yourAppealSection.appealStatement.uploadedFile.location }}"
                 data-file-size="{{ appeal.yourAppealSection.appealStatement.uploadedFile.size }}">
-              {{- appeal.yourAppealSection.appealStatement.uploadedFile.name -}}
+              <a href="{{ ['/document', appeal.id, appeal.yourAppealSection.appealStatement.uploadedFile.id ] | join('/') }}" title="Download {{ appeal.yourAppealSection.appealStatement.uploadedFile.name }}" data-cy="appeal-upload-file-link">
```
a6software, line 126: all the links follow this pattern of 

`[ 'build', 'link', 'here' ].join('/')`

I gave some thought to this and felt a custom filter or macro would be too specific and the layer of indirection didn't offer much. 

Doing it this way is more verbose, yet more customisable, so it seemed to be preferable.

packages/forms-web-app/src/lib/documents-api-wrapper.js
pbradley-f4: I think keep the one. now poorly named, yes, but for now good enough.

packages/forms-web-app/src/views/appellant-submission/appeal-statement.njk
pbradley-f4: :+1:

packages/forms-web-app/src/controllers/document-service-proxy.js
pbradley-f4: :+1:

pbradley-f4 (approved):
🚢

PR #1338, opened 2021-06-30
fix(lpa-q): filter files on delete

a6software:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/lpa-questionnaire-web-app/src/controllers/upload-question.js
+++ b/packages/lpa-questionnaire-web-app/src/controllers/upload-question.js
@@ -46,6 +46,8 @@ exports.postUpload = async (req, res) => {
       if (deleteId !== 'undefined') await deleteDocument(appealReply.id, deleteId);
     } catch (err) {
       req.log.error({ err }, `Error deleting ${deleteId} from ${name}`);
+    } finally {
+      uploadedFiles = uploadedFiles.filter((file) => file.id !== deleteId);
     }
   } else if (documents.length) {
     // Chance for files to be attached due to non-JS solution, these need to be uploaded (with appropriate errors);


packages/lpa-questionnaire-web-app/src/controllers/upload-question.js
```diff
@@ -46,6 +46,8 @@ exports.postUpload = async (req, res) => {
       if (deleteId !== 'undefined') await deleteDocument(appealReply.id, deleteId);
     } catch (err) {
       req.log.error({ err }, `Error deleting ${deleteId} from ${name}`);
+    } finally {
+      uploadedFiles = uploadedFiles.filter((file) => file.id !== deleteId);
```
a6software, line 50: hopefully evidently, this is not a one line simple fix. 

There are several problems with the current implementation that need addressing. 

This change above fixes a problem with the none JS variant - files are / were not being filtered from the `appealReply` object and so would re-appear on subsequent page reload. This flags a few more issues.

Firstly, I have not had time to investigate this, but it feels like there may be some requirement to concat and de-dupe the two ways the files can be represented:

```
let uploadedFiles = docArrayFromInputString(req.body.tempDocs);
// and
appealReply[sectionName][taskName].uploadedFiles
```

To me it feels like these two concepts need merging, and then any duplicate `id`'s should be filtered. 

This may be a false positive, as above, not had time to investigate further.

Secondly, and requiring more thought is this:

https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/lpa-questionnaire-web-app/src/controllers/files.js#L46

The issue here is that this 'blindly' deletes documents from the document store. Whilst this works, there is no subsequent logic to update the `appealReply` - either in `session` or subsequently in the DB - that this deletion has occurred. The result of this is two fold:

1. Documents appear to be deleted on the front end, but 'mysteriously' re-appear on subsequent hard page refresh;
2. Subsequent calls to delete that already deleted document result in a 404, which appears to do nothing anyway.

The solution here appears to be that same as the none JS equivalent - filter the deleted file from the array of uploaded files.

However. This is non trivial as the `deleteFile` functionality is not being told what type of file is being deleted. In this particularly case I believe we are trying to delete an element from `appealReply.requiredDocumentsSection.plansDecision.uploadedFiles`.

In order for this to work, there needs to be further logic to handle this - left up to the developer implementing to determine the best way to process that. This is made more difficult than it might otherwise initially seem by the use of this hook:

https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/lpa-questionnaire-web-app/src/views/layouts/multi-file-upload.njk#L133

LouiseMcMahon (approved):

PR #1342, opened 2021-07-01
fix(ping-horizon): Changes horizon ping endpoint and frequency

Russ-Clarke:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-2845

## Description of change
The ping was setup to check the availability of the Horizon REST API Wrapper and trigger an alert if it was down. It had been intended that it would ping the wrapper once a minute, but it also pings it once every 10 seconds. The frequent pings are made by the readinessProbe which I have consequently removed from the horizon-ping.yaml file.

Also, sometimes the /horizon endpoint can be unavailable but the /status endpoint still is, meaning messages won't be received from the Appeal Service and we're not alerted to it until a message fails. Therefore, I have changed the endpoint to /horizon on Dev initially.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/charts/functions/templates/horizon-ping.yaml
+++ b/charts/functions/templates/horizon-ping.yaml
@@ -43,10 +43,6 @@ spec:
             httpGet:
               path: /ping
               port: ping-hzn-srv
-          readinessProbe:
-            httpGet:
-              path: /ping
-              port: ping-hzn-srv
           env:
             - name: APP_NAME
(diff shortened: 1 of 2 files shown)


swdpcomputing (dismissed):

swdpcomputing (dismissed):

AnthonyJWalsh (approved):
LGTM

PR #1474, opened 2021-07-20
feat(as-2836): change forms web app to use the business rules for decision date

shepda:
## Ticket Number
https://pins-ds.atlassian.net/browse/AS-2836

## Description of change
Update the Forms Web App to use validation and rules from the business-rules package. This is a like for like swap so if we decide that we need to change the way the validation is done then this can be done in a later ticket.

This change also includes a check for if the deadline date has passed on each page (except the Decision Date page) and if so then the user is taken to the Decision Date Passed end page. To cover the scenario that the user may accidentally enter an incorrect date, the Decision Date Passed end page now as a link to return to the Decision Date page (this page is excluded from the deadline date check).

E2E tests have been updated/added to cover this new scenario as follows:

1. Update the E2E test for the user entering a date past the deadline date, which takes the user to the Decision Date Passed end page and now tests that the user can click a link to return to the Decision Date page.
2. Add an E2E test for the user entering a date past the deadline date, being taken to the Decision Date Past end page and then navigating to any other page, which tests that the user is taken back to the Decision Date Past end page. 

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -150,6 +150,7 @@ services:
       - mongodb
       - clamav
     volumes:
+      - ./packages/business-rules:/opt/app/node_modules/@pins/business-rules
       - ./packages/common:/opt/app/node_modules/@pins/common
       - ./packages/forms-web-app:/opt/app
     command: npm run start:dev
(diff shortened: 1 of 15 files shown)


AnthonyJWalsh (approved):
LGTM

PR #1481, opened 2021-07-21
Subtask/as 2867 magic link screens

adaMiclos:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-2867
Parent ticket https://pins-ds.atlassian.net/browse/AS-58

## Description of change

Add routes, controller and templates for the 'enter your email' and 'email confirmation' pages that are required by the magic link functionality.

## Checklist

- [X] New features have tests

## Important

### DIFF
--- a/packages/lpa-questionnaire-web-app/src/routes/authentication.js
+++ b/packages/lpa-questionnaire-web-app/src/routes/authentication.js
@@ -1,0 +1,13 @@
+const express = require('express');
+const authenticationController = require('../controllers/authentication');
+
+const router = express.Router();
+
+router.get(
+  '/:id/authentication/your-email/:error(session-expired|link-expired)?',
+  authenticationController.showEnterEmailAddress
+);
+router.post('/:id/authentication/your-email', authenticationController.processEmailAddress);
+router.get('/:id/authentication/confirm-email', authenticationController.showEmailConfirmation);
+
+module.exports = router;
--- a/packages/lpa-questionnaire-web-app/tests/unit/controllers/authentication.test.js
+++ b/packages/lpa-questionnaire-web-app/tests/unit/controllers/authentication.test.js
@@ -8,85 +8,87 @@ const mockAppealId = '123';
 const res = mockRes();
 const req = mockReq({}, mockAppealId);
 
-describe('controllers /authentication/your-email', () => {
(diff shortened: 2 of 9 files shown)


packages/lpa-questionnaire-web-app/src/routes/authentication.js
```diff
@@ -0,0 +1,13 @@
+const express = require('express');
```
adaMiclos, line 1: @LouiseMcMahon I know that when we spoked about the endpoints you were thinking to group the pages under the '/authentication/' path but since then the ticket has been updated and now it expects the following endpoints.  

'/your-email'
'/confirm-email'

I have kept your initial idea as I think is better to group the pages together, since they will never be used independently. Also the endpoint is more descriptive. 

We should chat with Pippa tomorrow and find out if this is ok.

swdpcomputing: Agreed, however Pippa on holiday.

packages/lpa-questionnaire-web-app/tests/unit/controllers/authentication.test.js
```diff
@@ -0,0 +1,93 @@
...
+const { mockReq, mockRes } = require('../mocks');
+const { VIEW } = require('../../../src/lib/views');
+
+const mockEmail = 'test.email@test.uk';
+const mockAppealId = '123';
+
+const res = mockRes();
+const req = mockReq({}, mockAppealId);
+
+describe('controllers /authentication/your-email', () => {
```
swdpcomputing, line 11: I think these tests could be under one describe, with each test naming the endpoint.

swdpcomputing (approved):

packages/lpa-questionnaire-web-app/tests/unit/controllers/authentication.test.js
adaMiclos: Yes, sounds good. I have updated them.

LouiseMcMahon (approved):

packages/lpa-questionnaire-web-app/src/routes/authentication.js
adaMiclos: Anthony and the content team has approved this change.

PR #1530, opened 2021-07-27
Feature/as 2984 update google tag manager consent management

AnthonyJWalsh:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-2984

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/lib/client-side/google-tag-manager.js
+++ b/packages/forms-web-app/src/lib/client-side/google-tag-manager.js
@@ -1,0 +1,29 @@
+/* eslint-env browser */
+
+function applyConsent(consent) {
+  window.dataLayer = window.dataLayer || [];
+
+  function gtag() {
+    // eslint-disable-next-line no-undef, prefer-rest-params
+    dataLayer.push(arguments);
+  }
+
+  if (consent === 'denied' || consent === 'granted') {
+    gtag('consent', 'update', {
+      analytics_storage: consent, // We only care about analytics_storage
+    });
+  }
+}
+
+const denyConsent = () => {
+  applyConsent('denied');
+};
+
+const grantConsent = () => {
+  applyConsent('granted');
+};
+
+module.exports = {
+  grantConsent,
+  denyConsent,
+};
--- a/packages/forms-web-app/tests/unit/lib/client-side/javascript-requiring-consent.test.js
+++ b/packages/forms-web-app/tests/unit/lib/client-side/javascript-requiring-consent.test.js
@@ -1,3 +1,8 @@
+/**
+ * @jest-environment jsdom
+ */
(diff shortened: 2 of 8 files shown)


packages/forms-web-app/src/lib/client-side/google-tag-manager.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-env browser */
```
shepda, line 1: Ideally we should have unit tests for this file. The test coverage is failing and adding these tests should make that pass.

packages/forms-web-app/tests/unit/lib/client-side/javascript-requiring-consent.test.js
```diff
@@ -37,39 +46,202 @@ describe('lib/client-side/javascript-requiring-consent', () => {
...
 
       readCookie.mockImplementation(() => JSON.stringify({ usage: false }));
 
-      initialiseOptionalJavaScripts();
+      const config = {
+        server: {
+          googleTagManagerId: 'some-test-value',
+        },
+        featureFlag: {
+          googleTagManager: true, // no googletagmanager feature flag.
```
shepda, line 153: This comment looks like it needs the 'no' removed :)

packages/forms-web-app/src/lib/client-side/google-tag-manager.js
AnthonyJWalsh: Have added tests.

packages/forms-web-app/tests/unit/lib/client-side/javascript-requiring-consent.test.js
AnthonyJWalsh: Have removed 'no'

packages/forms-web-app/src/lib/client-side/google-tag-manager.js
```diff
@@ -0,0 +1,29 @@
...
+
+function applyConsent(consent) {
+  window.dataLayer = window.dataLayer || [];
+
+  function gtag() {
+    // eslint-disable-next-line no-undef, prefer-rest-params
+    dataLayer.push(arguments);
+  }
+
+  if (consent === 'denied' || consent === 'granted') {
```
shepda, line 11: These unit tests look good but the coverage is failing on this conditional due to the else path not being tested.

I just looked into it a bit and this means it either needs 

1. a test to test the scenario where `consent` is set to something other than `denied` or `granted`
2. tell istanbul to ignore the else condition with `/* istanbul ignore else  */` on the line before the conditional

Happy either way because of the way this code is currently written a scenario where `consent` doesn't equal one of these values isn't possible.

shepda (approved):

PR #1704, opened 2021-09-15
feat(adding filename into response header): document download

biroldif:
adding filename into response header

## Ticket Number
https://pins-ds.atlassian.net/browse/AS-2696

## Description of change
adding filename into response header

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/document-service-api/src/controllers/documents.js
+++ b/packages/document-service-api/src/controllers/documents.js
@@ -88,9 +88,11 @@ const serveDocumentById = async (req, res) => {
       return;
     }
 
+    const fileName = metadata.name;
     const mimeType = metadata.mime_type;
     req.log.info({ mimeType }, 'Displaying file');
     res.set('content-type', mimeType);
+    res.set('x-original-file-name', fileName);
     res.send(fileBuffer);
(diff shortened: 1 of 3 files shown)


packages/document-service-api/src/controllers/documents.js
```diff
@@ -88,9 +88,11 @@ const serveDocumentById = async (req, res) => {
       return;
     }
 
+    const fileName = metadata.name;
     const mimeType = metadata.mime_type;
     req.log.info({ mimeType }, 'Displaying file');
     res.set('content-type', mimeType);
+    res.set('x-original-file-name', fileName);
```
shepda, line 95: Why do we need to return the filename here?

biroldif: it can be useful to know the name of the file so that that name can be used to naming the file on download @shepda

shepda: We already have the name of the file in the appeal data in the UI so couldn't we use that for the name of the downloaded file?

shepda (approved):

PR #1717, opened 2021-09-20
Sub task/as 3386 create tests

nssdesikan:
## Ticket Number

Sub-task:- https://pins-ds.atlassian.net/browse/AS-3385
Sub-task:- https://pins-ds.atlassian.net/browse/AS-3386

Feature story:- https://pins-ds.atlassian.net/browse/AS-3291

## Description of change
Requirements of this development is described in the _Feature_ card
A new page is developed to which the users will be redirected when the use selects 'Granted' permission status. Unit tests and E2E tests are also created

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/__tests__/unit/controllers/eligibility/granted-or-refused-permission.test.js
+++ b/packages/forms-web-app/__tests__/unit/controllers/eligibility/granted-or-refused-permission.test.js
@@ -32,6 +32,14 @@ describe('controllers/eligibility/granted-or-refused-permission', () => {
     });
   });
 
+  describe('getGrantedOrRefusedPlanningPermissionOut', () => {
+    it('should call the permission out template', () => {
+      grantedOrRefusedPlanningPermissionController.getGrantedOrRefusedPermissionOut(req, res);
+
+      expect(res.render).toHaveBeenCalledWith(VIEW.ELIGIBILITY.GRANTED_REFUSED_PERMISSION_OUT);
+    });
+  });
+
   describe('postGrantedOrRefusedPlanningPermission', () => {
     it('should re-render the template with errors if there is any validation error', async () => {
       const mockRequest = {
@@ -110,6 +118,30 @@ describe('controllers/eligibility/granted-or-refused-permission', () => {
       });
 
       expect(res.redirect).toHaveBeenCalledWith(`/${VIEW.ELIGIBILITY.DECISION_DATE}`);
+    });
+
+    it(`should redirect to '/${VIEW.ELIGIBILITY.GRANTED_REFUSED_PERMISSION_OUT}' if planniong PermissionStatus is 'granted'`, async () => {
+      const planningPermissionStatusGranted = 'granted';
+      const mockRequest = {
+        ...req,
+        body: { 'granted-or-refused-permission': planningPermissionStatusGranted },
+      };
+      await grantedOrRefusedPlanningPermissionController.postGrantedOrRefusedPermission(
+        mockRequest,
+        res
+      );
+
+      expect(createOrUpdateAppeal).toHaveBeenCalledWith({
+        ...appeal,
+        eligibility: {
+          ...appeal.eligibility,
+          planningPermissionStatus: planningPermissionStatusGranted,
+        },
+      });
+
+      expect(res.redirect).toHaveBeenCalledWith(
+        `/${VIEW.ELIGIBILITY.GRANTED_REFUSED_PERMISSION_OUT}`
+      );
     });
   });
 });
--- a/packages/forms-web-app/__tests__/unit/controllers/eligibility/granted-or-refused-permission.test.js
+++ b/packages/forms-web-app/__tests__/unit/controllers/eligibility/granted-or-refused-permission.test.js
@@ -120,7 +120,7 @@ describe('controllers/eligibility/granted-or-refused-permission', () => {
       expect(res.redirect).toHaveBeenCalledWith(`/${VIEW.ELIGIBILITY.DECISION_DATE}`);
     });
 
-    it(`should redirect to '/${VIEW.ELIGIBILITY.GRANTED_REFUSED_PERMISSION_OUT}' if planniong PermissionStatus is 'granted'`, async () => {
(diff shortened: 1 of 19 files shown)


packages/forms-web-app/src/controllers/eligibility/granted-or-refused-permission.js
```diff
@@ -30,6 +37,12 @@ exports.postGrantedOrRefusedPermission = async (req, res) => {
     return;
   }
 
+  let isPlanningPermissionRefused = null;
+
+  if(validHouseholderPlanningPermissionStatusOptions.includes(planningPermissionStatus)){
+    isPlanningPermissionRefused = planningPermissionStatus.toLowerCase() === 'refused';
```
shepda, line 43: I would suggest that we define these 3 status options in a constants file because even though they are being referenced from the validator, we've started to need to hard-code them in other places.

The validator can also reference the constants file.

For example https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/master/packages/business-rules/src/constants.js

packages/forms-web-app/src/views/eligibility/granted-or-refused-permission-out.njk
```diff
@@ -0,0 +1,18 @@
...
+  You need to use a different service - Eligibility - Appeal a householder planning decision - GOV.UK
+{% endblock %}
+
+{% block content %}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">You need to use a different service</h1>
+      <p class="govuk-body">This is a new service. It is currently only available for appeals about refused planning permission.</p>
+      <p>You can <a data-cy="appeal-decision-service" href="https://acp.planninginspectorate.gov.uk/">appeal using our appeals casework portal</a>.</p>
```
shepda, line 14: This paragraph element should have a class of `govuk-body` and the anchor should have a class of `govuk-link` to keep the typography consistent.

packages/forms-web-app/src/views/layouts/main.njk
```diff
@@ -38,15 +40,17 @@
...
+        {% block backButton %}
+          {{ govukBackLink({
+            text: "Back",
+            href: navigation[1],
+            attributes: {
+              'data-cy': 'back'
+            }
+          }) }}
+        {% endblock %}
+      {% endif %}
```
shepda, line 53: If a block doesn't need to be displayed in a template then an empty block can be added to that template as it will override the default functionality of the block definition, so we don't need to add a conditional to display/hide it.

So for example if you add the following to `granted-or-refused-permission-out.njk` then the back button will be hidden

```
{% block backButton %}
{% endblock %}
```

packages/forms-web-app/src/routes/eligibility/granted-or-refused-permission.js
```diff
@@ -9,6 +9,12 @@ const {
 
 const router = express.Router();
 
+/* Get eligibility granted-or-refused-permission out page */
```
shepda, line 12: I don't think we need these comments in the code to explain the routes, the should be explanatory enough from the route url

packages/forms-web-app/src/routes/eligibility/granted-or-refused-permission.js
```diff
@@ -9,6 +9,12 @@ const {
...
+router.get(
+  '/granted-or-refused-permission-out',
+  grantedOrRefusedPermissionController.getGrantedOrRefusedPermissionOut
+);
```
shepda, line 16: It look like a unit test needs to be added for this new route

packages/forms-web-app/__tests__/unit/controllers/eligibility/granted-or-refused-permission.test.js
```diff
@@ -111,5 +119,29 @@ describe('controllers/eligibility/granted-or-refused-permission', () => {
 
       expect(res.redirect).toHaveBeenCalledWith(`/${VIEW.ELIGIBILITY.DECISION_DATE}`);
     });
+
+    it(`should redirect to '/${VIEW.ELIGIBILITY.GRANTED_REFUSED_PERMISSION_OUT}' if planniong PermissionStatus is 'granted'`, async () => {
```
shepda, line 123: There is a typo here, `planniong` should be `planning` :)

Should `planniong PermissionStatus` be `planningPermissionStatus` or `planning permission status`? It seems to be halfway between the variable name and text description :)

packages/forms-web-app/src/controllers/eligibility/granted-or-refused-permission.js
```diff
@@ -13,8 +20,8 @@ exports.postGrantedOrRefusedPermission = async (req, res) => {
   const { appeal } = req.session;
   const { errors = {}, errorSummary = [] } = body;
 
-  const planningPermissionStatus = body['granted-or-refused-permission'];
-
+  let planningPermissionStatus = body['granted-or-refused-permission'];
```
shepda, line 23: The value of `planningPermissionStatus` isn't being changed so does it need to change from `const` to `let`?

shepda (commented):
There are also linting errors in the code, please review those too.

packages/forms-web-app/src/constants.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  constants: {
```
shepda, line 2: We don't need the additional `constants` level in the object and removing it would reduce the number of params needed to access a value.

packages/forms-web-app/src/constants.js
```diff
@@ -0,0 +1,11 @@
...
+    eligibility: {
+      planningPermissionStatus: {
+        GRANTED: 'granted',
```
shepda, line 5: For consistency for the params it would be better to use either uppercase or camel case but not both.

shepda (commented):
Please also check the linting errors in the following:

lib/views.js
validators/eligibility/granted-or-refused-permission.js

karunak21 (commented):
E2E tests look fine.
Some cypress tests are failing on the branch name, linting errors and few dev comments need to be addressed before this can be merged.

shepda:
Happy to approve once the failing branch name test has been fixed

PR #1718, opened 2021-09-20
test: updated failing cypress tests for lpa-questionnaire

demzzydev:
fix(as-3319): fix for failing test when returning from supplementary docs page to tasklist in lpaq

fix(as-3380): remove mock id from url

fix(as-3360): fix supplementary uploaded document

fix(as-3360): fix supplementary document delete

## Ticket Number
AS-3380
https://pins-ds.atlassian.net/browse/AS-3380

## Description of change
Fixed minor issues with routing

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -23,6 +23,7 @@ on:
       - 'dependabot/**' # Automatic dependabot updates
       - 'feature/**'
       - 'subtask/**'
+      - 'bugfix/**'
       - 'fix/**'
 
 jobs:
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -13,6 +13,7 @@ on:
       - 'dependabot/**' # Automatic dependabot updates
       - 'feature/**'
       - 'subtask/**'
+      - 'bugfix/**'
       - 'fix/**'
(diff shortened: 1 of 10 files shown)


shepda:
Not sure we need this as these look like the changes I did to get the unit tests working in the pipeline and these should already be on this branch?

Also don't think we need to add `bugfix` to the triggers as this is the same as `fix` which is already there?

demzzydev:
> Not sure we need this as these look like the changes I did to get the unit tests working in the pipeline and these should already be on this branch?
> 
> Also don't think we need to add `bugfix` to the triggers as this is the same as `fix` which is already there?

We don't but in order to get over the issue that Shilpa was talking about I had to add an additional flag so that the builds can pass.

shepda:
How does this help the issue, are there specific branches failing because of the prefix? I'd just like to understand a bit more about the reason for adding it.

demzzydev:
> How does this help the issue, are there specific branches failing because of the prefix? I'd just like to understand a bit more about the reason for adding it.

This helps because the check looks for the prefix on the commits and PRs, so even though we change the branch name because we've made commits on a branch with the wrong prefix, the check still fails.

PR #1735, opened 2021-09-23
fix(as-3452): fix document api by modifying object shape

demzzydev:
## Ticket Number
AS-3452
https://pins-ds.atlassian.net/browse/AS-3452

## Description of change
 When uploading on the develop environment, it seems the metadata object within the blob payload is missing. As such this meant that the api was referring to the wrong shape for the document blobs when it came to deleting data. this fix remedies that.

## Checklist

- [ ] New features have tests

## Important


shepda:
Would like to a bit more investigation into why the issue is happening in dev before we merge this

PR #2380, opened 2022-04-22
refactor(pipelines): update existing pipelines to be multi-stage dockerfiles, and remove unused

andylangridge:
Removed all unused pipelines, updated the existing Dockerfiles to be multi-stage, and copy all the
files they need instead of relying on other Docker images.


packages/appeal-reply-service-api/Dockerfile
```diff
@@ -1,30 +1,38 @@
...
+COPY --from=build /opt/app ./
+
+COPY packages/appeals-service-api/__tests__ ./__tests__
+COPY packages/appeals-service-api/jest.config.js ./jest.config.js
+COPY packages/appeals-service-api/jest-mongodb-config.js ./jest-mongodb-config.js
+
 RUN npm run documentation; echo $? > /npm.exitcode
+
+# Stage 3/3: App Run
+FROM node:14-alpine AS app
```
guiribeirokc, line 28: If we are changing the source image here we should change in docker-compose too to keep the sane

packages/business-rules/Dockerfile
```diff
@@ -1,13 +1,28 @@
-FROM node:14-alpine
+# Stage 1/3: Build
```
guiribeirokc, line 1: I'm not sure if we need those business-rules and common as docker image anymore cause now we copy directly from the source when building the other images

packages/appeal-reply-service-api/Dockerfile
andylangridge: This one is actually Alpine in docker-compose! But you're right, I'll check the rest.

PR #2589, opened 2022-08-01
fix(as-4809): fix appeal already submitted functionality

edpackard-work:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-4809

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/__tests__/unit/controllers/full-appeal/submit-appeal/enter-code.test.js
+++ b/packages/forms-web-app/__tests__/unit/controllers/full-appeal/submit-appeal/enter-code.test.js
@@ -27,6 +27,7 @@ describe('controllers/full-appeal/submit-appeal/enter-code', () => {
 	});
 	describe('getEnterCode', () => {
 		it('should render enter code page when receiving the token from email', async () => {
+			req.session.appeal.state = 'DRAFT';
 			const url = '/full-appeal/submit-appeal/request-new-code';
 			calculateDeadline.hasDeadlineDatePassed.mockReturnValue(false);
 			getSavedAppeal.mockReturnValue({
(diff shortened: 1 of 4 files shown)


chrisHowarthkc (dismissed):

packages/forms-web-app/src/controllers/full-appeal/submit-appeal/enter-code.js
```diff
@@ -14,6 +14,10 @@ const getEnterCode = async (req, res) => {
 	const {
 		session: { appeal }
 	} = req;
+
+	if (!appeal || !appeal.state || appeal.state === 'SUBMITTED') {
```
guiribeirokc, line 18: this should be `if ( appeal && appeal.state && appeal.state === 'SUBMITTED)`

guiribeirokc (changes requested):

packages/forms-web-app/__tests__/unit/controllers/full-appeal/submit-appeal/enter-code.test.js
```diff
@@ -35,7 +36,26 @@ describe('controllers/full-appeal/submit-appeal/enter-code', () => {
...
+			expect(res.redirect).toBeCalledWith(`/${APPEAL_ALREADY_SUBMITTED}`);
+		});
+
+		it('should redirect to appeal submitted page if no appeal state', async () => {
+			req.session.appeal.state = null;
+			await getEnterCode(req, res);
+			expect(res.redirect).toBeCalledWith(`/${APPEAL_ALREADY_SUBMITTED}`);
+		});
+
+		it('should redirect to appeal submitted page if no appeal object', async () => {
```
guiribeirokc, line 52: Why it is redirecting to "Appeal already submitted if no appeal was found?

AlexMoulin (dismissed):

packages/forms-web-app/__tests__/unit/controllers/full-appeal/submit-appeal/enter-code.test.js
edpackard-work: once a user submits an appeal, it removes the appeal object from req.session - therefore if we want the 'enter code' link in the email to redirect to the 'appeal already submitted' page once an appeal is submitted (given the 'enter-code' link doesn't contain any information about the specific appeal), we check if the appeal no longer exists. Same logic for the !appeal.state, because if the user has gone back to the before-you-start section, there will be a new appeal object in req.session but it won't have a state field yet. Basically, this is all a workaround for the fact the 'enter-code' link in the email is not actually linked to the specific appeal.

packages/forms-web-app/src/controllers/full-appeal/submit-appeal/enter-code.js
edpackard-work: this wouldn't work, per comment below

PR #2593, opened 2022-08-08
feat(as-5270 update accessibility link in footer): as-5270 update acc…

Constantine-KC:
…essibility link in footer

## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5270

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/config.js
+++ b/packages/forms-web-app/src/config.js
@@ -67,6 +67,5 @@ module.exports = {
 		googleTagManager: process.env.FEATURE_FLAG_GOOGLE_TAG_MANAGER === 'true'
 	},
 	accessibilityLinkUrl:
-		`${process.env.HOST_URL}/accessibility-stamenent` ||
-		'http://localhost:9003/accessibility-statement'
+		`${this.server}/accessibility-statement` || 'http://localhost:9003/accessibility-statement'
 };
--- a/packages/forms-web-app/src/config.js
+++ b/packages/forms-web-app/src/config.js
@@ -65,5 +65,8 @@ module.exports = {
 	},
 	featureFlag: {
 		googleTagManager: process.env.FEATURE_FLAG_GOOGLE_TAG_MANAGER === 'true'
-	}
+	},
+	accessibilityLinkUrl:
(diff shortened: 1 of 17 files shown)


packages/forms-web-app/src/config.js
```diff
@@ -65,5 +65,8 @@ module.exports = {
 	},
 	featureFlag: {
 		googleTagManager: process.env.FEATURE_FLAG_GOOGLE_TAG_MANAGER === 'true'
-	}
+	},
+	accessibilityLinkUrl:
+		`${process.env.HOST_URL}/accessibility-stamenent` ||
```
edpackard-work, line 70: typo

edpackard-work (commented):
should the routers and controllers be unit tested?

packages/forms-web-app/src/config.js
Constantine-KC: Thank you Ed

guiribeirokc: This can be `${server.host}/accessibility-statement`

packages/forms-web-app/src/routes/accessibility-statement/accessibility-statement.js
```diff
@@ -0,0 +1,8 @@
+const express = require('express');
+const accessibilityStatementController = require('../../controllers/accessibility-statement/accessibility-statement');
```
guiribeirokc, line 2: Can you please use the `{ getAccessibilityStatement }`  way?

guiribeirokc (changes requested):

packages/forms-web-app/src/config.js
Constantine-KC: server.host points to localhost:3000 instead of localhost:9003 so had to work around that, and I don't think we can reference an object during initialization when using object literal syntax so that I can use server.host.
I need to have access to the dynamically changing base URL depending on the ENV, which is accomplished by the HOST_URL env variable.

guiribeirokc (approved):

edpackard-work (approved):

PR #2624, opened 2022-09-22
Feature/as 5367 section name to any documents

guiribeirokc:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/horizon-functions/horizon-add-document/__tests__/test.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('gjhgjh', () => {
```
chrisHowarthkc, line 1: Does this test have a purpose or left in by mistake @guiribeirokc ?

packages/forms-web-app/__tests__/unit/controllers/full-appeal/submit-appeal/application-form.test.js
```diff
@@ -27,6 +27,7 @@ describe('controllers/full-appeal/submit-appeal/application-form', () => {
 	const taskName = documentTypes.originalApplication.name;
 	const errors = { 'file-upload': 'Select a file upload' };
 	const errorSummary = [{ text: 'There was an error', href: '#' }];
+	const sectionTag = 'APPLICATION FORM';
```
chrisHowarthkc, line 30: @guiribeirokc potentially out of scope, would it make sense to add the sectionTag to the documentTypes object in common (being used on line 27 already) and call from there for a single source of truth to reference to update should these section tags be the same between appeal types? Rather than hardcoded in each controller, only a thought though.

chrisHowarthkc (approved):

PR #2765, opened 2023-01-26
Integrate cucumber plugin with cypress

willadams199:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5737

## Description of change

Integrating cucumber cypress plugin to platform-feature-tests

## Checklist

- [X] I have updated the documentation accordingly
- [ ] New features have tests

## Important


mlk-kc:
Hey @willadams199 ! This looks sensible however, is there any way we could get a feature file plugged in to see if it all works as expected? I appreciate this is a fair amount of extra code, but I don't know what we have here is actually an integration? What do you think?

mlk-kc:
Morning @willadams199 :) Updates look good! I think we should run a platform test through with a feature file to verify that this is all present and correct, what do you think? I'm happy to do this on a call with you, or if you'd prefer to do a screen record that would be grand as well?

Apologies for the ask here, its just that the platform tests aren't hooked into the delivery pipeline (these tests will take longer to run, and only need to be run once a feature is thought to be complete via feature completeness analysis by the team) so we can't verify via a pull request whether everything works as it once did. Also, since we're adding the feature file layer, it seems prudent that we check to see if that's been set-up correctly with the glue code too?

kwesiplanning (commented):
All looking good. Thanks @willadams199

ob-pins (approved):

edpackard-work (approved):

PR #2815, opened 2023-03-07
feat(as-5769): add route, controller, view for comments-question page

edpackard-work:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5769

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/final-comment/comments-question.js
+++ b/packages/forms-web-app/src/controllers/final-comment/comments-question.js
@@ -1,0 +1,43 @@
+const { VIEW } = require('../../lib/views');
+
+exports.getCommentsQuestion = async (req, res) => {
+	let hasComments = null;
+
+	if (
+		req.session &&
+		req.session.finalComment &&
+		Object.prototype.hasOwnProperty.call(req.session.finalComment, 'hasComments')
+	) {
+		hasComments = req.session.finalComment.hasComments;
+	}
+
+	res.render(VIEW.FINAL_COMMENT.COMMENTS_QUESTION, {
+		hasComments
+	});
+};
+
+exports.postCommentsQuestion = async (req, res) => {
+	const {
+		body,
+		body: { errors = {}, errorSummary = [] }
+	} = req;
+
+	if (Object.keys(errors).length > 0) {
+		return res.render(VIEW.FINAL_COMMENT.COMMENTS_QUESTION, {
+			errors,
+			errorSummary
+		});
+	}
+
+	// DEV ONLY - this check should not be necessary once this page is integrated into the full final-comment journey
+	if (!Object.keys(req.session).includes('finalComment')) {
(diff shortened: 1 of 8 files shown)


packages/forms-web-app/src/controllers/final-comment/comments-question.js
```diff
@@ -0,0 +1,38 @@
...
+
+exports.postCommentsQuestion = async (req, res) => {
+	const {
+		body,
+		body: { errors = {}, errorSummary = [] }
+	} = req;
+
+	if (Object.keys(errors).length > 0) {
+		return res.render(VIEW.FINAL_COMMENT.COMMENTS_QUESTION, {
+			errors,
```
james-hill-kc, line 27: Would it make sense to pass in hasComments here as well? Not sure if we want to persist the user's choice in the event of errors

packages/forms-web-app/src/controllers/final-comment/comments-question.js
```diff
@@ -0,0 +1,38 @@
...
+			errorSummary
+		});
+	}
+
+	const hasComments = body['comments-question'] === 'yes';
+	req.session.finalComment = { hasComments: hasComments };
```
james-hill-kc, line 33: This should probably account for the possibility that the finalComment object could already exist (for example if the user has navigated back from a subsequent step)

packages/forms-web-app/src/views/final-comment/comments-question.njk
```diff
@@ -0,0 +1,61 @@
+{# todo: change layout based on appeal type when component reused #}
+{% extends "layouts/full-appeal-banner/main.njk" %}
+
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set title = "Do you want to submit final comment? - Appeal a planning decision - GOV.UK" %}
```
james-hill-kc, line 8: It feels like this should read "Do you want to submit a final comment" but I can see it matches the copy in the ticket. Perhaps one to check with design?

packages/forms-web-app/src/controllers/final-comment/comments-question.js
edpackard-work: looking at other radio button pages in the full appeal journey, I don't see the user choice being persisted (the typical error would be that they haven't made a choice).

packages/forms-web-app/src/controllers/final-comment/comments-question.js
edpackard-work: i've put in the 'dev only' workaround from your earlier ticket

james-hill-kc (approved):

PR #2820, opened 2023-03-09
feat(as-5749): add route, controller and view for final comment docum…

edpackard-work:
…ents-check page

## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5749

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/final-comment/documents-check.js
+++ b/packages/forms-web-app/src/controllers/final-comment/documents-check.js
@@ -1,0 +1,35 @@
+const { VIEW } = require('../../lib/views');
+
+exports.getDocumentsCheck = async (req, res) => {
+	console.log('getDocumentsCheck', req.session);
+	res.render(VIEW.FINAL_COMMENT.DOCUMENTS_CHECK, {
+		hasDocuments: req.session?.finalComment?.hasDocuments
+	});
+};
+
+exports.postDocumentsCheck = async (req, res) => {
+	console.log('postDocumentsCheck', req.session);
+	const {
+		body,
+		body: { errors = {}, errorSummary = [] }
+	} = req;
+
+	if (Object.keys(errors).length > 0) {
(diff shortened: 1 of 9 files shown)


packages/forms-web-app/src/controllers/final-comment/documents-check.js
```diff
@@ -0,0 +1,43 @@
+const { VIEW } = require('../../lib/views');
+
+exports.getDocumentsCheck = async (req, res) => {
+	//setting hasDocuments here will likely not be necessary when the
+	//final comment object is complete and this page is integrated
+	//into the full final comment journey
+	let hasDocuments = null;
+	if (
```
james-hill-kc, line 8: Just a suggestion, as this is fine, but you might be able to shorten this by using optional chaining, eg. 
`res.render(VIEW.FINAL_COMMENT.DOCUMENTS_CHECK, { req.session?.finalComment?.hasDocuments });`
Because the expression will evaluate to undefined if any of the props in the chain are undefined, I think this should work.

james-hill-kc (approved):

packages/forms-web-app/src/controllers/final-comment/documents-check.js
edpackard-work: this is a good idea, if it works I'll change the `comments-question` controller as well

packages/forms-web-app/src/views/full-appeal/submit-final-comment/final-comment.njk
```diff
@@ -49,7 +49,7 @@
           name: "does-not-include-sensitive-information",
           items: [
             {
-              checked: (hasSensitiveInformation === false),
```
edpackard-work, line 52: removed this as otherwise the checkbox state is hardcoded - hasSensitiveInformation state is now passed from controller

james-hill-kc (approved):

PR #2830, opened 2023-03-23
feat(AS-5759): amendment to fix save and return …

guiribeirokc:
…to be able to continue

amendment to fix save and return to be able to continue

feat#AS-5759

## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/package.json
+++ b/packages/appeals-service-api/package.json
@@ -29,7 +29,6 @@
 		"body-parser": "^1.19.1",
 		"body-parser-csv": "^1.1.0",
 		"compression": "^1.7.4",
-		"crypto-js": "^4.1.1",
 		"csv-parse": "^5.3.0",
 		"date-fns": "^2.28.0",
 		"express": "^4.17.2",
--- a/packages/appeals-service-api/package.json
+++ b/packages/appeals-service-api/package.json
@@ -29,6 +29,7 @@
 		"body-parser": "^1.19.1",
 		"body-parser-csv": "^1.1.0",
 		"compression": "^1.7.4",
+		"crypto-js": "^4.1.1",
 		"csv-parse": "^5.3.0",
 		"date-fns": "^2.28.0",
 		"express": "^4.17.2",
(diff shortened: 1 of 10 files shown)


chrisHowarthkc (dismissed):

packages/appeals-service-api/src/lib/notify.js
```diff
@@ -111,12 +112,12 @@ const sendSaveAndReturnContinueWithAppealEmail = async (appeal) => {
 			appeal.appealType,
 			appeal.eligibility.applicationDecision
 		);
+		const encrypted = CryptoJS.AES.encrypt(appeal.id, '');
```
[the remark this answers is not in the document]
guiribeirokc: Yes you're right I've forgot to remove the unused encrypt. I can actually get the functionality and then we can discuss the need of encrypt later.

packages/business-rules/src/config.js
```diff
@@ -175,12 +175,12 @@ const config = {
...
 						reference: appeal.id
 					}),
-					saveAndReturnContinueAppeal: (appeal, baseUrl, deadlineDate) => ({
+					saveAndReturnContinueAppeal: (appeal, encryptedId, baseUrl, deadlineDate) => ({
 						recipientEmail: appeal.email,
 						variables: {
 							'application number': appeal.planningApplicationNumber,
 							date: format(deadlineDate, 'dd MMMM yyyy'),
-							link: `${baseUrl}/full-appeal/submit-appeal/enter-code`
+							link: `${baseUrl}/appeal-householder-decision/enter-code/${encryptedId}`
```
PaulGruffyddAmaze, line 183: This appears to be using the same URL structure as the householder appeal above. Is that correct for this appeal type?

packages/appeals-service-api/package.json
```diff
@@ -29,6 +29,7 @@
 		"body-parser": "^1.19.1",
 		"body-parser-csv": "^1.1.0",
 		"compression": "^1.7.4",
+		"crypto-js": "^4.1.1",
```
PaulGruffyddAmaze, line 32: Is this still being used? If not, we should remove it.

guiribeirokc:
The changes are here: packages/business-rules/src/config.js
We don't need to change the template

PaulGruffyddAmaze (approved):
I think that addresses all of my comments so gets a thumbs up from me.

PR #2837, opened 2023-03-28
Feature/as 5786 processing final comments upload documents

edpackard-work:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/lib/find-target-value-in-json.js
```diff
@@ -0,0 +1,12 @@
+const findTargetValueInJSON = (obj, target) => {
+	return target in obj
+		? obj[target]
+		: Object.values(obj).reduce((acc, val) => {
+				if (acc !== undefined) return acc;
+				if (typeof val === 'object' && val) return findTargetValueInJSON(val, target);
+		  }, undefined);
+};
```
MarkTaylor-KandC, line 8: There is quite a lot going on in this single statement, it would be good if we could simplify this a bit:

I have had a bit of a go at it although this is written in github rather than tested in an IDE so probably needs a few tweaks 

```js
const findTargetValueInJSON = (obj, target) => {
  if (target in obj) {
    return obj[target];
  }
  
  for (const value of Object.values(obj)) {
    if (typeof value === 'object' && value) {
      const result = findTargetValueInJSON(value, target);
      if (result !== undefined) {
        return result;
      }
    }
  }

  return undefined;
};
```

packages/forms-web-app/src/lib/multifile-upload.js
```diff
@@ -0,0 +1,44 @@
...
+	return {
+		id: document.id,
+		name: file.name,
+		// needed for MoJ multi-file upload display
+		message: {
+			text: file.name
+		},
+		fileName: file.name,
+		originalFileName: file.name,
+		// needed for Cypress testing
```
MarkTaylor-KandC, line 34: It would be good to dig into this need a bit more, ideally we wouldn't be changing the production code to make it easier to test, although I understand that sometimes this is needed so just wanted to make sure we had explored all alternatives to this.

packages/forms-web-app/src/validators/common/schemas/check-document-uploaded-schema.js
```diff
@@ -0,0 +1,38 @@
+const { findTargetValueInJSON } = require('../../../lib/find-target-value-in-json');
+
+const schema = (path, documentType, submissionType) => ({
+	[path]: {
+		custom: {
+			options: async (value, { req }) => {
+				if (!req.files) {
+					let uploadedFiles;
+
```
MarkTaylor-KandC, line 9: I think we could simplify the selection of the uploaded files somewhat (again tried in github not an IDE :) )
```js
const uploadedFiles = req.session[submissionType]?.[documentType]?.uploadedFiles;
```

packages/forms-web-app/src/lib/multifile-upload.js
edpackard-work: good call - I will investigate today - this object (and comment) is a cut and paste extraction from existing code so I don't know the history: we probably don't need to keep this in, especially as we're not using the cypress e2e tests.

packages/forms-web-app/src/lib/find-target-value-in-json.js
edpackard-work: looks like this works Mark - thanks, much clearer

packages/forms-web-app/src/validators/common/schemas/check-document-uploaded-schema.js
edpackard-work: this would work for final comments, where the uploaded documents appear in the location you indicate - i.e. `req.session.finalComment.supportingDocuments.uploadedFile`, but in an appeal object, uploaded documents have an extra layer of nesting , i.e. `req.session.appeal.planningApplicationDocumentsSection.plansDrawingsSupportingDocuments.uploadedFiles`, the `findTargetValueInJSON` function is a way around this.

alternatively, for appeals, I believe that [sectionName] is part of the req for specific pages, so this could be used - i.e. if `req.session[submissionType]?.[documentType]?.uploadedFiles` is undefined, try ``req.session[submissionType]?.[sectionName]?.[documentType]?.uploadedFiles` - but then we're becoming quite tied to the structure of the objects?

packages/forms-web-app/src/lib/multifile-upload.js
edpackard-work: I think you are right that this needs digging into more - how many of the fields that we populate with data from the document API response (after successfully uploading a file) do we actually need? There is also a comment in the `document.js` controller along the lines of 'should we be returning any document data beyond the document's ID upon successful upload?' - this data is available via the `serveDocumentById()` function, accessible via a get request. I would suggest adding this to the tech debt ticket for further investigation.

edpackard-work: I also found an existing mapper in `src/mappers/document-mapper`, so have adapted that to include the required field for multi-file uploads and removed the mapping function from this module.

packages/forms-web-app/src/lib/multifile-upload.js
```diff
@@ -0,0 +1,27 @@
+const getValidFiles = (errors, files) => {
```
james-hill-kc, line 1: I think this file could potentially have a more specific name to better describe its purpose

james-hill-kc (dismissed):

packages/forms-web-app/src/lib/multifile-upload.js
edpackard-work: changed to `multi-file-upload-helpers.js`

chrisHowarthkc (approved):

james-hill-kc (approved):

PR #2854, opened 2023-04-16
Fix/as 5759 save and return fixes

edpackard-work:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5759

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/middleware/check-path-allowed.js
```diff
@@ -9,6 +9,10 @@ const checkPathAllowed = (req, res, next) => {
 		'/before-you-start/use-existing-service-local-planning-department',
 		'/appellant-submission/submission-information',
 		'/full-appeal/submit-appeal/declaration-information',
+		'/full-appeal/submit-appeal/code-expired',
+		'/full-appeal/submit-appeal/request-new-code',
+		'/appeal-householder-decision/code-expired',
+		'/appeal-householder-decision/request-new-code',
```
PaulGruffyddAmaze, line 15: For consistency, should these URL exemptions not be handled using the `skipMiddlewareForPaths` middleware in the same way as the enter code pages?

edpackard-work: possibly - that middleware is used for `enter-code` because the url is different depending on id, whereas these urls - like the others in the list - are fixed.

chrisHowarthkc (approved):

PR #2857, opened 2023-04-20
fix: set google tag manager default consent setting

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5870

## Description of change

Google tag manager needs a default cookie consent value set before the script is run, this has been added to head.njk

The js that sets consent then needs to have access to the gtm id so that has been passed through in compilation using webpack.

javascript-requiring-consent.js is refactored with the intention of having a clearer distinction of what occurs when the feature flag is toggled

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/__tests__/unit/lib/client-side/javascript-requiring-consent.test.js
+++ b/packages/forms-web-app/__tests__/unit/lib/client-side/javascript-requiring-consent.test.js
@@ -4,9 +4,11 @@ const {
 
 const { readCookie } = require('../../../../src/lib/client-side/cookie/cookie-jar');
 const { initialiseGoogleAnalytics } = require('../../../../src/lib/client-side/google-analytics');
+const googleTagManager = require('../../../../src/lib/client-side/google-tag-manager');
 
 jest.mock('../../../../src/lib/client-side/cookie/cookie-jar');
 jest.mock('../../../../src/lib/client-side/google-analytics');
+jest.mock('../../../../src/lib/client-side/google-tag-manager');
 
 describe('lib/client-side/javascript-requiring-consent', () => {
 	describe('initialiseOptionalJavaScripts', () => {
@@ -21,6 +23,7 @@ describe('lib/client-side/javascript-requiring-consent', () => {
 			expect(console.log).toHaveBeenCalledWith('Consent not yet given for optional JavaScripts.');
 
 			expect(initialiseGoogleAnalytics).not.toHaveBeenCalled();
+			expect(googleTagManager.grantConsent).not.toHaveBeenCalled();
 		});
 
 		test('return early if `usage` is not defined', () => {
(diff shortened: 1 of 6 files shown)


packages/forms-web-app/__tests__/unit/lib/client-side/javascript-requiring-consent.test.js
```diff
@@ -31,29 +34,90 @@ describe('lib/client-side/javascript-requiring-consent', () => {
...
-		test('return early if `usage=false`', () => {
-			jest.spyOn(console, 'log').mockImplementation();
+		describe('tagmanager active', () => {
+			const OLD_ENV = process.env;
 
-			readCookie.mockImplementation(() => JSON.stringify({ usage: false }));
+			beforeEach(() => {
+				jest.resetModules();
+				jest.resetAllMocks();
+				process.env = OLD_ENV;
```
AlexMoulin, line 46: I don't think you need both line 46 and 47, just line 47?

packages/forms-web-app/__tests__/unit/lib/client-side/javascript-requiring-consent.test.js
```diff
@@ -31,29 +34,90 @@ describe('lib/client-side/javascript-requiring-consent', () => {
 			initialiseOptionalJavaScripts();
 
 			expect(initialiseGoogleAnalytics).not.toHaveBeenCalled();
+			expect(googleTagManager.grantConsent).not.toHaveBeenCalled();
 		});
 
-		test('return early if `usage=false`', () => {
-			jest.spyOn(console, 'log').mockImplementation();
+		describe('tagmanager active', () => {
+			const OLD_ENV = process.env;
```
AlexMoulin, line 41: Doesn't this create a shallow copy? So if you change one you change the other?

packages/forms-web-app/__tests__/unit/lib/client-side/javascript-requiring-consent.test.js
```diff
@@ -31,29 +34,90 @@ describe('lib/client-side/javascript-requiring-consent', () => {
...
+
+				expect(googleTagManager.grantConsent).not.toHaveBeenCalled();
+				expect(initialiseGoogleAnalytics).not.toHaveBeenCalled();
+			});
 		});
 
-		test('calls through if `usage=true`', () => {
-			readCookie.mockImplementation(() => JSON.stringify({ usage: true }));
+		describe('tagmanager inactive', () => {
+			const OLD_ENV = process.env;
```
AlexMoulin, line 89: Another shallow copy?

packages/forms-web-app/__tests__/unit/lib/client-side/javascript-requiring-consent.test.js
```diff
@@ -31,29 +34,90 @@ describe('lib/client-side/javascript-requiring-consent', () => {
...
 
-			initialiseOptionalJavaScripts();
+			beforeEach(() => {
+				jest.resetModules();
+				jest.resetAllMocks();
+				process.env = OLD_ENV;
+				process.env = { ...OLD_ENV, googleTagManager: false };
```
AlexMoulin, line 95: Probably don't need line 94?

packages/forms-web-app/__tests__/unit/lib/client-side/javascript-requiring-consent.test.js
bryn-pins: Yes
But then process.env is set to a completely new object so becomes a reference to something else
The old object + ref to it with OLD_ENV remains unchanged.

Just double checked with the following:
`
var env = {
  a: 1
}

var OLD_ENV = env;

env = { b: 2 }
`

Will be clearer I think when I remove the pointless line you've spotted.

Found an existing pattern from cookie-jar.test.js which this is borrowed from

AlexMoulin (approved):

PR #2881, opened 2023-06-01
feat(as-5798): middleware to restrict access to final comments, debug…

bryn-pins:
…debug page to override deadline

## Ticket Number

https://pins-ds.atlassian.net/browse/as-5798
https://pins-ds.atlassian.net/browse/as-5842

## Description of change

A middleware function that only allows access to final comments routes if the deadline has not passed and code has been entered

A debug route that sets a value in session that will allow testers to bypass the deadline check.
Submitting would still fail as the API validates the deadline on submission, and the original date will not be updated by the debug route

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/middleware/check-debug-allowed.js
+++ b/packages/forms-web-app/src/middleware/check-debug-allowed.js
@@ -1,0 +1,12 @@
+const config = require('../config');
+
+const checkDebugAllowed = (req, res, next) => {
+	if (config.server.allowTestingOverrides) {
+		return next();
+	}
+
+	res.status(404);
+	return res.render('error/not-found');
+};
+
+module.exports = checkDebugAllowed;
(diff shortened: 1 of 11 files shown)


packages/forms-web-app/src/middleware/check-debug-allowed.js
```diff
@@ -0,0 +1,12 @@
+const config = require('../config');
+
```
chrishawldev, line 2: Nice, I like this approach. The only other thing I'd think about is if we always want to disallow debug mode on production environments. For example if domain contains myproddomain.com then prevent debug mode being enabled even if it's set to true. We could make that an application configuration value and support an array of domains as well, just so we have an additional mechanism in place to prevent potentially enabling debug mode in production.

chrishawldev (approved):
Looks good to me. In terms of my comment about hiding the debug feature based on domain, we can achieve the same through good communication and / or build pipelines (e.g prevent deployment to prod if debug enabled == true), just a thought more than anything else.

packages/forms-web-app/src/middleware/check-debug-allowed.js
bryn-pins: Seems sensible, might try and do that in a separate PR at some point

PR #2894, opened 2023-06-12
fix(token.js, token.test.js): token entropy improvement

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5942

## Description of change

Use crpytographically secure methods of random token generation

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/src/lib/token.js
```diff
@@ -1,10 +1,12 @@
...
+const BUFFER_LENGTH = 20;
 const createToken = () => {
-	const token = [];
-	for (let i = 0; i < 5; i += 1) {
-		const num = Math.floor(Math.random() * 9 + 1);
-		token.push(num);
-	}
-	return token.join('');
+	return crypto
+		.randomBytes(BUFFER_LENGTH)
```
PaulGruffyddAmaze, line 6: Nice. I like the approach and I can see you've allowed for ~20 more characters than we need, presumably to account for the fact that  we might get a few `+`s or `/`s amongst the characters in the base64 encoded string. I'm wondering whether we need a length check on the string before we return it though. Much as it's highly unlikely, there is a chance we could get over 20 non alpha-numeric characters so theoretically could end up with a string containing less than 5 characters.

packages/appeals-service-api/src/lib/token.js
```diff
@@ -1,10 +1,12 @@
...
+		.toString('base64')
+		.replace(/[^\w_]/g, '')
+		.slice(0, TOKEN_LENGTH);
```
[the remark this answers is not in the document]
chrishawldev: Is it worth picking a longer verification string as a quick way to avoid more profanities than not? According to this there are no profanities that are 9 characters long https://www.wired.com/2016/09/science-swear-words-warning-nsfw-af/
Could also allow special characters - but that may cause more problems than it's worth if we ever pass them in a url or there are any caveats in terms of queries with the mongo sdk.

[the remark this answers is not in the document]
suityou01: > Is it worth picking a longer verification string as a quick way to avoid more profanities than not? According to this there are no profanities that are 9 characters long https://www.wired.com/2016/09/science-swear-words-warning-nsfw-af/ Could also allow special characters - but that may cause more problems than it's worth if we ever pass them in a url or there are any caveats in terms of queries with the mongo sdk.

You could remove the vowels?

packages/appeals-service-api/src/lib/token.js
suityou01: How should it work in the failure case? Should it try again with the same algorithim or just return a 5 digit random number as a failover?

PaulGruffyddAmaze: I think that scenario would be sufficiently rare that you could try again and be pretty confident of a 5 digit code the next time.

packages/appeals-service-api/src/lib/token.js
```diff
@@ -1,10 +1,24 @@
...
+
+	if (token.length === 5) return token;
+
+	token = crypto
+		.randomBytes(BUFFER_LENGTH)
+		.toString('base64')
+		.replace(/[^\w_]/g, '')
+		.replace(/[AEIOUaeiou0]/g, '')
+		.slice(0, TOKEN_LENGTH);
```
PaulGruffyddAmaze, line 19: As this is a repeat of the token generation from above, it might be worth separating this into its own function to avoid duplication.

chrishawldev (commented):
LGTM

chrishawldev (approved):
LGTM

PR #2898, opened 2023-06-14
fix(as-xxxx): updating getCountry function to allow LPA codes instead…

chrisHowarthkc:
… of ONS

## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/models/entities/lpa-entity.js
+++ b/packages/appeals-service-api/src/models/entities/lpa-entity.js
@@ -52,7 +52,7 @@ class LpaEntity {
 		return this.#inTrial;
 	}
 	getCountry() {
-		const isWales = new RegExp('W[0-9]{8}');
+		const isWales = new RegExp('^W[0-9]{8}$');
 		if (isWales.test(this.#lpa19CD)) {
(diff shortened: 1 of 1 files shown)


packages/appeals-service-api/src/models/entities/lpa-entity.js
```diff
@@ -1,66 +1,77 @@
...
-            return "Wales"
-        } else {
-            return ""
-        }
-    }
+	isInTrial() {
+		return this.#inTrial;
+	}
+	getCountry() {
+		const isWales = new RegExp('W[0-9]{8}');
```
PaulGruffyddAmaze, line 55: Probably not an issue but this regex would check for any string which contains a W followed by 8 numbers, for example, `xxxxW1234567890`. It could be made more specific by defining the start and end of the string like this:
`const isWales = new RegExp('^W[0-9]{8}$');`

edpackard-work (approved):

edpackard-work (approved):

edpackard-work (approved):

PaulGruffyddAmaze (approved):

PR #2902, opened 2023-06-19
fix(as-5913): fix confirm email success box appearance

edpackard-work:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


chrishawldev (approved):

packages/appeals-service-api/src/controllers/token.js
```diff
@@ -42,8 +42,6 @@ async function tokenPost(req, res) {
 	}
 
 	if (document.attempts && document.attempts > 3) {
-		// reset attempts and code
-		await createOrUpdateTokenDocument(id);
 		return res.status(429).send({});
```
chrishawldev, line 45: I would leave this in for now because although this mechanism isn't ideal it is intended to stop brute force attacks to guess the code. Maybe create a separate issue to solve this problem?

edpackard-work: I can leave it in, but brute force attacks are still prevented without it - once the user attempts for a particular token object are above 3, the user cannot proceed even if they enter the correct code: they will shown the 'you need a new code' page.

edpackard-work: i've restored it - I have had to add the action property to prevent the action being set as 'null' if a user makes too many requests

chrishawldev (approved):

PR #2908, opened 2023-06-27
Feature/as 5882 lpa account enter code

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AS-5882

## Description of change

Add flow for LPA user to enter the code sent to them by email

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -131,7 +201,97 @@ const postEnterCode = (views) => {
...
+				logger.error(e);
+				return {
+					valid: false,
+					action: enterCodeConfig.actions.lpaDashboard
+				};
+			}
+		}
+
+		// check token
+		let token = await isTokenValid(id, emailCode, req.session, id);
```
bryn-pins, line 275: Don't think the 4th param is used here, unless I've missed something

packages/forms-web-app/src/lib/appeals-api-wrapper.js
```diff
@@ -149,6 +149,15 @@ exports.createUser = async (email, isAdmin, lpaCode) => {
 	});
 };
 
+exports.getUserById = async (id) => {
+	console.log(`/api/v1/users/${id}`);
```
bryn-pins, line 153: can remove this console log?

packages/forms-web-app/src/lib/is-token-valid.js
```diff
@@ -1,45 +1,54 @@
...
-	if (!id || typeof id !== 'string' || !token || typeof token !== 'string') {
-		return result;
-	}
-
-	// is test LPA + Test environment + test token
-	// then allow through as a confirm email token
-	if (
+// is test LPA + Test environment + test token
+// then allow through as a confirm email token
+const isTestLPAAndEnvironment = (token, session) => {
```
bryn-pins, line 10: think this should be renamed to include that it is checking the token, or perhaps a suggestion to be separated into 2 functions if name is too unwieldy

packages/forms-web-app/src/validators/common/check-id-is-email-address.js
bryn-pins, line 1: is this being used?

packages/forms-web-app/src/validators/lpa/enter-code.js
bryn-pins, line 1: More of a question, this is taking into account the changes to the code in the other PR

packages/forms-web-app/src/controllers/common/enter-code.js
suityou01: Or the 3rd. I wonder if it read req.session.id at some point... Good spot.

packages/forms-web-app/src/validators/lpa/enter-code.js
bryn-pins: Sorry not clear: https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/2894

suityou01: I have not integration tested it, but as the variable that is storing the code value is a string I don't see any issue. What did I miss?

suityou01: Phew! :-)

bryn-pins:
Think this looks good, only concern is we're not following the same pattern as the other enter code pages of sending the token to the user's email when hitting the GET for enter-code

Think it is perhaps more logical not to, but then will need to ensure the token is sent before we get to this page, including on the various, need-new-code, code expired etc. routes

packages/forms-web-app/src/routes/lpa-dashboard/lpa-dashboard.js
```diff
@@ -7,6 +7,7 @@ const { getServiceInvite } = require('../../../src/controllers/lpa-dashboard/ser
 const router = express.Router();
 
 router.use(featureFlagMiddleware('lpa-dashboard'));
+
 router.use(
 	skipMiddlewareForPaths(requireUser, [
 		'input-code',
```
chrisHowarthkc, line 13: Should 'input-code' be replaced with 'enter-code' for this journey?

bryn-pins:
Ok so if we're keeping the same as the other enter token pages, I think we need to actually generate the token on the get for enter-code

So a call to: sendToken(userId, action, userEmail)

In order to get userEmail, we can lookup via the getUserById api method for now

bryn-pins:
Happy to approve given outside conversation once conflicts are resolved

bryn-pins (approved):

packages/forms-web-app/src/lib/is-token-valid.js
```diff
@@ -1,45 +1,52 @@
...
 		} else {
 			throw err;
 		}
 	}
+};
+
+const isTokenValid = async (id, token) => {
+	let result = {
+		valid: false,
+		action: ''
```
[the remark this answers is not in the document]
chrishawldev: Cool, I can see this is handled here const getToken = async (id, token, session) https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/2908/files#diff-732692fed0da9475bd6db5d66bc33777d85ec9b06a11e34f2bfa883cff722e2bR29
if (err.message === 'Too Many Requests') {
			return {
				valid: false,
				action: session?.enterCode?.action,
				tooManyAttempts: true
			};

chrisHowarthkc (approved):

PR #2942, opened 2023-07-12
Feature/aapd 34 add user to database

edpackard-work:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/__tests__/developer/users.test.js
```diff
@@ -202,6 +202,21 @@ describe('Users', () => {
 		expect(userResponse.body.enabled).toBe(true);
 	});
 
+	it('should set status flag as added for user added by admin', async () => {
+		// given a non-admin user is added
+		const email = 'testuser1@planninginspectorate.gov.uk';
```
chrishawldev, line 207: The doamin for test emails should be @exmaple.com just as an extra guard rail to be avoid ever sending out emails when running tests to a genuine domain and because @example.com is a reserved doamin commonly used for documentation etc

packages/appeals-service-api/src/services/user.service.js
```diff
@@ -61,10 +62,11 @@ const createUser = async (user) => {
 		throw ApiError.badRequest();
 	}
 
-	user.enabled = true;
+	user.enabled = true; //todo: decide what to do with this field in light of status field
 	user.createdAt = new Date();
 
 	if (!user.isAdmin) {
+		user.status = 'added';
```
chrishawldev, line 69: This would be better as a string constant somewhere that we use across files so that if we change the value it will change everywhere, it also makes the code and tests less brittle if there is ever a typo in  a string for example.
Instead something like user.status = userStatus.added

Javascript doesn't properly support enum I don't think, but in most other languages this would be an enum.

chrishawldev (commented):
A few bits of feedback, nothing major

packages/appeals-service-api/__tests__/developer/users.test.js
edpackard-work: will check, but I think the domain has to be planninginspectorate.gov.uk for the integration test to work properly

chrishawldev: Cool. In the short term if this a bigger change than just changing that email address then don't worry about it. We'll just flag it as a dev task.

edpackard-work: fixed it - all domains are example.com on this test now

bryn-pins (approved):

PR #2944, opened 2023-07-13
Feature/aapd 13 db status removed

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-13

## Description of change

Add remove status to constants

## Checklist

- [X] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/services/user.service.js
+++ b/packages/appeals-service-api/src/services/user.service.js
@@ -37,7 +37,10 @@ const getUsers = async (lpaCode) => {
 	}
 
 	try {
-		const cursor = await mongodb.get().collection('user').find({ lpaCode: lpaCode, enabled: true });
+		const cursor = await mongodb
+			.get()
+			.collection('user')
+			.find({ lpaCode: lpaCode, enabled: true, status: 'confirmed' });
 		// to sort from mongo instead of locally
 		// .sort({ isAdmin: -1, email: 1 });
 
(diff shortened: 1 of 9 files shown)


packages/appeals-service-api/src/services/user.service.js
```diff
@@ -183,7 +188,7 @@ const setUserStatus = async (id, status) => {
 			.collection('user')
 			.findOneAndUpdate(
 				{ _id: new ObjectId(id) },
-				{ $set: { updatedAt: new Date(), status: status } },
+				{ $set: { confirmedAt: new Date(), status: status } },
```
bryn-pins, line 191: Would suggest confirmedAt is only set if status is confirmed, so it's not overwritten by other status changes

Something like:

```
let fieldsToSet = { status: status };

	if (status === STATUS_CONSTANTS.CONFIRMED) {
		fieldsToSet.confirmedAt = new Date();
	}
```

chrishawldev (dismissed):
LGTM

packages/appeals-service-api/src/services/user.service.js
bryn-pins: Or can merge and do later, as I'm waiting on those statuses for one of my tickets

suityou01: should there be other datetimestamps on this document like "removedAt"? Or is it better to stick with "updatedAt"?

bryn-pins: Maybe just leave as is for now and can figure out in one of my PRs as I'm not sure
Only business requirement for these is to order by the confirmed at date

bryn-pins (approved):

bryn-pins (approved):

PR #2945, opened 2023-07-13
feat(aapd-14): add confirm-remove-users page

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aapd-14

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/__tests__/unit/routes/lpa/confirm-remove-user.test.js
+++ b/packages/forms-web-app/__tests__/unit/routes/lpa/confirm-remove-user.test.js
@@ -1,0 +1,26 @@
+const { get, post } = require('../router-mock');
+const confirmAddUserController = require('../../../../src/controllers/lpa-dashboard/confirm-remove-user');
+
+jest.mock('../../../../src/controllers/lpa-dashboard/confirm-remove-user');
+
+describe('routes/lpa-dashboard/confirm-remove-user', () => {
+	beforeEach(() => {
+		// eslint-disable-next-line global-require
+		require('../../../../src/routes/lpa-dashboard/confirm-remove-user');
+	});
+
+	afterEach(() => {
+		jest.resetAllMocks();
+	});
+
+	it('should define the expected routes', () => {
+		expect(get).toHaveBeenCalledWith(
+			'/confirm-remove-user/:id',
+			confirmAddUserController.getConfirmRemoveUser
+		);
+		expect(post).toHaveBeenCalledWith(
+			'/confirm-remove-user/:id',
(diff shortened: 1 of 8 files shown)


packages/forms-web-app/__tests__/unit/routes/lpa/confirm-remove-user.test.js
```diff
@@ -0,0 +1,26 @@
...
+		require('../../../../src/routes/lpa-dashboard/confirm-remove-user');
+	});
+
+	afterEach(() => {
+		jest.resetAllMocks();
+	});
+
+	it('should define the expected routes', () => {
+		expect(get).toHaveBeenCalledWith(
+			'/confirm-remove-user/:id',
```
chrishawldev, line 18: I mentioned this on another PR earlier, and I guess this was the point raised re the value of testing the routes. Personally I think there is very limited value in testing this as we're basically just testing the express router. Not a blocker to passing review, just a talking point.

bryn-pins: I agree and is one of the things I had mentioned on our dev standards call. 
Happy to remove once we decide that's what we're doing

packages/forms-web-app/__tests__/unit/views/lpa/confirm-remove-user.njk.test.js
```diff
@@ -0,0 +1,43 @@
...
+			);
+
+			console.log(document.body.innerHTML);
+
+			expect(document.querySelector('#main-content h1').textContent.trim()).toEqual(
+				'Confirm you want to remove this user'
+			);
+		});
+
+		it('should render the email address as expected', () => {
```
chrishawldev, line 30: I'm also not sure on these tests, are we just testing the nun jucks rendering engine and the gov uk templates here in places? I think the main value is in testing the the nun jucks renderer has been called with the expected values that we pass in. Once we integrate the cypress tests we might also be duplicating effort here. Again not a blocker just food for thought.

chrishawldev (approved):
Approving with comments

packages/forms-web-app/__tests__/unit/views/lpa/confirm-remove-user.njk.test.js
bryn-pins: We previously were doing snapshots of the views, which I did think was pointless as you just updated the snapshots mindlessly
Charlie started adding tests like this and felt it was a little more useful.
Happy to remove going forward.
This was some of the things I wanted to raise so glad you largely agree

PR #2951, opened 2023-07-17
feat(aapd-17): set up notify to send lpa dashboard invite email

edpackard-work:
]

## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/__tests__/unit/lib/notify.test.js
```diff
@@ -1,58 +1,102 @@
...
-					})
-			)
-		})
-	};
-	return {
-		//__esModule: true,
-		...originalModule,
-		...newModule
-	};
+jest.mock('@pins/common/src/lib/notify/notify-builder', () => ({
```
suityou01, line 11: Nice :-)

packages/appeals-service-api/src/lib/notify.js
```diff
@@ -236,6 +236,38 @@ const sendFailureToUploadToHorizonEmail = async (appealId) => {
...
+			.setDestinationEmailAddress(recipientEmail)
+			.setTemplateVariablesFromObject(variables)
+			.setReference(reference)
+			.sendEmail(
+				config.services.notify.baseUrl,
+				config.services.notify.serviceId,
+				config.services.notify.apiKey
+			);
+	} catch (err) {
+		logger.error({ err, userId: user._id }, 'Unable to send LPA dashboard invitation email');
```
suityou01, line 267: Do you really want to swallow the error here?

chrishawldev: Yeah if we're triggering this from some downstream client it would be good to pass the exception back up the call stack if there is an opportunity to give an end user or whatever triggers this a chance to replay it or at least tell them it's failed. Got to be honest I haven't looked at the notify stuff at all so that might not be relevant in this instance.
Also, again with the caveat of not knowing how the notify sdk works, so please feel free to ignore this, but from line 256 onwards seems like the kind of thing you'd usually look to bootstrap once and not have to worry about each time you called it.

edpackard-work: the other notify calls are done within a try/catch, so continued that here.

can look at refactoring the await NotifyBuilder calls - there are quite a few of them in the notify.js file

bryn-pins (approved):
I'm happy, if we wanted to refactor notify bits would make sense to do in a separate PR

PR #2962, opened 2023-07-21
feat(appeals-api): added endpoint get api/v1/appeals/:lpaCode:/:id

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-47/AAPD-89

## Description of change

Added end point to appeals api to retrieve appeal by lpacode and id

## Checklist

- [x] New features have tests

## Important


bryn-pins (commented):
Sorry if mistaken but should this instead be using the new appealsCaseData collection instead?
My understanding was we were planning to have a new collection for data coming back from the back office from the appeals we collect from frontend and we would use that for lpa-dashboard appeals.
If this is for another purpose obviously ignore this

suityou01:
> Sorry if mistaken but should this instead be using the new appealsCaseData collection instead? My understanding was we were planning to have a new collection for data coming back from the back office from the appeals we collect from frontend and we would use that for lpa-dashboard appeals. If this is for another purpose obviously ignore this

Yes I think I misunderstood the requirements rather. Where do I get the data from the interim?

bryn-pins:
> > Sorry if mistaken but should this instead be using the new appealsCaseData collection instead? My understanding was we were planning to have a new collection for data coming back from the back office from the appeals we collect from frontend and we would use that for lpa-dashboard appeals. If this is for another purpose obviously ignore this
> 
> Yes I think I misunderstood the requirements rather. Where do I get the data from the interim?

So data will eventually come in from service bus topics and end up in a collection called appealsCaseData.
We're going to fake that data for now by just manually adding data to collection.

I have added an initial example Paul gave us into a new collection in the db seed data, feel free to add more to that array if you need specific data for different scenarios: (004-appealsCaseData.json). We may need to manually add to the test db as well to allow for testing.

Ed has added an initial GET:many route to the api which has gone into main.
Although I have suggested we rename some of it to avoid the references to lpa-dashboard-appeals

Could either wait for that rename work
Or add the GET:id action to existing code, but that will require a reasonable merge/changes at some point from someone

suityou01:
> > > Sorry if mistaken but should this instead be using the new appealsCaseData collection instead? My understanding was we were planning to have a new collection for data coming back from the back office from the appeals we collect from frontend and we would use that for lpa-dashboard appeals. If this is for another purpose obviously ignore this
> > 
> > 
> > Yes I think I misunderstood the requirements rather. Where do I get the data from the interim?
> 
> So data will eventually come in from service bus topics and end up in a collection called appealsCaseData. We're going to fake that data for now by just manually adding data to collection.
> 
> I have added an initial example Paul gave us into a new collection in the db seed data, feel free to add more to that array if you need specific data for different scenarios: (004-appealsCaseData.json). We may need to manually add to the test db as well to allow for testing.
> 
> Ed has added an initial GET:many route to the api which has gone into main. Although I have suggested we rename some of it to avoid the references to lpa-dashboard-appeals
> 
> Could either wait for that rename work Or add the GET:id action to existing code, but that will require a reasonable merge/changes at some point from someone

Does this mean we need a new repository object in the appeals-service-api?
I added a route for this already in this ticket. What is this route that Ed added? Is it a duplication?

packages/appeals-service-api/__tests__/developer/fixtures/mongo-db-data.js
bryn-pins, line 1: Presume this is auto generated somehow but it looks a little like real user data, realise it's just for the abstract repo tests and it needs something, could we either use a different dataset that doesn't have user data or anonymize it somehow (thinking at the least emails should be @example.com, and from a brief look I don't think the addresses look real?). Whatever is easiest

packages/appeals-service-api/jest.config.js
bryn-pins, line 1: Can we undo the changes here, only needed to run the integration tests with jest runner locally

bryn-pins (commented):
couple of comments

packages/appeals-service-api/jest.config.js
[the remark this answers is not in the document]
bryn-pins: the PR changed halfway through review and this comment has been duplicated

bryn-pins (dismissed):

bryn-pins (approved):

PR #2963, opened 2023-07-22
feat(aapd-70): retrieve appeals case data from db based on lpa code

edpackard-work:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/__tests__/developer/lpa-dashboard-appeals.test.js
+++ b/packages/appeals-service-api/__tests__/developer/lpa-dashboard-appeals.test.js
@@ -1,0 +1,106 @@
+const http = require('http');
+const supertest = require('supertest');
+const { MongoClient } = require('mongodb');
+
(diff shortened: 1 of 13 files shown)


bryn-pins (approved):
Happy to approve for now to get things moving but thinking this should be potentially stick to the appealsCaseData/appeals-case-data naming rather than lpa-dashboard-appeals? Thinking the api could be used in future for something unrelated to lpa-dashboard so would be better to have a generic name

thoughts?

edpackard-work:
> Happy to approve for now to get things moving but thinking this should be potentially stick to the appealsCaseData/appeals-case-data naming rather than lpa-dashboard-appeals? Thinking the api could be used in future for something unrelated to lpa-dashboard so would be better to have a generic name
> 
> thoughts?

happy to do this - let's get it merged, and I'll fix as part of main task. It won't impact your work in the meantime :)

PR #2973, opened 2023-07-27
Bug/aapd 82 unable to see added users

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-82

## Description of change

Added id param to routes on manage-appeals/your-email-address route

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/common/your-email-address.js
+++ b/packages/forms-web-app/src/controllers/common/your-email-address.js
@@ -1,3 +1,5 @@
+const { getUserByEmail } = require('../../lib/appeals-api-wrapper');
+
 const getYourEmailAddress = (views) => {
 	return (req, res) => {
 		const { email } = req.session;
@@ -9,10 +11,7 @@ const getYourEmailAddress = (views) => {
 
 const postYourEmailAddress = (views) => {
 	return async (req, res) => {
-		const {
-			body,
-			params: { id }
-		} = req;
+		const { body } = req;
 		const { errors = {}, errorSummary = [] } = body;
 		if (!body['email-address'] || body['email-address'] === '') {
 			const customErrorSummary = [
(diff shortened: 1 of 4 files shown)


suityou01:
> I don't understand this I'm afraid Why would we need to find out the user's email via this login page if we already have the user's id in the url?
> 
> Surely the your email page should not use an id in the url and then on post it should look the user up by email and redirect them onto the enter code with their id in that url?

You and me both. If you look at the controller it expects an id from the url path.

packages/forms-web-app/src/controllers/common/your-email-address.js
```diff
@@ -32,6 +29,8 @@ const postYourEmailAddress = (views) => {
 		}
 
 		const { email } = body['email-address'];
+		const id = await getUserByEmail(email);
```
bryn-pins, line 32: Believe this returns the actual user so will need to use the _id prop from this method call

Realise there's no jsdocs on the api wrapper and no openapi specs
Was looking into adding some automated openapi specs and will try and find time to achieve that soon

bryn-pins (changes requested):

bryn-pins (approved):

bryn-pins (dismissed):

chrishawldev:
We don't actually need the id in the url, I think this is just from the Heroku template so feel free to just remove the id from the controller action if we don't need it. I'd rather we did that anyway so Google analytics isn't scraping all of our user id's

bryn-pins (approved):

packages/forms-web-app/src/controllers/common/your-email-address.js
```diff
@@ -30,6 +28,9 @@ const postYourEmailAddress = (views) => {
 		}
 
 		const { email } = body['email-address'];
+		const user = await getUserByEmail(email);
+		const id = user._id;
```
chrishawldev, line 32: Nice one

bryn-pins:
Closing as fixed in https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/2980

PR #2978, opened 2023-07-31
feat(appeals-service-api): documentMetadata repository

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-130

## Description of change

Document metadata repository

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/document-metadata-repository.js
+++ b/packages/appeals-service-api/src/repositories/document-metadata-repository.js
@@ -1,0 +1,37 @@
+const { MongoRepository } = require('./mongo-repository');
+
+class DocumentMetadataRepository extends MongoRepository {
+	constructor() {
+		super('documentMetadata');
+	}
+	/**
+	 * @typedef {'Supporting Documents' | 'Appeal Statement' | 'Planning application form' | 'Decision notice'} DocumentType
+	 */
+
+	/**
+	 *
+	 * @param {string} caseRef
+	 * @param {DocumentType} documentType
+	 * @return {Promise<any>}
+	 */
+	async getDocumentMetadata(caseRef, documentType) {
+		try {
+			return await this.findOneByQuery(
+				{
+					caseRef: caseRef,
+					documentType: documentType,
+					virusCheckStatus: 'scanned',
+					redactedStatus: 'redacted',
+					publishedStatus: 'published'
+				},
+				{
+					version: -1
+				}
+			);
+		} catch (e) {
+			console.log(e);
+		}
+	}
+}
+
+module.exports = { DocumentMetadataRepository };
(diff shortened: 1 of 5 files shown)


packages/appeals-service-api/src/repositories/document-metadata-repository.js
```diff
@@ -0,0 +1,37 @@
+const { MongoRepository } = require('./mongo-repository');
+
+class DocumentMetadataRepository extends MongoRepository {
```
chrishawldev, line 3: Nice

packages/appeals-service-api/src/repositories/document-metadata-repository.js
```diff
@@ -0,0 +1,37 @@
...
+			return await this.findOneByQuery(
+				{
+					caseRef: caseRef,
+					documentType: documentType,
+					virusCheckStatus: 'scanned',
+					redactedStatus: 'redacted',
+					publishedStatus: 'published'
+				},
+				{
+					version: -1
```
chrishawldev, line 28: Nice elegant query, only think I would say is it might make unit testing a bit easier if the different statuses were constants, but we're not planning on using them anywhere else so I'm not too fussed if we keep them as inline strings.

packages/appeals-service-api/src/routes/documentMetadata.js
```diff
@@ -0,0 +1,21 @@
+const express = require('express');
+const { getDocumentMetadata } = require('../services/document-metadata.service');
+const router = express.Router();
+
+router.get('/:caseref', async (req, res) => {
```
chrishawldev, line 5: The only question mark here is that the current examples we have of appeals case ref values all include protected url characters that we might need som kind of parsing function for to share in different places. @bryn-pins Did we end up needing to do this with any of the other appeals list views on the links to the appeal details views, because I think we mentioned we might need it there as well?

suityou01: I used URL encoded values for my testing and it worked fine as is. The front end will need to urlencode the caseRef.

packages/appeals-service-api/src/repositories/document-metadata-repository.js
suityou01: Unit testing? Where? 😷

chrishawldev: I guess when we write the client for these functions there might be some mock document metadata with statuses that will be the same as these?

chrishawldev (approved):

packages/appeals-service-api/src/repositories/document-metadata-repository.js
suityou01: It's more than likely. Is there anything to do for this ticket?

chrishawldev: No all good 👍

PR #2999, opened 2023-08-10
fix(aapd-177): appeals details page handles null document metadata

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aapd-177

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/lpa-dashboard/appeal-details.js
+++ b/packages/forms-web-app/src/controllers/lpa-dashboard/appeal-details.js
@@ -1,3 +1,4 @@
+const logger = require('../../lib/logger');
 const {
 	getAppealByLPACodeAndId,
 	getAppealDocumentMetaData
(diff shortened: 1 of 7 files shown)


edpackard-work (approved):

packages/forms-web-app/src/controllers/lpa-dashboard/appeal-details.js
```diff
@@ -14,50 +15,86 @@ const { getLPAUserFromSession } = require('../../services/lpa-user.service');
...
-		},
-		supportingDocuments
+		applicationForm: null,
+		decisionLetter: null,
+		appealStatement: null,
+		supportingDocuments: null
 	};
 
+	try {
+		const applicationFormMetaData = await getAppealDocumentMetaData(
```
chrishawldev, line 57: My only suggestion would be, is there a way to make this use an inline function to wrap all the try catch?
Something like
const documents = {applicationForm: null,
    decisionLetter: null,
    appealStatement:null,
    supportingDocuments:null}

async function getMetadata(caseReference, uri){
    try {
            let docMetaData = await getAppealDocumentMetaData(
                caseReference,
                encodeURI(uri)
            );
            let view = getDocumentViewData(docMetaData);
            return view;
        } catch (err) {
        
            logger.error(err);
        }
    }

// then do something like this to set all the values one by one
documents.applicationForm = await getMetadata(caseReference, 'Planning application form')

PR #3001, opened 2023-08-11
Feature/aapd 111

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aapd-111

## Description of change

Moved feature flag strings into common constants file
Hid questionnaire links without feature flag turned on
Allowed for lpa user's lpa to be used as a control variable for feature flags

Used default back link for appeal details page as it wasn't working locally, presume this is ok?

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/middleware/feature-flag.js
+++ b/packages/forms-web-app/src/middleware/feature-flag.js
@@ -1,4 +1,5 @@
 const { isFeatureActive } = require('../featureFlag');
+const { getLPAUserFromSession } = require('../services/lpa-user.service');
 
 const featureFlagMiddleware = (featureFlag, lpaCode) => {
 	return async (req, res, next) => {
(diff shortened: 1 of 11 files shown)


packages/forms-web-app/src/views/manage-appeals/appeal-details.njk
```diff
@@ -22,10 +22,6 @@
 </style>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      {{ govukBackLink({
-        text: "Back to appeals",
```
edpackard-work, line 26: should be ok to remove this but if using default backlink functionality, it will take user back to last page they visited which might not always be the appeals page. Because the text of the backlink in this case is specific, I think that's why it was hardcoded.

bryn-pins: ah ok, I'll see if I can get that to work somehow through the template as the current solution appears too low down as it's within the main.
If I can't figure anything out, I'll fix the link and leave positioning as is for now

packages/forms-web-app/src/middleware/feature-flag.js
```diff
@@ -10,4 +11,18 @@ const featureFlagMiddleware = (featureFlag, lpaCode) => {
...
 };
 
-module.exports = featureFlagMiddleware;
+const lpaUserFeatureFlagMiddleware = (featureFlag) => {
+	return async (req, res, next) => {
+		const user = getLPAUserFromSession(req);
+		if (await isFeatureActive(featureFlag, user?.lpaCode)) {
+			next();
+		} else {
+			return res.status(404).render('error/not-found');
```
chrishawldev, line 20: Nice

chrishawldev (approved):

PR #3009, opened 2023-08-16
feat(AAPD-148): base class for all validators

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-148

## Description of change

Abstract base class for all validators

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/validator/baseValidator.test.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/baseValidator.test.js
@@ -1,0 +1,14 @@
+const BaseValidator = require('./baseValidator');
+
+describe('src/dynamic-forms/validator/baseValidator.js', () => {
+	it('should not be possible to instantiate the base class', () => {
+		expect(() => new BaseValidator()).toThrow("Abstract classes can't be instantiated.");
+	});
+	it('should be possible to inherit from the base class', () => {
+		class MyValidator extends BaseValidator {}
+
+		const myValidator = new MyValidator();
+		expect(myValidator.constructor).toEqual(MyValidator);
+		expect(myValidator instanceof BaseValidator).toEqual(true);
+	});
+});


packages/forms-web-app/src/dynamic-forms/validator/baseValidator.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @abstract
+ * @class BaseValidator
+ */
+class BaseValidator {
+	constructor() {
+		if (this.constructor == BaseValidator) {
+			throw new Error("Abstract classes can't be instantiated.");
```
chrishawldev, line 8: Can we get a unit test on this please, looks like a nice enhancement good shout

chrishawldev (commented):
Do you think it's worth implementing the type utility method on this class
class BaseValidator {
	// The type of the vlaidator
	type;

}
Or do you think it's just as easy to do it inline / via utility method in the validate middleware here https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/spike/dynamic-questionnaire-renderer-validation-examples/packages/questionnaire-web-app/src/validator/validator.js#L14 when we tie it all together?

chrishawldev (commented):
Only other thing is this discussion point https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/3009#pullrequestreview-1580600723, we don't necessarily need to change anything but need to capture our decision point and we can add it to one of the other validation tickets if we're looking to stick it on the middleware, either way works for me so I'll leave it your discretion

suityou01:
> Do you think it's worth implementing the type utility method on this class class BaseValidator { // The type of the vlaidator type;
> 
> } Or do you think it's just as easy to do it inline / via utility method in the validate middleware here https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/spike/dynamic-questionnaire-renderer-validation-examples/packages/questionnaire-web-app/src/validator/validator.js#L14 when we tie it all together?

I don't think it's needed? Is the "instanceof" keyword not sufficient?

chrishawldev:
> > Do you think it's worth implementing the type utility method on this class class BaseValidator { // The type of the vlaidator type;
> > } Or do you think it's just as easy to do it inline / via utility method in the validate middleware here https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/spike/dynamic-questionnaire-renderer-validation-examples/packages/questionnaire-web-app/src/validator/validator.js#L14 when we tie it all together?
> 
> I don't think it's needed? Is the "instanceof" keyword not sufficient?

Cool that works for me. I'll add a comment to the Jira ticket. Nice one

chrishawldev (approved):

PR #3011, opened 2023-08-17
feat(AAPD-113): boolean validator

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-113

## Description of change

Boolean Validator

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/validator/booleanValidator.js
```diff
@@ -0,0 +1,10 @@
+const { body } = require('express-validator');
+const BaseValidator = require('./baseValidator.js');
+
+class BooleaValidator extends BaseValidator {
```
edpackard-work, line 4: typo

PaulGruffyddAmaze, line 4: Looks like there's a typo here - we're missing the `n` on `Boolean`.

packages/forms-web-app/src/dynamic-forms/validator/validOptionValidator.js
```diff
@@ -0,0 +1,14 @@
+const { body } = require('express-validator');
+const BaseValidator = require('./baseValidator.js');
+
+class ValidOptionValidator extends BaseValidator {
+	validate(questionObj, errorMessage) {
+		return body(questionObj.fieldName)
+			.custom((value) => {
+				if (!questionObj.options.includes(value)) throw new Error();
```
PaulGruffyddAmaze, line 8: As each option in the options array is an object with the `value` property being the one we're interested in, I think, rather than `includes()` we'd have to use `some()` like this:
`if (!questionObj.options.some(option => option.value == value)) throw new Error();`

packages/forms-web-app/src/dynamic-forms/validator/validOptionValidator.test.js
```diff
@@ -0,0 +1,29 @@
...
+	it('should return an error message if the field value is not in the list of options defined on the question', async () => {
+		const req = {
+			body: {
+				favouriteFruit: 'bananas'
+			}
+		};
+		const errorMessage = 'Please select only from the supplied options';
+		const question = {
+			fieldName: 'favouriteFruit',
+			options: ['apples', 'pears', 'oranges']
```
PaulGruffyddAmaze, line 21: As mentioned above, the options array contains objects so this would need to be something like:
`options: [{text: 'Apples', value: 'apples'}, {text: 'Pears', value: 'pears'}, {text: 'Oranges', value: 'oranges'}]`

packages/forms-web-app/src/dynamic-forms/validator/requiredValidator.test.js
```diff
@@ -0,0 +1,11 @@
+const RequiredValidator = require('./requiredValidator');
+describe('./src/dynamic-forms/validator/booleanValidator.test.js', () => {
+	it('should return an error message if the boolean field is missing', () => {
```
PaulGruffyddAmaze, line 3: It's probably worth covering off the opposite scenario too to ensure we don't get a validation error when the value is valid.

packages/forms-web-app/src/dynamic-forms/validator/booleanValidator.js
suityou01: This has been renamed anyway

suityou01: This has been renamed anyway

packages/forms-web-app/src/dynamic-forms/validator/validOptionValidator.js
suityou01: I have assumed option.value and value are of the same type, and have used strict equality, i.e.'===

packages/forms-web-app/src/dynamic-forms/validator/validOptionValidator.test.js
```diff
@@ -0,0 +1,29 @@
...
+describe('./src/dynamic-forms/validator/booleanValidator.test.js', () => {
+	it('should build a ValidOptionValidator', () => {
+		const vov = new ValidOptionValidator();
+		const message = 'Please select an answer';
+		const rule = vov.validate({}, message).builder.build();
+		expect(rule.locations[0]).toEqual('body');
+		expect(rule.stack[0].message).toEqual(message);
+		expect(rule.stack[0].validator.name).toEqual('');
+	});
+	it('should return an error message if the field value is not in the list of options defined on the question', async () => {
```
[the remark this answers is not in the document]
suityou01: An extremely helpful suggestion that has changed my approach to how I implemetned the validator. Now rather than raising an error I return true or false.

chrishawldev (changes requested):
We don't need a base validator class because it's a middleware that handles all validation of different types based on a switch statement, we need to do this so we can control the execution of the validation chain, the order it happens in and how we exit it. See an example of this works here https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/main/packages/forms-web-app/src/dynamic-forms/validator/validator.js.

chrishawldev (commented):
Actually scrap my previous comment, I think this will be neater. We cab still have the main validation middleware that has dependencies on this.

PR #3013, opened 2023-08-17
Feature/aapd 182 base question class

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-183

## Description of change

Question - base question class

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -1,69 +1,21 @@
...
+		if (!question || question === '') throw new Error('question parameter is mandatory');
+		if (!fieldName || fieldName === '') throw new Error('fieldName parameter is mandatory');
 		this.title = title;
 		this.question = question;
 		this.description = description;
 		this.type = type;
 		this.fieldName = fieldName;
 	}
-
-	prepQuestionForRendering(answers) {
```
PaulGruffyddAmaze, line 28: The `prepQuestionForRendering` method is used in the controller so that shouldn't be removed. Arguably the part relating to options could be moved to an override in the OptionsQuestion class as it would only be needed on a question with options.

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -1,69 +1,21 @@
...
-// refactor
-// jsdoc
-
 class Question {
 	title;
 	question;
 	description;
 	type;
 	fieldName;
+	validators = [];
```
PaulGruffyddAmaze, line 7: These fields cover what we need for question 1 but longer term we'll also need the following:
`hint` (string)
`bulletPoints` (array of strings)
`details` (object with a `title` and `text` field)

packages/forms-web-app/src/dynamic-forms/question.js
suityou01: Can you update the ticket?

packages/forms-web-app/src/dynamic-forms/question.js
suityou01: Can you update the ticket?

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -5,19 +5,24 @@
...
+	hint;
+	bulletPoints = [];
+	details = {
+		title: '',
```
chrishawldev, line 18: Just sense checking my understanding  - if we evaluated if(details.title) with an empty string like so it would return false so we can use that to do conditional rendering of the nun jucks templates on the server https://developer.mozilla.org/en-US/docs/Glossary/Truthy?

PR #3014, opened 2023-08-17
feat(aapd-185): base journey class

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aadd-185

## Description of change

Journey class
- Tested and added jsdocs
- Removed url hardcoding
- Questionnaire url in const in journey implementation
- Refactored to avoid nesting

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/controller.js
+++ b/packages/forms-web-app/src/dynamic-forms/controller.js
@@ -1,14 +1,15 @@
 // common controllers for dynamic forms
 const { getAppealByLPACodeAndId } = require('../lib/appeals-api-wrapper');
 const { getLPAUserFromSession } = require('../services/lpa-user.service');
-const { HasJourney } = require('./has-questionnaire/journey');
+const {
+	getJourney,
+	getJourneyResponseByType,
+	saveResponseToSessionByType
+} = require('./journey-types');
 
 // todo:
 // test
-// refactor?
-// remove hard coded tie ins to has questionnaire: pass in journey type via route, get links from journey implementation?
 // cleaner means of handling question type in urls
-// jsdoc
 
 const {
 	VIEW: {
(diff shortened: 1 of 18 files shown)


packages/forms-web-app/src/dynamic-forms/has-questionnaire/journey.js
```diff
@@ -2,9 +2,11 @@ const { questions } = require('./questions');
 const { Journey } = require('../journey');
 const { Section } = require('../section');
 
+const baseHASUrl = '/manage-appeals/questionnaire';
+
 class HasJourney extends Journey {
-	constructor(response) {
-		super(response);
+	constructor(journeyId, response) {
```
PaulGruffyddAmaze, line 8: The reference ID should already be included in the `response` object so shouldn't need to be included separately. The response object should look like the following:
```
{
    referenceId: "123456",
    formId: "has-questionnaire",
    startDate: "2023-08-01T14:32:19Z",
    updateDate: "2023-08-02T16:09:47Z",
    answers: {
        "green-belt": "no",
        "impacts-listed-building": "yes",
        "listed-buildings": [
            {
                name: "Stone bridge to the rear of Greendale Farm",
                grade: "II",
                listingId: "1234567"
            }
        ]
    }
}
```

bryn-pins: I moved it into the constructor instead of each function as we will always know it upfront
Could move to response but I guess that means we will always need a blank response when we create a journey
I'll rename to referenceId to be consistent

packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -17,6 +18,15 @@ const {
 } = require('./dynamic-components/views');
 //todo: should this be tied to a particular view, or can this be obtained from Journey object?
```
edpackard-work, line 19: just wanted to flag this up - any thoughts on making the view dynamic as part of the Journey class?

bryn-pins: I've created that as subtask for Lauren to pick up, so yes but was thinking of getting this in separately

packages/forms-web-app/src/dynamic-forms/journey-response.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Defines a response to a journey, a set of Answers to the questions
+ * @class
+ */
+class JourneyResponse {
+	/**
+	 * @type {string} - the unique reference id used in the url for the journey
+	 */
+	referenceId;
```
PaulGruffyddAmaze, line 9: I think you'll also need a `formId` as the `referenceId` on its own won't necessarily be unique (as we may have multiple forms per appeal e.g. LPA questionnaire, final comments, LPA statements, etc.).

bryn-pins: Are you thinking just an arbitrary string e.g. has-questionnaire?

bryn-pins: have done this

PR #3017, opened 2023-08-21
Feature/aapd 120 validation as an array

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-120

## Description of change

Validation of an array of validators

## Checklist

- [x] New features have tests
- [] Breaking change (team conversation required)

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/validator/validator.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/validator.js
@@ -20,14 +20,18 @@ const validate = () => {
 			}
 		});
 
-		const result = await Promise.all(validations.map((validation) => validation.run(req)));
-		const errors = result.reduce((a, b) => a.concat(b.errors), []);
-		if (errors.length === 0) {
-			next();
-			return true;
-		} else {
-			return result;
+		const results = [];
+		let overallResult = true;
+		for (const validation of validations) {
+			let validationResult = await validation.run(req);
+			results.push(validationResult);
+			if (validationResult.errors.length > 0) {
+				next();
+				overallResult = false;
+				break;
+			}
 		}
+		return overallResult;
 	};
 };
 
--- a/packages/forms-web-app/src/dynamic-forms/validator/validator.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/validator.js
@@ -1,49 +1,28 @@
-// const { body, validationResult } = require('express-validator');
-// const hasQuestionnaire = require('../definitions/questionnaire/hasQuestionnaire');
-// const questionnaire = hasQuestionnaire.questionnaire;
-// const formHelper = require('../lib/dynamicformhelper');
+const { body } = require('express-validator');
+const hasQuestionnaire = require('../definitions/questionnaire/hasQuestionnaire');
+const questionnaire = hasQuestionnaire.questionnaire;
+const formHelper = require('../lib/dynamicformhelper');
 
-// const validate = () => {
-//     return async (req, res, next) => {
-//         const { appealId, section, question } = req.params;
-//         var questionObj = formHelper.getQuestionBySectionAndName(questionnaire, section, question);
-//         const validations = [];
-//         switch (questionObj.validator?.type) {
-//             case "text":
-//                 validations.push(textValidations(questionObj))
-//                 break;
-//             case "boolean":
-//                 validations.push(booleanValidations(questionObj))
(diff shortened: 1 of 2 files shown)


packages/forms-web-app/src/dynamic-forms/validator/validator.js
```diff
@@ -1,49 +1,34 @@
...
+					break;
+				default:
+					throw new Error(`Validator of type ${validator.constructor} not implemented`);
+			}
+		});
 
-//        next();
-//     };
-// };
+		const result = await Promise.all(validations.map((validation) => validation.run(req)));
```
chrishawldev, line 23: We'll need to do this in some kind of loop instead https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/spike/dynamic-questionnaire-renderer-validation-examples/packages/questionnaire-web-app/src/validator/validator.js#L12 so we can check if there are any errors after each validation and exit out the validation middleware once any errors have been raised, just so we can support controlling the order of validations and avoid running any validation rules that don't need to be evaluated because an earlier validation has failed.

packages/forms-web-app/src/dynamic-forms/validator/validator.js
```diff
@@ -1,49 +1,38 @@
...
+		for (const validation of validations) {
+			let validationResult = await validation.run(req);
+			results.push(validationResult);
+			if (validationResult.errors.length > 0) {
+				next();
+				overallResult = false;
+				break;
+			}
+		}
+		return overallResult;
```
chrishawldev, line 34: We don't want to return a result here, we need to call next() to go to the next middleware.

packages/forms-web-app/src/dynamic-forms/validator/validator.js
```diff
@@ -1,49 +1,38 @@
...
-//     };
-// };
+		const results = [];
+		let overallResult = true;
+		for (const validation of validations) {
+			let validationResult = await validation.run(req);
+			results.push(validationResult);
+			if (validationResult.errors.length > 0) {
+				next();
+				overallResult = false;
```
chrishawldev, line 30: Not sure these will get called as we're calling next() on the middleware pipeline

chrishawldev (changes requested):
See comments

packages/forms-web-app/src/dynamic-forms/validator/validator.js
suityou01: I thought we only call the next middleware if the validation fails?

packages/forms-web-app/src/dynamic-forms/validator/validator.js
```diff
@@ -1,49 +1,38 @@
...
+			}
+		});
 
-//        next();
-//     };
-// };
+		const results = [];
+		let overallResult = true;
+		for (const validation of validations) {
+			let validationResult = await validation.run(req);
```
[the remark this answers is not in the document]
suityou01: I am assuming this is what the next middleware will do.

packages/forms-web-app/src/dynamic-forms/validator/validator.js
chrishawldev: It's not a javascript thing but rather part of the express middleware, because we're registering the middleware on the route we're ceding control to the express middleware pipeline so in this context calling next() will tell the express middleware to stop executing and hand over to the next middleware - e.g we've raise an error, stop processing and call the next middleware

packages/forms-web-app/src/dynamic-forms/validator/validator.js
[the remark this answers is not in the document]
chrishawldev: Possibly something like this will achieve what we want, reading up on some examples we should be able to check the presence of any errors how you have it there 👍 
let validationResult = await validation.run(req);
if (validationResult.errors.length > 0) {

				next();
}

packages/forms-web-app/src/dynamic-forms/validator/validator.js
chrishawldev: Ah I see because we're not explicitly calling return next() it will execute the rest of the stack. Do we want to just call return next() then to stop execution and move onto the error handler middleware as soon as an error is raised then?

suityou01:
> See comments

Seen

chrishawldev (approved):

PR #3018, opened 2023-08-21
feat(db-indexes): add indexes for appealCaseData and documentMetadata…

bryn-pins:
… tables

## Description of change

Adding indexes for lookups/sorts used on appealsCaseData and documentMetadata tables

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/db/setup.js
+++ b/packages/appeals-service-api/src/db/setup.js
@@ -17,8 +17,15 @@ const mongodb = require('./db');
 // 	try {
 // 		const saveAndReturnCollection = mongodb.get().collection('appeal');
 
-// 		await saveAndReturnCollection.createIndex({ 'appeal.horizonId': 1 }, { unique: true });
 // 		await saveAndReturnCollection.createIndex({ 'appeal.id': 1 }, { unique: true }); // todo: replace calls with _id and this index won't be needed?
+
+// 		await saveAndReturnCollection.createIndex({ 'appeal.horizonId': 1 }, { unique: true });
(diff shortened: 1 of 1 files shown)


packages/appeals-service-api/src/db/setup.js
```diff
@@ -50,10 +57,46 @@ async function setupUserIndexes() {
 	}
 }
 
+async function setupAppealCaseDataIndexes() {
+	try {
+		const caseDataCollection = mongodb.get().collection('appealsCaseData');
+
+		await caseDataCollection.createIndex({ caseReference: 1 }, { unique: true });
```
chrishawldev, line 64: @bryn-pins @PaulGruffyddAmaze Apparently I started a review on this, but I can't actually find my comment. I just wanted to query the uniqueness of caseReference. I know we had some questions over whether this would be unique and in one place we are looking at creating a synthetic partition key using another property concatenated with the caseReference. In this case we are sharding on the LPA code, so as long as the caseReference is guaranteed to be unique for a particular LPA then we're good without any further changes

bryn-pins: without the business knowledge to really know for sure, thinking it may be possible for either multiple people to submit an appeal for the same application so might not be guaranteed, maybe best to remove unique constraint if we're not sure?

PaulGruffyddAmaze: As far as I'm aware, the case reference should be unique within the `appealsCaseData` collection. The reason it isn't unique in the responses collection is that there could be more than one type of form submitted against a given appeal. In this instance though, we're dealing with the cases themselves so that value should be unique.

chrishawldev (approved):

PR #3019, opened 2023-08-23
Feature/aapd 234 post appeals case data

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-234

## Description of change

Post endpoint in appeals-service-api for case data

## Checklist

- [x] New features have tests

## Important


packages/appeals-service-api/src/controllers/appeals-case-data.js
```diff
@@ -33,7 +34,15 @@ const getAppealByCaseRefAndLpaCode = async (req, res) => {
 	}
 };
 
+const postAppealCase = async (req, res) => {
```
bryn-pins, line 37: Think this function should have error handling and the service to throw predefined errors if we are to follow the pattern from elsewhere

bryn-pins (changes requested):

packages/appeals-service-api/src/repositories/appeals-case-data-repository.js
```diff
@@ -59,6 +59,14 @@ class AppealsCaseDataRepository extends MongoRepository {
 
 		return result;
 	}
+	async postAppealCaseData(caseData) {
+		return await this.upsertManyById([
```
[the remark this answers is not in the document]
suityou01: Because I don't know the mongo api that well and this is the only method I could find with the word "upsert" in. Does updateOne do upsert in the background?

packages/appeals-service-api/src/controllers/appeals-case-data.js
suityou01: image

Linter gives this error in the service 

```
Unnecessary try/catch wrapper.eslint[no-useless-catch](https://eslint.org/docs/rules/no-useless-catch)
```

suityou01: Can you think of any specific errors I should be trapping? The pattern seems to be used rather sporadically and I can't find a consistent example of this use case (adding a JSON object)

packages/appeals-service-api/src/repositories/appeals-case-data-repository.js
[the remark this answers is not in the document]
bryn-pins: It looks to be a option you can pass in, see updateOne in mongoRepo

packages/appeals-service-api/src/controllers/appeals-case-data.js
chrishawldev: @bryn-pins @suityou01 For now we have a requirement on the ticket to return a 400 if the payload does not contain an LPACode property as we are using this for sharding. Apart from that there are no other specific error handling scenarios so we shouldn't catch any exceptions, just let them bubble up and return as the default 500 response.

chrishawldev (changes requested):

packages/appeals-service-api/src/controllers/appeals-case-data.js
bryn-pins: So the only reason I mentioned it is because I believe the api handler in forms-web-app throws a json parsing error if the api doesn't respond in a specific format, which then isn't obvious in the frontend logs what that error was. i.e. a 500 status code with no json body will error frontend side

But happy to leave as is for now until we know if we need a friendly error for user

chrishawldev: Ah good shout if we're not handling this at a higher level. @bryn-pins @suityou01 We could make a middleware that we can re use on routes that require a payload of json content type instead of handling it verbosely on the controller than have the middleware return 400 if the payload type is invalid? Or do we already do something like that @bryn-pins ?

packages/appeals-service-api/src/repositories/appeals-case-data-repository.js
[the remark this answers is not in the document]
bryn-pins: image

packages/appeals-service-api/src/controllers/appeals-case-data.js
bryn-pins: If you mean api side to automatically send a json body in the event of an error that could work, although part of me just wants to redo the frontend (handler() in appeals-api-wrapper.js)

Either way would expect that in a different ticket than this one as it would impact all endpoints and I'm not sure I'd trust our test suite to catch any issue that came out of it which is why I've avoided it so far.

packages/appeals-service-api/src/repositories/appeals-case-data-repository.js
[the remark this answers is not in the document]
suityou01: Where in the heck is this?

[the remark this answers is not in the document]
bryn-pins: Line 83:
https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/main/packages/appeals-service-api/src/repositories/mongo-repository.js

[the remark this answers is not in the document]
bryn-pins: Unless I've lost my mind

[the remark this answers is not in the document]
suityou01: This is what I have at line 83

```
* }
```

packages/appeals-service-api/src/controllers/appeals-case-data.js
bryn-pins: Am going to resolve this thread for now

bryn-pins (approved):
My comments have been resolved, so happy from my perspective

chrishawldev (approved):

PR #3023, opened 2023-08-23
feat(aapd-217): add validation pipeline to dynamic forms

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aapd-217

## Description of change

- moved error message to validator constructor
- refactored validator.js to avoid redundant loop
- stopped valid-option validator from enforcing required for a question

## Checklist

- [x] New features have tests

## Important


edpackard-work (dismissed):

edpackard-work (dismissed):

packages/forms-web-app/src/dynamic-forms/validator/valid-option-validator.js
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	/**
+	 * validates the response body, checking the value sent for the questionObj's fieldname is within the predefined list of options
+	 * @param {OptionsQuestion} questionObj
+	 */
+	validate(questionObj) {
+		return body(questionObj.fieldName)
+			.custom((value) => {
+				return questionObj.options.some((option) => option.value === value);
```
PaulGruffyddAmaze, line 34: I realise this isn't really something introduced in this PR but, at present, I think this validator might also be acting as a required field validator. If a field is required, we should enforce that with a required field validator so I think this will need to be modified to allow empty values through like this:
`return !value || questionObj.options.some((option) => option.value === value);`

PaulGruffyddAmaze (approved):
That looks good to me.

PR #3026, opened 2023-08-24
feat(aapd-216): add return to list link to question page

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aapd-216

## Description of change

- added to questionnaire link
- used viewFolder property instead of fieldName for clarity

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/radio/index.njk
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/radio/index.njk
@@ -1,4 +1,4 @@
-{% extends "layouts/no-banner-link/main.njk" %}
+{% extends layoutTemplate %}
 
 {% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
 {% from "govuk/components/radios/macro.njk" import govukRadios %}
--- a/packages/forms-web-app/src/dynamic-forms/controller.js
+++ b/packages/forms-web-app/src/dynamic-forms/controller.js
@@ -9,15 +9,15 @@ const {
 } = require('./journey-factory');
 
 // todo:
-// test
-// cleaner means of handling question type in urls
+// test save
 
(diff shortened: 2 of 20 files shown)


packages/forms-web-app/src/dynamic-forms/dynamic-components/radio/index.njk
```diff
@@ -41,6 +41,10 @@
           attributes: { "data-cy":"button-save-and-continue"}
         }) }}
       </form>
+
+      <p class="govuk-body">
+        <a href="{{listLink}}" class="govuk-link">Return to questionnaire homepage</a>
```
edpackard-work, line 46: maybe for future, but do we want the displayed text to be fixed, or something that can be passed into the template?

edpackard-work (approved):

packages/forms-web-app/src/dynamic-forms/dynamic-components/radio/index.njk
PaulGruffyddAmaze, line 46: Should the return link not be on the layout file rather than a specific individual component?

bryn-pins: I was worried that not every question would have it but probably makes more sense to put in template for now and figure out what to do if that comes up

bryn-pins: I'll move the whole thing to the template

PaulGruffyddAmaze: It looks like the link still needs to be removed from the radio component.

bryn-pins: apologies

PaulGruffyddAmaze (approved):

PR #3028, opened 2023-08-25
feat(aapd-119): add controller functionality and api changes for pers…

edpackard-work:
…isting user answers

## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -238,23 +242,44 @@ exports.save = async (req, res, journeyId) => {
...
+			if (propName.startsWith(questionObj.fieldName + '_')) {
+				journeyResponse.answers[propName] = req.body[propName];
+			}
 		}
-	}
 
-	// save response to session
-	saveResponseToSessionByType(req, journeyResponse);
+		// save answer to database
+		const answerToSave = { [questionObj.fieldName]: req.body[questionObj.fieldName] };
```
PaulGruffyddAmaze, line 261: Looking at this in conjunction with the section above, there may be more than one answer to save. In the section above, any field prefixed with the field name is also persisted in the version of the response stored in the session, but the patch sent to the DB only updates the field whose name exactly matches the specified fieldname. We'll need those other fields persisted to allow us to handle things like address collection and checkboxes with a conditional field (e.g. radio buttons with an "Other - please specify" option which shows a text field).

PaulGruffyddAmaze (approved):
Looks like you've addressed my comment so looks good to me.

PR #3029, opened 2023-08-25
feat(aapd-220): set titles on question pages

bryn-pins:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -198,6 +200,7 @@ exports.question = async (req, res, journeyId) => {
 		listLink: journey.baseUrl,
 		navigation: ['', backLink],
 		layoutTemplate: hasJourneyTemplate,
+		pageCaption: sectionObj?.name,
```
PaulGruffyddAmaze, line 203: It looks like the component views already write out a caption if a value of `caption` rather than `pageCaption` is passed in so, for consistency, we should either just use `caption` and add something to the summary list view to show a caption, or we should use `pageCaption` and remove the captions from the individual views.

PaulGruffyddAmaze (approved):
That's addressed my comments so looks good to me.

PR #3030, opened 2023-08-25
feat(AAPD-213): adding get response route

chrisHowarthkc:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/controllers/responses.js
+++ b/packages/appeals-service-api/src/controllers/responses.js
@@ -1,11 +1,11 @@
 const logger = require('../lib/logger');
-const { patchResponse } = require('../services/responses.service');
+const { patchResponse, getResponse } = require('../services/responses.service');
 
 const patchResponseByReferenceId = async (req, res) => {
 	let statusCode = 200;
 	let body = {};
 	try {
-		body = await patchResponse(req.params.formId, req.params.referenceId, req.body.answers);
+		body = await patchResponse(req.params.journeyId, req.params.referenceId, req.body.answers);
 	} catch (error) {
 		logger.error(`Failed to patch response: ${error.code} // ${error.message.errors}`);
 		statusCode = error.code;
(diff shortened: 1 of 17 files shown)


packages/appeals-service-api/src/controllers/responses.js
```diff
@@ -15,4 +15,18 @@ const patchResponseByReferenceId = async (req, res) => {
...
 };
 
-module.exports = { patchResponseByReferenceId };
+const getResponseByReferenceId = async (req, res) => {
+	let statusCode = 200;
+	let body = {};
+	try {
+		body = await getResponse(req.params.formId, req.params.referenceId, req.params?.projection);
+	} catch (error) {
+		logger.error(`Failed to patch response: ${error.code} // ${error.message.errors}`);
```
PaulGruffyddAmaze, line 24: Shouldn't this say `Failed to get response` rather than `Failed to patch response`?

packages/appeals-service-api/src/repositories/responses-repository.js
```diff
@@ -14,6 +14,25 @@ class ResponsesRepository extends MongoRepository {
...
+		if (projection) {
+			let formattedProjection = 'answers.' + projection;
+			responseProjection = {
+				[formattedProjection]: 1
+			};
+		}
+
+		const filter = { uniqueId: id };
+
+		result = await this.findOneByQuery(filter, sort, responseProjection);
```
PaulGruffyddAmaze, line 32: If I'm reading this correctly, it looks like it's receiving a single answer from the response but I don't know where we'd need to do that. I'd have thought we'd always want to retrieve the full response to ensure what we have in the session is in sync with the server.

chrisHowarthkc: As long as no projection is set we are currently getting the whole response object, however this can be limited to single answers when you pass through a projection, this was on the ticket as a requirement but if that's not correct happy to rip out the projection logic!

PaulGruffyddAmaze: Ah, ok. As long as we have a means of retrieving the full response, that's fine with me. @chrishawldev - I suspect in the scenario you mention, we'd want to retrieve the full response rather than just an individual answer as we need the full response to work out the next and previous steps in the journey. There may well be a need to retrieve an individual answer at some point though so I think it would be fine to keep that option available.

chrishawldev: I'd just been thinking about this and you're right we will want to retrieve the full response all the time when filling out questionnaires anyway. Possibly there might be a use case for the single answer projection in some other scenario's though, so if we support it there's no need to get rid, might just need to refactor to return the entire response by default at the moment if thats not the case already.

chrishawldev (dismissed):

packages/forms-web-app/src/dynamic-forms/middleware/getJourney.js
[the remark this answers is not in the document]
chrisHowarthkc: updated!

bryn-pins (approved):

PR #3044, opened 2023-09-01
feat(aapd-253): dynamic form file upload

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aapd-253
https://pins-ds.atlassian.net/browse/aapd-255

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/common/src/document-types.js
+++ b/packages/common/src/document-types.js
@@ -221,6 +221,14 @@ const documentTypes = {
 		involvement: '',
 		horizonDocumentType: 'Appellant Final Comments',
 		horizonDocumentGroupType: 'Evidence'
+	},
+	dynamic: {
+		name: 'dynamic',
+		multiple: true,
+		displayName: 'dynamic',
+		involvement: '',
+		horizonDocumentType: '',
+		horizonDocumentGroupType: ''
 	}
 };
 
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.js
@@ -1,15 +1,195 @@
+const logger = require('../../../lib/logger');
+const { patchQuestionResponse } = require('../../../lib/appeals-api-wrapper');
+// const { removeFiles } = require('../../../lib/multi-file-upload-helpers');
+const { createDocument } = require('../../../lib/documents-api-wrapper');
+const { mapMultiFileDocumentToSavedDocument } = require('../../../mappers/document-mapper');
+const { documentTypes } = require('@pins/common');
+
 const Question = require('../../question');
 
+/**
+ * @typedef {import('../../journey').Journey} Journey
+ * @typedef {import('../../journey-response').JourneyResponse} JourneyResponse
+ * @typedef {import('../../section').Section} Section
+ */
+
+/**
+ * A multi file upload question
+ * @class
+ */
 class MultiFileUploadQuestion extends Question {
-	constructor({ title, question, description, fieldName, validators } = {}) {
+	/**
+	 * @type {Object} document type
+	 */
+	documentType;
+
+	constructor({ title, question, description, fieldName, url, documentType, validators }) {
 		super({
 			title,
 			question,
 			description,
 			viewFolder: 'multi-file-upload',
 			fieldName,
+			url,
 			validators
 		});
+
+		if (documentType) {
+			this.documentType = documentType;
+		} else {
+			// default doc type
+			this.documentType = documentTypes.dynamic;
+		}
+	}
+
+	/**
+	 * Save an uploaded file
+	 * @param {ExpressRequest} req
+	 * @param {ExpressResponse} res
+	 * @param {Journey} journey
+	 * @param {Section} sectionObj
+	 * @param {JourneyResponse} journeyResponse
+	 * @returns
+	 */
+	saveAction = async (req, res, journey, sectionObj, journeyResponse) => {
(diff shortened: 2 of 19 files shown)


packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
```diff
@@ -79,7 +79,7 @@ exports.questions = {
 		title: 'Who did you notify?',
 		question: 'Who did you notify about this application?',
 		fieldName: 'notified-who',
-		validators: [new RequiredValidator('You must add your documents')]
```
bryn-pins, line 82: removed this for now until validators decided

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.js
```diff
@@ -10,6 +32,146 @@ class MultiFileUploadQuestion extends Question {
...
+			// 		href: '#'
+			// 	}));
+			// 	return this.#showError(res, viewModel);
+			// }
+
+			// map the response back to journeyResponse
+			journeyResponse.answers[this.fieldName] = responseToSave.answers[this.fieldName];
+
+			// update journey based on response
+			const updatedJourney = new journey.constructor(journeyResponse);
```
bryn-pins, line 120: wasn't sure on the best way to achieve this, alternative could be to have an update journey function

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.js
```diff
@@ -10,6 +32,146 @@ class MultiFileUploadQuestion extends Question {
...
+				);
+			}
+
+			return res.redirect(
+				updatedJourney.getNextQuestionUrl(sectionObj.segment, this.fieldName, false)
+			);
+		} catch (err) {
+			logger.error(err);
+
+			return this.#showError(res, journey, sectionObj, {
```
chrishawldev, line 136: I'm guessing this is for a scenario where validation ext passes but there is some other issue that prevents the files being saved?

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.js
```diff
@@ -10,6 +32,146 @@ class MultiFileUploadQuestion extends Question {
...
+				const mappedDoc = mapMultiFileDocumentToSavedDocument(document, document?.name, file.name);
+
+				result.push(mappedDoc);
+			}
+		}
+
+		return result;
+	}
+
+	#showError(res, journey, sectionObj, customViewData) {
```
chrishawldev, line 168: Would this work at a level above? As in should we implement showError on the base question? Perhaps we should be doing this in every scenario where we save data (e.g when cosmos fails to save an answer)?

chrishawldev: Ah I can see you're not handling this with the render page method and passing the error summary in there in the catch block, looks good 👍

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.js
bryn-pins: Potentially yes, have been unable to trigger it, but can leave in for now

chrishawldev: I think it's a good thing to get a test against. Could we trigger it in a unit test by mocking out one of the downstream dependencies to throw an error?

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.js
chrishawldev: I think thats a good thing to keep in mind if we need to do this anywhere else in the future, but if this behaviour is just a one off then this should be fine.

bryn-pins:
@chrishawldev I can't seem to comment on the unresolved thread.
That was for the upload-and-remain-on-page bit I left a comment on not knowing what it was for right?
Was there anything left to do on this?

packages/common/src/document-types.js
```diff
@@ -221,6 +221,14 @@ const documentTypes = {
 		involvement: '',
 		horizonDocumentType: 'Appellant Final Comments',
 		horizonDocumentGroupType: 'Evidence'
+	},
+	dynamic: {
```
chrishawldev, line 225: What's the use case for the dynamic document type?

chrishawldev (approved):

chrishawldev (approved):

packages/common/src/document-types.js
bryn-pins: So the document api requires meta data to be added alongside the file upload which comes from these objects.

It also runs integration style tests against each of the document types from this file to ensure they can be uploaded and fetched again.

So a large part of this is so I didn't need to rewrite too much of the existing document upload logic/tests/potentially break the existing appeals journey

I was under the impression we would want to continue using this, but would be adding new document types for each upload question, at which point we have a single source of every document type we have.

However I wasn't sure what the meta data needed for the various uploads was going to be so I created a generic one as a placeholder (the tickets for uploads don't specify any metadata needed at the moment)

So:
Maybe we don't need and we need to refactor the existing document upload/tests
Maybe we can just leave as this as is, as a catch all for docs that don't need this metadata
Maybe we add a new doc type for each file upload question

PR #3049, opened 2023-09-04
feat(AAPD-194): upload letters to neighbours

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-194

## Description of change

Upload letters to neighbours

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
+++ b/packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
@@ -97,6 +97,13 @@ exports.questions = {
 		fieldName: 'letters-to-neighbours',
 		validators: [new RequiredValidator()]
 	}),
+	uploadLettersToNeighbours: new MultiFileUploadQuestion({
+		title: 'Uploaded letters',
+		question: 'Upload the letters and emails',
+		fieldName: 'upload-letters-emails',
+		html: 'resources/upload-letters-emails/content.html',
+		validators: [new RequiredFileUploadValidator(), new MultifileUploadValidator()]
+	}),
 	pressAdvert: new BooleanQuestion({
 		title: 'Press Advert',
 		question: 'Did you put an advert in the local press?',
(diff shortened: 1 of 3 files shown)


packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
```diff
@@ -95,6 +95,12 @@ exports.questions = {
 		fieldName: 'letters-to-neighbours',
 		validators: [new RequiredValidator()]
 	}),
+	uploadLettersToNeighbours: new MultiFileUploadQuestion({
+		title: 'Upload the letters and emails',
```
chrisHowarthkc, line 99: The title here should be 'Uploaded letters' according to the Prototype

suityou01: But not according to the JIRA ticket.

chrishawldev (dismissed):
Please make changes to reflect the prototype attached to the jira ticket and the feedback received from colleagues.

packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
suityou01: Agreed, but the prototype is secondary to the JIRA ticket as the JIRA ticket has the estimates, history, comment history and as such is the single source of truth and the project record of what was built. If there is a discrepancy between the ticket and the prototype then this should be resoved between the technical leadership team and the business. I will escalate this up to Helen for clarification.

packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
```diff
@@ -95,6 +95,12 @@ exports.questions = {
 		fieldName: 'letters-to-neighbours',
 		validators: [new RequiredValidator()]
 	}),
+	uploadLettersToNeighbours: new MultiFileUploadQuestion({
+		title: 'Uploaded letters',
+		question: 'Upload the letters and emails',
+		fieldName: 'upload-letters-emailss',
+		validators: [new RequiredValidator('You must add your documents')]
```
edpackard-work, line 102: for multifile validation, this line should now read:
`validators: [new RequiredFileUploadValidator(), new MultifileUploadValidator()]`

edpackard-work: (you won't need to specify the error message as it is the default for the RequiredFileUploadValidator)

packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
```diff
@@ -95,6 +95,12 @@ exports.questions = {
 		fieldName: 'letters-to-neighbours',
 		validators: [new RequiredValidator()]
 	}),
+	uploadLettersToNeighbours: new MultiFileUploadQuestion({
+		title: 'Uploaded letters',
+		question: 'Upload the letters and emails',
+		fieldName: 'upload-letters-emailss',
```
chrishawldev, line 101: We can now use the new html property from AAPD-301 in here. Main suggestion would be to use a naming convention like resources/upload-letters-emails/content.html for the location of the html

chrishawldev (approved):

PR #3061, opened 2023-09-08
feat(AAPD-301): optional html property added to multiform upload comp…

suityou01:
…onent

## Ticket Number

AAPD-301

Optional html property for base question class

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.test.js
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.test.js
@@ -21,6 +21,7 @@ const DESCRIPTION = 'File question';
 const FIELDNAME = 'files';
 const URL = 'url';
 const VALIDATORS = [1, 2];
+const HTML = 'resources/question12/content.html';
 const DOCUMENT_TYPE = {
 	name: '1'
 };
@@ -81,7 +82,8 @@ function getMultiFileUpload(
 	question = QUESTION,
 	description = DESCRIPTION,
 	url = URL,
-	validators = VALIDATORS
+	validators = VALIDATORS,
+	html= HTML
 ) {
 	return new MultiFileUploadQuestion({
 		title: title,
@@ -90,6 +92,7 @@ function getMultiFileUpload(
 		fieldName: fieldName,
 		url: url,
 		validators: validators,
+		html: html,
 		documentType: documentType
 	});
 }
@@ -114,7 +117,6 @@ describe('MultiFileUploadQuestion', () => {
 
 	it('should create', () => {
 		const multiFileQuestion = getMultiFileUpload();
-
 		expect(multiFileQuestion.title).toEqual(TITLE);
 		expect(multiFileQuestion.question).toEqual(QUESTION);
 		expect(multiFileQuestion.description).toEqual(DESCRIPTION);
@@ -123,6 +125,7 @@ describe('MultiFileUploadQuestion', () => {
 		expect(multiFileQuestion.validators).toEqual(VALIDATORS);
 		expect(multiFileQuestion.viewFolder).toEqual('multi-file-upload');
 		expect(multiFileQuestion.documentType).toEqual(DOCUMENT_TYPE);
+		expect(multiFileQuestion.html).toEqual(HTML);
 	});
 
 	it('should create a MultiFileUploadQuestion with default documentType', () => {
--- a/packages/forms-web-app/src/dynamic-forms/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/question.js
@@ -87,7 +87,12 @@ class Question {
 	 */
 	url;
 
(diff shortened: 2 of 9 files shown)


packages/forms-web-app/src/dynamic-forms/has-questionnaire/template.njk
```diff
@@ -12,6 +12,10 @@
   {% endif %}
 {% endblock before_content %}
 
+{% block optional_html %}
+  <div>Optional HTML</div>
```
chrishawldev, line 16: Don't think we need this block

packages/forms-web-app/src/views/layouts/multi-file-upload.njk
```diff
@@ -61,7 +61,8 @@
       {% if body %}
          <p class="govuk-body">{{body}}</p>
       {% endif %}
-      
+      {% block optional_html %}
+      {% endblock optional_html %}
```
chrishawldev, line 65: Don't think we need this block

packages/forms-web-app/src/dynamic-forms/has-questionnaire/template.njk
suityou01: removed

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -77,7 +77,21 @@ class Question {
...
+
+	constructor({
+		title,
+		question,
+		description,
+		viewFolder,
+		fieldName,
+		validators,
+		url,
+		html = '/blank.html'
```
chrishawldev, line 93: What about instead of a default value in the nun jucks template we did something like? 
`{% if showBackToListLink %}
{% block optional_html %}
    {% include question.html ignore missing %}
{% endblock optional_html %}
{% endif %}`

packages/forms-web-app/src/views/layouts/multi-file-upload.njk
suityou01: I think we do because if I remove it it breaks the unit tests

chrishawldev: I don't think we should be unit testing the nun jucks templates, we're using the nun jucks rendering engine which is outside of our control. If that doesn't render what we're expecting then it's either an issue with nun jucks or the values we're supplying, which is tested in other places.
The rendered UI will be tested as part of acceptance as well.

packages/forms-web-app/src/dynamic-forms/question.js
chrishawldev: Apologies, yes exactly that. Thats the example value from the nun jucks docs that I forgot to replace sorry.

packages/forms-web-app/src/views/layouts/multi-file-upload.njk
suityou01: I shall remove the unit tests.

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-file-upload/question.test.js
```diff
@@ -22,5 +63,40 @@ describe('./src/dynamic-forms/dynamic-components/boolean/question.js', () => {
 		expect(multiFileQuestion.fieldName).toEqual(FIELDNAME);
 		expect(multiFileQuestion.validators).toEqual(VALIDATORS);
 		expect(multiFileQuestion.viewFolder).toEqual('multi-file-upload');
+		expect(multiFileQuestion.html).toEqual(HTML);
+	});
+	it('should render optional html correctly', () => {
+		const VIEWFOLDER = 'multi-file-upload';
+		const TEMPLATEPATH = `dynamic-components/${VIEWFOLDER}/index.njk`;
+		document.body.innerHTML = nunjucksTestRenderer.render(TEMPLATEPATH, viewModel);
```
chrishawldev, line 71: As mentioned above, if we have to support unnecessary code in the templates then I think the cost out weighs the benefit. They are also potentially brittle in that if we upgrade the gov uk packages for the nun jucks templates it could break these tests as the html elements could easily change.

suityou01: I have now removed the unit test

chrishawldev (approved):
Nice work, LGTM

packages/forms-web-app/src/public/resources/question12/content.html
```diff
@@ -0,0 +1,10 @@
+<p class="govuk-body">We need a:</p>
```
edpackard-work, line 1: naming: this might not always be question 12 - maybe use the fieldname instead?

chrishawldev: I think this might just be in there as example for now, but fieldName is a good shout. We've got the actual user story here https://pins-ds.atlassian.net/browse/AAPD-194 so we can tweak naming conventions in there.

PR #3064, opened 2023-09-11
fix(aapd-195): change url properties for uploading press advert

edpackard-work:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/has-questionnaire/questions.js
```diff
@@ -107,7 +107,7 @@ exports.questions = {
 		title: 'Upload the press advertisement',
```
chrishawldev, line 107: Is this correct as it duplicates the text in the question field? Do we want to make use of the pageTitle property in the PR that @bryn-pins has open?

edpackard-work: I believe this ticket duplicates the question field for the title, it's aapd-178 where they differ

chrishawldev (approved):

PR #3076, opened 2023-09-14
feat(AAPD-319): string validator

suityou01:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-319

## Description of change

String validation

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
@@ -1,0 +1,51 @@
+/* eslint-disable no-unused-vars */
+const { body } = require('express-validator');
+const BaseValidator = require('./base-validator.js');
+
+class StringValidator extends BaseValidator {
+	regex = {
+		regex: '',
+		regexMessage: ''
+	};
+	minLength = {
+		minLength: 0,
+		minLengthMessage: ''
+	};
+	maxLength = {
+		maxLength: 0,
+		maxLengthMessage: ''
+	};
+	constructor({ regex, minLength, maxLength } = {}) {
+		super();
+
+		if (!regex && !minLength && !maxLength) throw new Error('Nothing to do!');
+		if (regex && regex.regex) this.regex = regex;
+		if (minLength && minLength.minLength) this.minLength = minLength;
+		if (maxLength && maxLength.maxLength) this.maxLength = maxLength;
+	}
+	validate(questionObj) {
+		const regexValidator = (chain) =>
+			chain.matches(new RegExp(this.regex.regex)).withMessage(this.regex.regexMessage);
+		const minLengthValidator = (chain) =>
+			chain
+				.isLength({ min: this.minLength.minLength })
+				.withMessage(this.minLength.minLengthMessage);
+		const maxLengthValidator = (chain) =>
+			chain
+				.isLength({ max: this.maxLength.maxLength })
+				.withMessage(this.maxLength.maxLengthMessage);
+		let chain = body(questionObj.fieldName);
+		if (this.regex.regex) {
+			chain = regexValidator(chain);
+		}
+		if (this.minLength.minLength) {
+			chain = minLengthValidator(chain);
+		}
+		if (this.maxLength.maxLength) {
+			chain = maxLengthValidator(chain);
+		}
+		return chain;
+	}
+}
+
+module.exports = StringValidator;
--- a/packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
@@ -22,6 +22,15 @@ class StringValidator extends BaseValidator {
 		if (regex && regex.regex) this.regex = regex;
 		if (minLength && minLength.minLength) this.minLength = minLength;
 		if (maxLength && maxLength.maxLength) this.maxLength = maxLength;
+		this.regex.regexMessage = this.regex.regexMessage
+			? this.regex.regexMessage
+			: 'Please enter only the allowed characters';
+		this.minLength.minLengthMessage = this.minLength.minLengthMessage
+			? this.minLength.minLengthMessage
+			: `Input too short - Please enter at least ${this.minLength.minLength} characters`;
+		this.maxLength.maxLengthMessage = this.maxLength.maxLengthMessage
+			? this.maxLength.maxLengthMessage
+			: `Input too long - Please enter no more than ${this.maxLength.maxLength} characters`;
 	}
 	validate(questionObj) {
 		const regexValidator = (chain) =>
--- a/packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-unused-vars */
 const { body } = require('express-validator');
 const BaseValidator = require('./base-validator.js');
 
@@ -18,7 +17,8 @@ class StringValidator extends BaseValidator {
 	constructor({ regex, minLength, maxLength } = {}) {
 		super();
 
-		if (!regex && !minLength && !maxLength) throw new Error('Nothing to do!');
+		if (!regex && !minLength && !maxLength)
+			throw new Error('String validator is invoked without any validations set!');
 		if (regex && regex.regex) this.regex = regex;
 		if (minLength && minLength.minLength) this.minLength = minLength;
 		if (maxLength && maxLength.maxLength) this.maxLength = maxLength;
--- a/packages/forms-web-app/src/dynamic-forms/validator/string-validator.test.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/string-validator.test.js
@@ -1,0 +1,218 @@
+const StringValidator = require('./string-validator');
+
+describe('src/dynamic-forms/validator/string-validator.js', () => {
+	it('should throw if nothing passed to constructor', () => {
+		// eslint-disable-next-line no-unused-vars
+		expect(() => {
+			new StringValidator();
+		}).toThrow('Nothing to do!');
+	});
+	it('should invalidate string that does not match regex', async () => {
+		const req = {
+			body: {
+				numberOfToes: 'three'
+			}
+		};
+		const options = { regex: { regex: '^\\d', regexMessage: 'Number of toes should be a number' } };
+		const question = {
+			fieldName: 'numberOfToes'
+		};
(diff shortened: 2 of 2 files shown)


packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
```diff
@@ -0,0 +1,51 @@
...
+		minLengthMessage: ''
+	};
+	maxLength = {
+		maxLength: 0,
+		maxLengthMessage: ''
+	};
+	constructor({ regex, minLength, maxLength } = {}) {
+		super();
+
+		if (!regex && !minLength && !maxLength) throw new Error('Nothing to do!');
```
chrishawldev, line 21: Not sure we want to throw an error as it's a valid scenario that we might not have any validation on a string

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
```diff
@@ -0,0 +1,51 @@
...
+		if (regex && regex.regex) this.regex = regex;
+		if (minLength && minLength.minLength) this.minLength = minLength;
+		if (maxLength && maxLength.maxLength) this.maxLength = maxLength;
+	}
+	validate(questionObj) {
+		const regexValidator = (chain) =>
+			chain.matches(new RegExp(this.regex.regex)).withMessage(this.regex.regexMessage);
+		const minLengthValidator = (chain) =>
```
chrishawldev, line 29: This would maybe be readable as a traditional express validator chain fluent declaration like
```
chain.matches(new RegExp(this.regex.regex)).withMessage(this.regex.regexMessage)
.isLength({ min: this.minLength.minLength })
.withMessage(this.minLength.minLengthMessage)
.isLength({ max: this.maxLength.maxLength })
.withMessage(this.maxLength.maxLengthMessage);

```
Theres a good example here https://github.com/jayeshchoudhary/express-validator-example/blob/main/validations/user.validation.js

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
```diff
@@ -0,0 +1,51 @@
...
+	regex = {
+		regex: '',
+		regexMessage: ''
+	};
+	minLength = {
+		minLength: 0,
+		minLengthMessage: ''
+	};
+	maxLength = {
+		maxLength: 0,
```
chrishawldev, line 15: maxLength should be a default maximum value like 10000 so if we don't pass in a value for max then it defaults to the largest allowable string value

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable no-unused-vars */
+const { body } = require('express-validator');
+const BaseValidator = require('./base-validator.js');
+
+class StringValidator extends BaseValidator {
+	regex = {
+		regex: '',
```
chrishawldev, line 7: Should this be a * so if we don't pass in a regex validation then we are assuming anything is good and that the regex validation is essentially optional?

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
```diff
@@ -0,0 +1,51 @@
...
+		regexMessage: ''
+	};
+	minLength = {
+		minLength: 0,
+		minLengthMessage: ''
```
chrishawldev, line 12: Lets ask about some default error messages at stand up, hopefully there will be some content we can use on one of the tickets

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: Then don't use the validator in the first place?

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: I think you missed the point of what I was doing here. The chain is built dynamically based on inputs.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: This was not a requirement. Rather a suggestion on how to implement. I have implemetned this a better way.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: No.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: Please update the JIRA ticket with any defaults

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
chrishawldev: We might want to only evaluate max length and the allowed characters and not worry about a min length for example, or maybe we only want to do a regex evaluation.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
chrishawldev: We want it to always be a complete chain with either default values or by leveraging the optional syntax of the express validator as suggested in the ticket, the code example above would achieve the same effect but with less code and more consistency with the express validator docs and existing string evaluation in the code base.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
chrishawldev: The changes are required please, we either need default values or to use the optional syntax. Whether something is better is subjective but this implementation diverges from the express validator and the existing code base, both of which are suitable implementations and will provide consistency.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
chrishawldev: A link to the appropriate unit test would be helpful in this case please.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
chrishawldev: Will do, I'll raise it now

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
chrishawldev: I'm more familiar with a wild card regex like the ones here https://howtodoinjava.com/java/regex/match-any-set-of-characters/#3-match-any-character-zero-or-more-occurrences. I think it would be clearer that we're matching anything, including special characters, than an empty string as the default?

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: Please read the unit tests.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: Please read the unit tests

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: You clearly do not understand what you are seeing here. Please read the unit tests.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
suityou01: Please read the unit tests

chrishawldev (changes requested):
Please make changes in line with feedback

packages/forms-web-app/src/dynamic-forms/validator/string-validator.test.js
```diff
@@ -0,0 +1,218 @@
...
+				maxLength: 4,
+				maxLengthMessage: 'Short statement should be a maxmimum of 4 characters in length'
+			}
+		};
+		const question = {
+			fieldName: 'shortStatement'
+		};
+		const validationResult = await new StringValidator(options).validate(question).run(req);
+		expect(validationResult.errors.length).toEqual(1);
+		expect(validationResult.errors[0].msg).toEqual(options.maxLength.mLengthMessage);
```
beejjacobs, line 216: Looks like a typo here: `mLengthMessage` which is why tests are failing.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
beejjacobs, line 21: This error message could be more descriptive, while a stack trace would indicate where the error is it's helpful to have useful error messages.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable no-unused-vars */
```
beejjacobs, line 1: Is this rule-disable required?

beejjacobs (changes requested):
This looks like a reasonable implementation to me. There is a minor error in the tests causing them to fail, see the comment below.

I suggest if defaults (values and messages), are required for convenience, these are defined in a ticket. These can either be added to the constructor, or a static method such as `withDefaults`.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
beejjacobs: Most of the operators for express validator seem to be geared around branching within a chain based on request data. Since this is configuration based, it's a slightly different use-case.

Either this or some custom chain checks should work fine.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
beejjacobs: Looks like we're using `''` as falsey (and therefore not including the regex validation in the chain), I generally tend to avoid this (with a preference to explicit false or null) but that's a style preference, it'll work OK.

packages/forms-web-app/src/dynamic-forms/validator/string-validator.js
beejjacobs: If defaults are required let's just get them defined with values and messages.

chrishawldev (approved):

beejjacobs (approved):

PR #3079, opened 2023-09-15
feat(aapd-323): add Address class

edpackard-work:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/common/src/lib/address.js
+++ b/packages/common/src/lib/address.js
@@ -1,0 +1,52 @@
+/**
+ * Defines the shape of an address object
+ * @class
+ */
+class Address {
+	/**
+	 * @type {string} - the first line of the address
+	 */
+	addressLine1;
+
+	/**
+	 * @type {string} - the second line of the address
+	 */
+	addressLine2;
+
+	/**
+	 * @type {string} - the name of the town, city or other settlement
+	 */
+	cityTown;
+
+	/**
+	 * @type {string} - the postcode
+	 */
+	postcode;
+
+	/**
+	 * creates an instance of an address
+	 * @param {Object} params
+	 * @param {string} params.addressLine1
+	 * @param {string} params.addressLine2
+	 * @param {string} params.cityTown
+	 * @param {string} params.postcode
+	 */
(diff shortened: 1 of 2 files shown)


packages/common/src/lib/address.js
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * @type {string} - the name of the town, city or other settlement
+	 */
+	town;
+
+	/**
+	 * @type {string} - the name of the county
+	 */
+	county;
```
chrishawldev, line 24: Did we have a separate county field? I thought we just had 4 fields. If so then all good.

packages/common/src/lib/address.js
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * @type {string} - the second line of the address
+	 */
+	addressLine2;
+
+	/**
+	 * @type {string} - the name of the town, city or other settlement
+	 */
+	town;
```
chrishawldev, line 19: Maybe call this CityTown as its value could be either

packages/common/src/lib/address.js
```diff
@@ -0,0 +1,49 @@
...
+	 * @param {string} params.addressLine2
+	 * @param {string} params.town
+	 * @param {string} params.county
+	 * @param {string} params.postcode
+	 */
+	constructor({ addressLine1, addressLine2, town, county, postcode }) {
+		this.addressLine1 = addressLine1;
+		this.addressLine2 = addressLine2;
+		this.town = town;
+		this.county = county;
```
chrishawldev, line 44: I think the other thing we talked about was that addressLine2 and postcode are optional so we could make them the last 2 constructor parameters and then also maybe thrown an error if addressLine1 and City / Town aren't provided (and maybe county if we need that?)

packages/common/src/lib/address.js
edpackard-work: oops, I probably got this from looking at addresses on the appellant journey - if we don't need county, I will eliminate it

chrisHowarthkc (approved):

chrishawldev (approved):
LGTM

PR #3083, opened 2023-09-15
Feature/aapd 306

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/aapd-306

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/options-question.js
+++ b/packages/forms-web-app/src/dynamic-forms/options-question.js
@@ -1,11 +1,29 @@
+//const nunjucks = require('nunjucks');
 const Question = require('./question');
 
 const ValidOptionValidator = require('./validator/valid-option-validator');
 
 /**
+ * @typedef {import('./question').QuestionViewModel} QuestionViewModel
+ * @typedef {import('./journey').Journey} Journey
+ * @typedef {import('./section').Section} Section
+ */
+
+/**
  * @typedef {Object} Option
  * @property {string} text - text shown to user
  * @property {string} value - value on form
+ * @property {boolean|undefined} [checked] - if the
+ * @property {Object|undefined} [conditional]
+ */
+
+/**
+ * @typedef {QuestionViewModel & { question: OptionsProperty }} OptionsViewModel
+ */
+
+/**
+ * @typedef {Object} OptionsProperty
+ * @property {Array.<Option> | Array} options - An array of options.
  */
 
 class OptionsQuestion extends Question {
@@ -51,7 +69,47 @@ class OptionsQuestion extends Question {
 			description,
 			validators: optionsValidators
 		});
+
 		this.options = options;
+	}
+
+	/**
+	 * gets the view model for this question
+	 * @param {Section} section - the current section
+	 * @param {Journey} journey - the journey we are in
+	 * @param {Object|undefined} [customViewData] additional data to send to view
+	 * @returns {OptionsViewModel}
+	 */
+	prepQuestionForRendering(section, journey, customViewData) {
+		const answer = journey.response.answers[this.fieldName] || '';
+
+		const viewModel = super.prepQuestionForRendering(section, journey, customViewData);
+
+		viewModel.question.options = [];
+
+		for (const option of this.options) {
+			let optionData = { ...option };
+			if (optionData.value !== undefined) {
+				optionData.checked = (',' + answer + ',').includes(',' + optionData.value + ',');
+			}
+
+			// also need to handle dependent fields & set their answers
+			// if (optionData.conditional !== undefined) {
+			// 	let conditionalField = { ...optionData.conditional };
+			// 	conditionalField.fieldName = this.fieldName + '_' + conditionalField.fieldName;
+			// 	conditionalField.value = journey.response.answers[conditionalField.fieldName] || '';
+			// 	optionData.conditional = {
+			// 		html: nunjucks.render(
+			// 			`../views/questions/conditional/${conditionalField.type}.njk`,
+			// 			conditionalField
+			// 		)
+			// 	};
+			// }
+
+			viewModel.question.options.push(optionData);
+		}
+
+		return viewModel;
 	}
 }
 
--- a/packages/forms-web-app/src/dynamic-forms/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/question.js
@@ -1,17 +1,31 @@
-const nunjucks = require('nunjucks');
+const { patchQuestionResponse } = require('../lib/appeals-api-wrapper');
 
 /**
  * @typedef {import('./validator/base-validator')} BaseValidator
+ * @typedef {import('./journey').Journey} Journey
+ * @typedef {import('./journey-response').JourneyResponse} JourneyResponse
+ * @typedef {import('./section').Section} Section
  */
 
 /**
- * @typedef {Object} ViewData
- * @property {string} layoutTemplate - path to template file
- * @property {string} pageCaption - caption displayed above question
- * @property {string} backLink - url to go back 1 page
- * @property {string} listLink - url for listing page
- * @property {Object} answers - all answers to journey so far
- * @property {Object} answer - answer to current question
+ * @typedef {Object} PreppedQuestion
+ * @property {Object} value
+ * @property {string} question
+ * @property {string} fieldName
+ * @property {string} pageTitle
+ * @property {string} [description]
+ * @property {string} [html]
+ */
+
+/**
+ * @typedef {Object} QuestionViewModel
+ * @property {PreppedQuestion} question
+ * @property {string} layoutTemplate
+ * @property {string} pageCaption
+ * @property {Array.<string>} navigation
+ * @property {string} backLink
+ * @property {boolean} showBackToListLink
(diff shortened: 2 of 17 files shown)


packages/forms-web-app/src/dynamic-forms/dynamic-components/add-more/index.njk
```diff
@@ -0,0 +1,61 @@
...
+      <h1>Hello</h1>
+
+      <ul>
+        {% for answer in addMoreAnswers %}
+            <li>{{answer.label}} :: {{answer.answer}} :: {{answer.removeLink}}</li>
+        {% endfor %}
+      </ul>
+
+      <form method="POST" novalidate>
+        <input type="hidden" name="add-more-question" value="yes"/>
```
bryn-pins, line 22: using a hidden input here to know if we are working on an add more question

packages/forms-web-app/src/dynamic-forms/validator/validator.js
```diff
@@ -11,12 +11,18 @@ const validate = () => {
...
 
-		const questionObj = journey.getQuestionBySectionAndName(section, question);
+		let questionObj = journey.getQuestionBySectionAndName(section, question);
 
 		if (!questionObj) {
 			throw new Error('unknown question type');
 		}
 
+		// add more question pages to validate subquestion if not currently adding a subquestion
+		const isAddMorePage = req.body['add-more-question'];
```
bryn-pins, line 21: wasn't happy with this but couldn't see a better way to know which validation to hit

packages/forms-web-app/src/dynamic-forms/options-question.js
```diff
@@ -51,8 +69,48 @@ class OptionsQuestion extends Question {
...
+
+		viewModel.question.options = [];
+
+		for (const option of this.options) {
+			let optionData = { ...option };
+			if (optionData.value !== undefined) {
+				optionData.checked = (',' + answer + ',').includes(',' + optionData.value + ',');
+			}
+
+			// also need to handle dependent fields & set their answers
```
bryn-pins, line 96: this isn't being used yet as far as I can tell

packages/forms-web-app/src/dynamic-forms/dynamic-components/add-more/question.js
```diff
@@ -0,0 +1,217 @@
...
+
+		// check for saving errors
+		const saveViewModel = this.subQuestion.checkForSavingErrors(req, section, journey);
+		if (saveViewModel) {
+			return this.renderAction(res, saveViewModel);
+		}
+
+		// move to the next question
+		const updatedJourney = new journey.constructor(journeyResponse);
+		return this.handleNextQuestion(res, updatedJourney, section.segment, this.fieldName);
```
bryn-pins, line 213: not sure it's obvious in this function why/when to use this or this.subQuestion

bryn-pins: it's also repeating a lot of the logic from question

packages/forms-web-app/src/dynamic-forms/dynamic-components/add-more/question.js
```diff
@@ -0,0 +1,217 @@
...
+		if (journeyResponse.answers[this.fieldName]) {
+			responseToSave.answers[this.fieldName] = [...journeyResponse.answers[this.fieldName]];
+		}
+
+		// get answer to subquestion
+		const individual = await this.subQuestion.getDataToSave(req, journeyResponse);
+
+		responseToSave.answers[this.fieldName].push({
+			...individual.answers,
+			addMoreId: uuid.v4()
```
bryn-pins, line 151: adding a uuid here to help identify which question for the remove link

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -102,71 +113,167 @@ class Question {
...
+	/**
+	 * Save the answer to the question
+	 * @param {ExpressRequest} req
+	 * @param {ExpressResponse} res
+	 * @param {Journey} journey
+	 * @param {Section} section
+	 * @param {JourneyResponse} journeyResponse
+	 * @returns {Promise<void>}
+	 */
+	async saveAction(req, res, journey, section, journeyResponse) {
```
bryn-pins, line 256: I've created a common saveAction
multi file uploads check for validation errors after saving, the others do not use checkForSavingErrors

An alternative I've been considering is potentially have an array of functions to pipe together

can then control exactly what order to call these functions and can override some if necessary, could help reduce the duplication + minor variations in the add more override of this

packages/forms-web-app/src/dynamic-forms/dynamic-components/add-more/question.js
```diff
@@ -0,0 +1,217 @@
...
+		const addMoreAnswer = req.body[this.fieldName];
+
+		if (isAddMorePage) {
+			const errorViewModel = this.checkForValidationErrors(req, section, journey);
+			if (errorViewModel) {
+				return this.renderAction(res, errorViewModel);
+			}
+
+			if (addMoreAnswer === 'yes') {
+				return this.renderAction(res, this.subQuestion.prepQuestionForRendering(section, journey));
```
bryn-pins, line 189: Do we need different urls for the subquestion?
If so how best to achieve that?

packages/forms-web-app/src/dynamic-forms/dynamic-components/add-more/index.njk
PaulGruffyddAmaze: As an alternative, you could give your `govukButton` below a name i.e.
```
{{ govukButton({
  name: "add-more-question",
  text: "Continue",
  type: "submit",
  attributes: { "data-cy":"button-save-and-continue"}
}) }}
```

packages/forms-web-app/src/dynamic-forms/dynamic-components/add-more/question.js
PaulGruffyddAmaze: I wouldn't have thought we'd need a separate URL as we don't have a scenario where we'd want the user to be able to directly access a sub-question.

chrishawldev (commented):
Discussed with @bryn-pins around the potential to split out the ListAddMore (responsibility for managing the list) and a separate AddMore class (responsibility for processing submissions to go on to the list) and we might look at doing the first AddMore question class implementation to flesh out this idea

packages/forms-web-app/src/dynamic-forms/middleware/dynamic-req-files-to-req-body-files.js
```diff
@@ -22,7 +23,10 @@ module.exports = () => (req, res, next) => {
 	const journeyResponse = res.locals.journeyResponse;
 
 	const journey = getJourney(journeyResponse);
-	const questionObj = journey.getQuestionBySectionAndName(section, question);
+	let questionObj = journey.getQuestionBySectionAndName(section, question);
+
+	questionObj = getSubquestionIfPresent(req, questionObj);
```
chrishawldev, line 28: I'm not sure we'd need this for multifile upload as it can already be used to manage a list of files within that component, unless we have an example on the existing work where we can have a question associated with multiple different file upload document types

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -51,6 +61,8 @@ class Question {
 	url;
 	/** @type {string|undefined} optional html content */
 	html;
+	/** @type {Question|undefined} */
+	subQuestion;
```
chrishawldev, line 65: I'm not sure we need to support this at this level, at least not yet. The only use case I'm aware of so far is on the ListAddMore type questions where we want to add items to a list? Would it make it simpler if we moved this down to the level where it is currently needed or does it not make that much difference in reality?

packages/forms-web-app/src/dynamic-forms/middleware/dynamic-req-files-to-req-body-files.js
bryn-pins: It's not working correctly with files anyway as it needs the ability to go back and edit the subquestion (due to the fact it saves then validates, so if one file fails it currently breaks), so yes happy to take this out, although not sure on the best way to indicate that a question type can't be used as a subquestion.

Was just getting it as close to working as possible with files as that was the only question with a custom saveAction, that might need to be handled alongside the addMore custom saveAction

packages/forms-web-app/src/dynamic-forms/question.js
bryn-pins: I need to know about it in validator.js
However I can probably check instance of is AddMore and then check subquestion
Not that it matters that much due to it being js, more a jsdoc thing

bryn-pins: So yes will move

chrishawldev: Good old jsdoc

chrishawldev (dismissed):
Left a few comments but I think we'll likely end up refactoring when working on the related tickets anyway

chrishawldev (approved):

PR #3090, opened 2023-09-20
feat(aapd-308): aligning task list values with up to date prototype

chrisHowarthkc:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/controller.js
+++ b/packages/forms-web-app/src/dynamic-forms/controller.js
@@ -1,4 +1,5 @@
 // common controllers for dynamic forms
+
 const { getAppealByLPACodeAndId } = require('../lib/appeals-api-wrapper');
 const { getLPAUserFromSession } = require('../services/lpa-user.service');
 const { SECTION_STATUS } = require('./section');
@@ -67,7 +68,7 @@ function buildSectionRowViewModel(key, value, action) {
 			text: key
 		},
 		value: {
-			text: value
+			html: value
 		},
 		actions: {
 			items: [action]
--- a/packages/forms-web-app/src/dynamic-forms/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/question.js
@@ -1,4 +1,5 @@
 const { patchQuestionResponse } = require('../lib/appeals-api-wrapper');
+const { capitalize } = require('../lib/string-functions');
 
 /**
  * @typedef {import('./validator/base-validator')} BaseValidator
@@ -276,7 +277,8 @@ class Question {
 	 * @returns {string}
 	 */
 	formatAnswerForSummary(answer) {
-		return this.altText ?? answer ?? 'Not started';
+		const formattedAnswer = capitalize(this.altText ?? answer ?? 'Not started');
+		return formattedAnswer;
 	}
 }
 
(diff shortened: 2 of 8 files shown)


packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -134,8 +135,18 @@ exports.list = async (req, res) => {
 
 			// default question format
 			const key = question.title ?? question.question;
-			const value =
+			const currentQuestion =
 				question.altText ?? journey.response?.answers[question.fieldName] ?? 'Not started';
+			let value = '';
+
+			if (currentQuestion.uploadedFiles) {
+				for (const item in currentQuestion.uploadedFiles) {
```
chrisHowarthkc, line 143: currently displaying the files as a plain text list, future ticket in the works to update this to downloadable links

packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -134,8 +135,18 @@ exports.list = async (req, res) => {
...
+					value = value + currentQuestion.uploadedFiles[item].originalFileName + '</br>';
+				}
+			} else {
+				value = capitalize(currentQuestion);
```
chrisHowarthkc, line 147: capitalization is to ensure yes/no values being pulled through are displayed with the correct case as pointed out by Steve

packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -134,8 +135,18 @@ exports.list = async (req, res) => {
 
 			// default question format
 			const key = question.title ?? question.question;
-			const value =
+			const currentQuestion =
 				question.altText ?? journey.response?.answers[question.fieldName] ?? 'Not started';
+			let value = '';
+
+			if (currentQuestion.uploadedFiles) {
```
chrishawldev, line 142: Main change I would recommend is looking to move this logic into the multifile upload question. I think on this branch we are doing this 
`const value = question.formatAnswerForSummary(answer); `
https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/9d5ff270b97bc7efb494b8429e9475d040d9bd5c/packages/forms-web-app/src/dynamic-forms/controller.js#L139, essentially it moves the responsibility for any task list formatting to the QuestionClasses so the controller can be ignorant of how that is handled and doesn't need to know anything about the implementation details of individual questions, like if they have any uploaded files or not, it just needs to know that for every question it will call 
`question.formatAnswerForSummary(answer);`
And either that will do the default behaviour here 
https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/9d5ff270b97bc7efb494b8429e9475d040d9bd5c/packages/forms-web-app/src/dynamic-forms/question.js#L276
Or if a question has an override it will call that. We might need to update the js doc spec on that base question class as it names the method as format rather than formatAnswerForSummary.

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -276,7 +277,17 @@ class Question {
 	 * @returns {string}
 	 */
 	formatAnswerForSummary(answer) {
-		return this.altText ?? answer ?? 'Not started';
+		let answerValue = this.altText ?? answer ?? 'Not started';
+		let formattedAnswer = '';
+		if (answer != undefined && answer.uploadedFiles) {
+			for (const item in answer.uploadedFiles) {
+				formattedAnswer = formattedAnswer + answer.uploadedFiles[item].originalFileName + '</br>';
```
chrishawldev, line 284: Ideally this would be in the multifile file upload question class rather than the base question.

chrishawldev (commented):
Just that one comment about moving that logic to the multifile upload question

PR #3098, opened 2023-09-25
fix(AAPD-347): eslint plugin  to warn about using custom types in cat…

suityou01:
…ch blocks without type guards

## Ticket Number

AAPD-347

https://pins-ds.atlassian.net/browse/AAPD-347

## Description of change

ESLint plugin to detect use of custom types in catch blocks without type guards

## Checklist

- [x] New features have tests
- [x] Breaking change (team conversation required)

## Important

### DIFF
--- a/custom-eslint-rules/catch-blocks/enforce-catch-block-error-type-guards.js
+++ b/custom-eslint-rules/catch-blocks/enforce-catch-block-error-type-guards.js
@@ -1,0 +1,102 @@
+module.exports = {
+	meta: {
+		type: 'suggestion',
+		version: '0.0.1',
+		name: 'eslint-plugin-enforce-catch-block-error-type-guards',
+		hasSuggestions: true,
+		fixable: 'code',
+		docs: {
+			description: 'Warn against the use of non type safe code in catch blocks',
+			category: 'Possible errors'
+		},
+		schema: []
+	},
+	create(context) {
+		const findParentNodeOfType = (node, parentType) => {
+			let parent = node;
+			while (parent) {
+				parent = parent.parent;
+				if (parent && parent.type === parentType) {
+					return parent;
+				}
+			}
+		};
+
+		const getName = (node) => {
+			if (node.name) return node.name;
+			if (node.object) return getName(node.object);
+			return null;
+		};
+
+		return {
+			ExpressionStatement(node) {
+				const catchClause = findParentNodeOfType(node, 'CatchClause');
+				if (!catchClause) return;
+
+				const value = node.expression.right;
+				if (!value) return;
+
+				const { param } = catchClause;
+
+				if (
+					value.type !== 'MemberExpression' ||
+					!value.object ||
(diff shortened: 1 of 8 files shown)


custom-eslint-rules/catch-blocks/enforce-catch-block-error-type-guards.js
```diff
@@ -0,0 +1,86 @@
...
+					) {
+						prop = value.property;
+						let e = new Error();
+						// eslint-disable-next-line no-prototype-builtins
+						if (!e.hasOwnProperty(prop.name)) {
+							let ifStatement = findParentNodeOfType(node, 'IfStatement');
+							if (!ifStatement) {
+								isProblem = true;
+							} else {
+								if (!ifStatement.test.left.name === param.name) isProblem = true;
```
beejjacobs, line 46: This appears to be where the error is, also it'd be clearer as `ifStatement.test.left.name !== param.name` if that is what is intended.

beejjacobs (changes requested):
Code looks good to me. There are a few bugs though:

If I run this in the root of the repo, I get an error:

```
>npx eslint .

Oops! Something went wrong! :(

ESLint: 8.50.0

TypeError: Cannot read properties of undefined (reading 'name')
Occurred while linting appeal-planning-decision/packages/horizon-functions/horizon-add-document/index.js:124    
Rule: "enforce-catch-block-error-type-guards/enforce-catch-block-error-type-guards"
    at ExpressionStatement (appeal-planning-decision/custom-eslint-rules/catch-blocks/enforce-catch-block-error-type-guards.js:46:36)
    ...
```

A temporary fix of that issue to make it work, and it does pick up the issue in the repo, for example at `src/controllers/appeals-case-data.js:15`

custom-eslint-rules/catch-blocks/enforce-catch-block-error-type-guards.test.js
beejjacobs, line 1: This errors for me when I run it:

```
appeal-planning-decision\node_modules\eslint\lib\rule-tester\rule-tester.js:458
        throw err;
        ^

Error: Use node.range[1] instead of node.end
Occurred while linting <input>:1
Rule: "enforce-catch-block-error-type-guards"
```

bryn-pins (changes requested):
There's an issue I've left a comment on that needs fixing.

Played around with this by taking the if check in/out from the listed building route and have 2 more general thoughts

- The fix was applied automatically when using npm run commit which to me feels a bit off due to extremity of the change, would prefer for it to be a suggestion that could be applied after checking it over
- the fix happened after formatting, leaving the indentation wrong

So if we keep to autofix then would prefer an auto formatter to apply as well

bryn-pins (approved):
I think this is ok now

bryn-pins (approved):

bryn-pins (approved):

beejjacobs (approved):

PR #3108, opened 2023-10-01
feat(AAPD-291): add neighbouring properties

chrishawldev:
Adds the ability to add neighbouring properties to a questionnaire response. Fixes a bug with list add more questions that causes a rendering failure when there is only on item in a list for a list add more question

## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-291

## Description of change

Adds the ability to add neighbouring properties to a questionnaire response. 
Fixes a bug with list add more questions that causes a rendering failure when there is only on item in a list for a list add more question. 
Custom format option on base question class

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/question.js
@@ -48,8 +48,6 @@ class Question {
 	fieldName;
 	/** @type {boolean} if the question should appear in the journey overview task list or not */
 	taskList;
-	/** @type {function|undefined} function that customises the formatting of the question in the task list */
-	format;
 	/** @type {Array.<BaseValidator>} array of validators that a question uses to validate answers */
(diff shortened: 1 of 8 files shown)


packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -306,6 +304,15 @@ class Question {
 		return action;
 	}
 
+	/**
+	 *
+	 * @param {Object.<Any>} answer
+	 * @returns The formatted address to be presented in the UI
```
edpackard-work, line 310: as this is the base class, should @returns description be something more generic? i.e `the formatted string to be presented in the UI`

edpackard-work (dismissed):

edpackard-work (approved):

PR #3109, opened 2023-10-02
feat(aapd-94): add listed building number question

edpackard-work:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-94

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/identifier/index.njk
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/identifier/index.njk
@@ -37,10 +37,18 @@
                         {% include question.html ignore missing %}
                     {% endif %}
 
-                    {{ govukLabel({
-                        text: question.question,
-                        for: question.fieldName
-                    }) }}
+                    {{ govukInput({
+                        label: {
+                            text: question.question
+                        },
+                        id: question.fieldName,
+                        name: question.fieldName,
+                        hint: { text: question.hint },
+                        errorMessage: errors[question.fieldName] and {
+                            text: errors[question.fieldName].msg
+                        },
+                    classes: question.inputClasses
+                }) }}
                 {% else %}
                     {{ govukLabel({
                         text: question.question,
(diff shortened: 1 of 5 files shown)


packages/forms-web-app/src/dynamic-forms/dynamic-components/identifier/index.njk
```diff
@@ -55,6 +55,9 @@
                 {% endif %}
 
                 {{ govukInput({
+                    label: {
```
bryn-pins, line 58: is this needed, a label is added in two ways above?

bryn-pins: allowing it to either be the page heading or to have them separate

edpackard-work: this is the label that is attached to the text input area: on the prototype, the text input area has a label, which is formatted distinctly (padding) from the other text - if we didn't want to have the option to set the text input area label, I could probably set the padding manually using the html property

edpackard-work: (or I could rename to inputBoxLabel or similar, for clarity)

bryn-pins: If you gave a question.pageHeading of 'Tell us the list entry number and question'
and a question.question of 'seven digit number'
It would display correctly, without needing a new label prop.

The current issue with this implementation is that you now have 2 labels for the same input and I believe the text for h1 (also first label) is also incorrect

I spoke to Steve about it when building the identifier class and he said:

GDS have a more compact version of forms where the H1 is also the label (or vice versa actually)
And in those pages there might be a hint that is linked 'ARIA described by'

But, if we add in more stuff, we need to go back to the older format where there is:

h1
paragraphs
form
  label
  hint
  input

bryn-pins: happy to go down the path of using a label prop _instead of_ pageHeading if that is clearer (question prop is then always h1 and optionally a label if none is provided) but it should handle both scenarios mentioned by Steve

bryn-pins (approved):

PR #3110, opened 2023-10-02
Feature/aapd 369

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-369

## Description of change

add listed building function
add listed building api
added joi package to api to validate schema of listedBuildings passed through
added error logging to default api error handler

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/mongo-repository.js
+++ b/packages/appeals-service-api/src/repositories/mongo-repository.js
@@ -117,6 +117,39 @@ class MongoRepository {
 
 	/**
 	 *
+	 * @param {string} filterProp the property on the colection to filter on
+	 * @param {any[]} updateOperations Mongo JSON structures that represent update operations. Their
+	 * structure should be:
+	 * {
+	 * 	filterProp: <the property to filter on with the same name as the filterProp param>,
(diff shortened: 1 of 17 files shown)


packages/appeals-service-api/src/repositories/mongo-repository.js
```diff
@@ -115,6 +115,39 @@ class MongoRepository {
...
+					},
+					update: {
+						$set: updateOneOperation.updateSet
+					},
+					upsert: true
+				}
+			};
+		});
+
+		return await mongodb.get().collection(this.collectionName).bulkWrite(updates);
```
chrishawldev, line 148: I'm not entirely sure that cosmos actually supports bulk write through the mongo api, if it does great, but the azure docs suggest this as an alternative https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/optimize-write-performance#set-ordered-to-false-in-the-mongodb-drivers

chrishawldev (approved):
Just one comment on the mongo method to use for batch insert, but looks good

packages/appeals-service-api/src/repositories/mongo-repository.js
bryn-pins: It's being used already in the existing upsertManyById function, so think it should be ok

PR #3112, opened 2023-10-03
feat(aapd-283): implement conditional rendering for BooleanTextQuestion

edpackard-work:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-283

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/dynamic-components/boolean-text/index.njk
```diff
@@ -8,18 +8,6 @@
 {% set title = question.pageTitle + " - " + journeyTitle + " - GOV.UK" %}
 {% block pageTitle %}{{ "Error: " + title if errors else title }}{% endblock %}
 
-{% if question.conditional %}
```
edpackard-work, line 11: we don't need this, because we have different conditional types (text, textarea, number etc) that are set during instantiation of the relevant question class

packages/forms-web-app/src/dynamic-forms/dynamic-components/boolean-text/question.js
```diff
@@ -34,6 +35,8 @@ class BooleanTextQuestion extends RadioQuestion {
 			options,
 			validators
 		});
+
+		this.html = html;
```
edpackard-work, line 39: arguably we want to inherit this from parent class(es)? the html field hasn't been universally implemented yet

packages/forms-web-app/src/dynamic-forms/dynamic-components/boolean-text/index.njk
```diff
@@ -51,26 +42,27 @@
           },
           fieldset: {
             legend: {
-              text: question.subQuestion,
+              text: question.label,
               isPageHeading: true,
               classes: "govuk-fieldset__legend--m"
             }
           },
           items: [
```
edpackard-work, line 50: iterate over these

packages/forms-web-app/src/dynamic-forms/options-question.js
```diff
@@ -93,18 +93,20 @@ class OptionsQuestion extends Question {
...
-			// 			`../views/questions/conditional/${conditionalField.type}.njk`,
-			// 			conditionalField
-			// 		)
-			// 	};
-			// }
+			// handle conditional (dependant) fields & set their answers
+			if (optionData.conditional !== undefined) {
+				let conditionalField = { ...optionData.conditional };
+
+				conditionalField.fieldName = this.fieldName + '_value';
```
PaulGruffyddAmaze, line 100: I don't know whether we have any examples of this, but it's possible to have conditional fields on more than one answer (e.g. https://design-system.service.gov.uk/components/checkboxes/#conditionally-revealing-a-related-question). By naming the conditional field `this.fieldName + '_value'` we'd potentially end up with multiple fields with the same name which would make it difficult to work out which answer goes with each selected option.

edpackard-work: should we use a conditional fieldname value (or use the existing value) as in the POC so you get something like:

`new-planning-conditions_yes-value`

(fieldName + conditional value + `-value`)

the -value string on the end is to help with identifying these conditional answers

packages/forms-web-app/src/dynamic-forms/dynamic-components/boolean-text/index.njk
```diff
@@ -6,19 +6,23 @@
...
+{% set options = [] %}
+
+{% for option in question.options %}
+  {% set options = (options.push({
+    value: option.value,
+    text: option.text,
+    attributes: { "data-cy": "answer-" + option.value },
+    checked: answer == option.value,      
+    conditional: {
+      html: option.conditional.html,
```
ms-ek, line 19: Interesting that this line doesn't throw an error (might just be Nunjucks behaviour 😄). `option.conditional.html` doesn't exist for option 'no'. Have you checked there are no errors/warnings here when the option has no conditional fields?

edpackard-work: if a value in nunjucks is undefined/null, nothing is rendered :)

packages/forms-web-app/src/dynamic-forms/options-question.js
```diff
@@ -93,18 +93,20 @@ class OptionsQuestion extends Question {
...
-			// 			`../views/questions/conditional/${conditionalField.type}.njk`,
-			// 			conditionalField
-			// 		)
-			// 	};
-			// }
+			// handle conditional (dependant) fields & set their answers
+			if (optionData.conditional !== undefined) {
+				let conditionalField = { ...optionData.conditional };
+
+				conditionalField.fieldName = this.fieldName + '_' + option.value + '-value';
```
ms-ek, line 100: Optional but if you want you could write it this way:
``${conditionalField.fieldName}_${option.value}-value`;`

edpackard-work: I've gone with the approach above to make it slightly clearer (imo) how the fieldname is constructed

PaulGruffyddAmaze, line 100: I think using the value as-is in the field name might cause some issues. The value of a form field could be any string which could include characters such as apostrophes, commas and quotation marks which would be invalid in a field name. To use the value, you'd need to sanitise it so it's a valid kebab-case string.

packages/forms-web-app/src/dynamic-forms/dynamic-components/radio/index.njk
```diff
@@ -5,6 +5,22 @@
...
+    value: option.value,
+    text: option.text,
+    attributes: { "data-cy": "answer-" + option.value },
+    checked: answer == option.value,      
+    conditional: {
+      html: option.conditional.html,
+      value: option.conditional.fieldName
+    }
+  }), options) %}
+{% endfor %}
```
PaulGruffyddAmaze, line 22: I'm wondering whether we need this logic in the template to define the options array. As far as I can see, `question.options` is almost identical, with items containing all of those fields except for `attributes` which could be added in the `prepQuestionForRendering` method of the options question as we do for the `checked` attribute.

PaulGruffyddAmaze (approved):
It looks like my comments have been addressed so looks good to me.

PR #3114, opened 2023-10-05
Feature/aapd 279 conditional rendering safety risks

ms-ek:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-279

## Description of change

Conditional rendering with validation added

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/options-question.js
+++ b/packages/forms-web-app/src/dynamic-forms/options-question.js
@@ -2,6 +2,7 @@ const nunjucks = require('nunjucks');
 const Question = require('./question');
 
 const ValidOptionValidator = require('./validator/valid-option-validator');
+const { getConditionalFieldName } = require('./dynamic-components/utils/question-utils');
 
 /**
  * @typedef {import('./question').QuestionViewModel} QuestionViewModel
@@ -80,10 +81,12 @@ class OptionsQuestion extends Question {
 	 * @param {Object|undefined} [customViewData] additional data to send to view
 	 * @returns {OptionsViewModel}
 	 */
-	prepQuestionForRendering(section, journey, customViewData) {
-		const answer = journey.response.answers[this.fieldName] || '';
+	prepQuestionForRendering(section, journey, customViewData, payload) {
+		const answer = payload
+			? payload[this.fieldName]
+			: journey.response.answers[this.fieldName] || '';
 
-		const viewModel = super.prepQuestionForRendering(section, journey, customViewData);
+		const viewModel = super.prepQuestionForRendering(section, journey, customViewData, payload);
 
 		viewModel.question.options = [];
 
@@ -100,14 +103,20 @@ class OptionsQuestion extends Question {
 			if (optionData.conditional !== undefined) {
 				let conditionalField = { ...optionData.conditional };
 
-				conditionalField.fieldName = this.fieldName + '_' + conditionalField.fieldName;
-				conditionalField.value = journey.response.answers[conditionalField.fieldName] || '';
+				conditionalField.fieldName = getConditionalFieldName(
+					this.fieldName,
+					conditionalField.fieldName
+				);
+				conditionalField.value = payload
+					? payload[conditionalField.fieldName]
+					: journey.response.answers[conditionalField.fieldName] || '';
 
 				optionData.conditional = {
-					html: nunjucks.render(
-						`./dynamic-components/conditional/${conditionalField.type}.njk`,
-						conditionalField
-					)
+					html: nunjucks.render(`./dynamic-components/conditional/${conditionalField.type}.njk`, {
+						payload,
+						...conditionalField,
+						...customViewData
+					})
 				};
 			}
 
--- a/packages/forms-web-app/src/dynamic-forms/validator/conditional-required-validator.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/conditional-required-validator.js
@@ -1,0 +1,53 @@
+const { body } = require('express-validator');
+const BaseValidator = require('./base-validator.js');
+
+/**
+ * @typedef {import('../question.js')} Question
+ */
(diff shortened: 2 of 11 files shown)


packages/forms-web-app/src/dynamic-forms/options-question.js
```diff
@@ -101,10 +101,10 @@ class OptionsQuestion extends Question {
 				conditionalField.value = journey.response.answers[conditionalField.fieldName] || '';
 
 				optionData.conditional = {
-					html: nunjucks.render(
-						`./dynamic-components/conditional/${conditionalField.type}.njk`,
-						conditionalField
-					)
+					html: nunjucks.render(`./dynamic-components/conditional/${conditionalField.type}.njk`, {
+						...conditionalField,
+						...customViewData
```
edpackard-work, line 106: what does the customviewdata provide here - is it the errors?

packages/forms-web-app/src/dynamic-forms/validator/conditional-required-validator.js
```diff
@@ -0,0 +1,53 @@
...
+						.if(body(questionObj.fieldName).equals(option.value))
+						.notEmpty()
+						.withMessage(this.errorMessage)
+				);
+			}
+			return schema;
+		}, []);
+	}
+
+	getConditionalFieldName(questionObj, option) {
```
edpackard-work, line 48: we could arguably have this method somewhere common, passing in fieldname and conditionalfieldname and it returning the correctly formatted string. This could then be called both by the validator and the options question class.

packages/forms-web-app/src/dynamic-forms/validator/conditional-required-validator.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * @typedef {import('../question.js')} Question
+ */
+
+/**
+ * enforces a field is not empty when condition is satisfied
+ * @class
+ */
+class ConditionalRequiredValidator extends BaseValidator {
```
edpackard-work, line 12: looks good, can you put in some unit tests? see other validators for examples - need to test constructor and validate method

packages/forms-web-app/src/dynamic-forms/options-question.js
edpackard-work: (I think this also causing the existing unit tests to fail)

ms-ek: yep it's providing the errors. will need to fix failing tests

packages/forms-web-app/src/dynamic-forms/validator/conditional-required-validator.js
```diff
@@ -0,0 +1,53 @@
+const { body } = require('express-validator');
```
chrishawldev, line 1: Another option for doing this might be to extend the existing stringValidator constructorsso we can override the field name to be validated to point at the conditional field instead, so instead of defaulting to use the questionObj.fieldName we instead use the passed in value. I think the below should work.

`const { body } = require('express-validator');
const BaseValidator = require('./base-validator.js');

class StringValidator extends BaseValidator {
	minLength = {
		minLength: 0,
		minLengthMessage: ''
	};
	maxLength = {
		maxLength: 0,
		maxLengthMessage: ''
	};
	regex = {
		regex: '',
		regexMessage: ''
	};
	constructor({ minLength, maxLength, regex, fieldName } = {}) {
		super();

		if (!minLength && !maxLength && !regex)
			throw new Error('String validator is invoked without any validations set!');
		if (minLength && minLength.minLength) this.minLength = minLength;
		if (maxLength && maxLength.maxLength) this.maxLength = maxLength;
		if (regex && regex.regex) this.regex = regex;
		this.minLength.minLengthMessage = this.minLength.minLengthMessage
			? this.minLength.minLengthMessage
			: `Input too short - Please enter at least ${this.minLength.minLength} characters`;
		this.maxLength.maxLengthMessage = this.maxLength.maxLengthMessage
			? this.maxLength.maxLengthMessage
			: `Input too long - Please enter no more than ${this.maxLength.maxLength} characters`;
		this.regex.regexMessage = this.regex.regexMessage
			? this.regex.regexMessage
			: 'Please enter only the allowed characters';
			this.fieldName = fieldName;
	}
	validate(questionObj) {
		const minLengthValidator = (chain) =>
			chain
				.isLength({ min: this.minLength.minLength })
				.withMessage(this.minLength.minLengthMessage);
		const maxLengthValidator = (chain) =>
			chain
				.isLength({ max: this.maxLength.maxLength })
				.withMessage(this.maxLength.maxLengthMessage);
		const regexValidator = (chain) =>
			chain.matches(new RegExp(this.regex.regex)).withMessage(this.regex.regexMessage);

		let chain = body(this.fieldName ? this.fieldName : questionObj.fieldName);
		if (this.minLength.minLength) {
			chain = minLengthValidator(chain);
		}
		if (this.maxLength.maxLength) {
			chain = maxLengthValidator(chain);
		}
		if (this.regex.regex) {
			chain = regexValidator(chain);
		}
		return chain;
	}
}

module.exports = StringValidator;`

This is a modified version of https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/main/packages/forms-web-app/src/dynamic-forms/validator/string-validator.js which already does max length validation based on configuration.

packages/forms-web-app/src/dynamic-forms/dynamic-components/boolean-text/index.njk
```diff
@@ -9,12 +9,14 @@
 
 {% set options = [] %}
 
+{% set lastAnswer = payload and payload[question.fieldName] or answer %}
```
chrishawldev, line 12: Nice solution to passing back the value in error scenario's

chrishawldev (approved):

PR #3115, opened 2023-10-05
Feature/aapd 313

chrisHowarthkc:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-313

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+					caseReference: journeyResponse.caseReference,
+					isAppealTypeAppropriate: journeyResponse.answers['correct-appeal-type'] ? true : false,
+					doesTheDevelopmentAffectTheSettingOfAListedBuilding: journeyResponse.answers[
+						'affects-listed-building'
+					]
+						? true
+						: false,
+					affectedListedBuildings: () => {
+						if (journeyResponse.answers['affects-listed-building']) {
+							return journeyResponse.answers['listed-building-number'];
```
chrishawldev, line 16: Do we need a bit of logic to transform this data to conform to the PINS data model (e.g without the id we attach to it?)

chrishawldev: It could maybe be a utility function to remove the id from any list add more form response?

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+						}
+					},
+					siteWithinGreenBelt: journeyResponse.answers['green-belt'] ? true : false,
```
chrishawldev, line 19: Do we need to check the value of the answer here (e.g yes vs no instead of truthy?). If so could be another utility / helper function?

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+					howYouNotifiedPeople: journeyResponse.answers['notification-method'] ? true : false,
```
chrishawldev, line 20: I think this might need to be an array to support  "howYouNotifiedPeople": [
            "A public notice at the site",
            "Letters to neighbours",
            "Advert in the local press"
        ]

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+						if (journeyResponse.answers['safety-risks']) {
+							return journeyResponse.answers['affects-listed-building'];
+						}
+					},
+					doesSiteRequireInspectorAccess: journeyResponse.answers['inspector-visit-appellant']
+						? true
+						: false,
+					//journeyResponse.answers['inspector-visit-neighbour'] - should this be housed in the same property as above possibly doNeightboursAffectNeighboringSite
+					// newPlanningConditions: journeyResponse.answers['new-planning-conditions'] ? true : false, // fieldname needs clarifying
+					// journeyResponse.answers['other-ongoing-appeals'] // fieldname needs clarifying
```
edpackard-work, line 32: the fieldname for the boolean is `new-planning-conditions`
the fieldname for the text value if the user answers 'yes' for the above is `new-planning-conditions_new-conditions-value`

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+						}
+					},
+					documents: [
+						{
+							notifyingParties: journeyResponse.answers['notified-who']
+						},
+						{
+							conservationAreaMap: () => {
+								if (journeyResponse.answers['conservation-area']) {
+									journeyResponse.answers['conservation-area-upload'];
```
chrishawldev, line 45: Think this could just be called conservation-upload

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+								}
+							}
+						},
+						{
+							siteNotices: () => {
+								if (journeyResponse.answers['display-site-notice']) {
+									journeyResponse.answers['upload-site-notice'];
```
chrishawldev, line 52: Looks good

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+								}
+							}
+						},
+						{
+							notifyingParties: () => {
+								if (journeyResponse.answers['letters-to-neighbours']) {
+									journeyResponse.answers['upload-letters-emails'];
```
chrishawldev, line 59: Looks good

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
chrishawldev: Just looked in the HAS journey definition and we're missing a withCondition on the associated question. Can we resolve that as part of this please?

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+						{
+							pressAdvert: () => {
+								if (journeyResponse.answers['press-advert']) {
+									journeyResponse.answers['upload-press-advert'];
+								}
+							}
+						},
+						{
+							representations: () => {
+								if (journeyResponse.answers['representations-other-parties']) {
```
chrishawldev, line 72: Looks good

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+						if (journeyResponse.answers['affects-listed-building']) {
+							return journeyResponse.answers['listed-building-number'];
+						}
+					},
+					siteWithinGreenBelt: journeyResponse.answers['green-belt'] ? true : false,
+					howYouNotifiedPeople: journeyResponse.answers['notification-method'] ? true : false,
+					doesSiteHaveHealthAndSafetyIssues: journeyResponse.answers['safety-risks'] ? true : false,
+					healthAndSafetyIssuesDetails: () => {
+						if (journeyResponse.answers['safety-risks']) {
+							return journeyResponse.answers['affects-listed-building'];
```
chrishawldev, line 24: We might have to wait on the safety issues question to be added to get the name of the conditional field to use here

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+						}
+					},
+					doesSiteRequireInspectorAccess: journeyResponse.answers['inspector-visit-appellant']
+						? true
+						: false,
+					//journeyResponse.answers['inspector-visit-neighbour'] - should this be housed in the same property as above possibly doNeightboursAffectNeighboringSite
+					// newPlanningConditions: journeyResponse.answers['new-planning-conditions'] ? true : false, // fieldname needs clarifying
+					// journeyResponse.answers['other-ongoing-appeals'] // fieldname needs clarifying
+					nearbyCaseReferences: () => {
+						if (journeyResponse.answers['other-ongoing-appeals']) {
```
chrishawldev, line 34: Looks good

packages/appeals-service-api/src/services/responses.service.js
```diff
@@ -1,8 +1,11 @@
 const ApiError = require('../errors/apiError');
 const logger = require('../lib/logger');
 const { ResponsesRepository } = require('../repositories/responses-repository');
-
 const responsesRepository = new ResponsesRepository();
+const { HasQuestionnaireMapper } = require('../mappers/questionnaire-submission/has-mapper');
+const questionnaireMapper = new HasQuestionnaireMapper();
+const { LocalEventClient } = require('@pins/common/src/event-client/local-event-client');
```
chrishawldev, line 7: This will need to use the service bus client. There's an example of how back office do this here, they basically allow you to configure if it should use the service bus client or the local at run time https://github.com/Planning-Inspectorate/back-office/blob/main/packages/event-client/src/event-client.js. I would suggest we default to true for using the service bus client then set false locally on the docker compose file for the appeals service api.

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,87 @@
...
+					healthAndSafetyIssuesDetails: () => {
+						if (journeyResponse.answers['safety-risks']) {
+							return journeyResponse.answers['affects-listed-building'];
+						}
+					},
+					doesSiteRequireInspectorAccess: journeyResponse.answers['inspector-visit-appellant']
+						? true
+						: false,
+					//journeyResponse.answers['inspector-visit-neighbour'] - should this be housed in the same property as above possibly doNeightboursAffectNeighboringSite
+					// newPlanningConditions: journeyResponse.answers['new-planning-conditions'] ? true : false, // fieldname needs clarifying
```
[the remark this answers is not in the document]
chrishawldev: I need to come back to this one actually. Let me take a look now.

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
```diff
@@ -0,0 +1,114 @@
+class HasQuestionnaireMapper {
+	mapToPINSDataModel(journeyResponse) {
+		return [
+			{
+				questionnaire: {
+					LPACode: journeyResponse.LPACode,
+					caseReference: journeyResponse.caseReference,
+					isAppealTypeAppropriate: this.#convertToBoolean(
+						journeyResponse.answers['correct-appeal-type']
```
edpackard-work, line 9: would it be possible to get these fieldName values directly from `questions.appealTypeAppropriate.fieldName` etc.?

bryn-pins (commented):
Couple of other questions:

- can't see a submit button locally
- after submitting manually by api, nothing seems to change on the appeal status, if we're waiting on back office, will this lead to users submitting multiple times thinking that it isn't working?

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
chrishawldev: It would, and I think Bryn has raised this below and it is something I considered. My thinking is that we'll have to do something similar for S78 and we can try to scope out opportunities for refactoring at that point, and that would be one of them I think.

packages/appeals-service-api/src/controllers/responses.js
```diff
@@ -29,4 +34,23 @@ const getResponseByReferenceId = async (req, res) => {
...
 
-module.exports = { patchResponseByReferenceId, getResponseByReferenceId };
+const submitQuestionnaireResponse = async (req, res) => {
+	let statusCode = 200;
+	let body = {};
+	let questionnaireResponse = {};
+	try {
+		questionnaireResponse = await getResponse(req.params.journeyId, req.params.referenceId);
+	} catch (error) {
+		statusCode = error.code;
```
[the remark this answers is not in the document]
chrishawldev: Possibly, but at this point I'm not sure what they would be. I'm just looking at this and actually I don't think we should have the finally block, it should all just be in the try catch to my mind. What do you reckon @bryn-pins ?

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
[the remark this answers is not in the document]
chrishawldev: Yeah I think thats a good refactoring point for the future with s78 on the Horizon and needing to build out the mapper for that.

packages/appeals-service-api/src/routes/responses.js
```diff
@@ -1,11 +1,13 @@
...
 const {
 	patchResponseByReferenceId,
-	getResponseByReferenceId
+	getResponseByReferenceId,
+	submitQuestionnaireResponse
 } = require('../controllers/responses');
 const router = express.Router();
 
-router.patch('/:journeyId/:referenceId', patchResponseByReferenceId);
+router.patch('/:journeyId/:referenceId/:lpaCode', patchResponseByReferenceId);
```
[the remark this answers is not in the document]
chrishawldev: It's primarily driven by the PINS data model and the need to ensure that we pass it on with questionnaire submissions https://pins-ds.atlassian.net.mcas.ms/wiki/spaces/AAPDS/pages/1249312781/LPA+Questionnaire+submission+data, however it means that we can use it for authorization purposes so for LPA users we can validate that the response they are working on belongs to their LPA, which is useful as the appeal references aren't necessarily secret. In terms of storing response for none LPA users, I wouldn't discount storing them in a different table and that at run time we would know the role of the users to determine where we should store their journey responses. We will also know the LPA code for those types of user response submissions as well.

packages/appeals-service-api/src/controllers/responses.js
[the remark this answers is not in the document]
bryn-pins: I was doing this with listed building:

https://github.com/Planning-Inspectorate/appeal-planning-decision/blob/main/packages/appeals-service-api/src/routes/listed-building.js

check for our custom error message type if it isn't then just rethrow and allow it to bubble up

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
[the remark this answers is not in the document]
bryn-pins: would be good to get tickets in regarding some of these refactoring issues that we're pushing down the line, as I'm sure there have been others but I've definitely forgotten them

packages/appeals-service-api/src/routes/responses.js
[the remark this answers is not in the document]
bryn-pins: In my head it could just be part of the object, not the url, but if you think this will be preferable that's fine. Although I would then ask should the other response api endpoints not also require the lpaCode?

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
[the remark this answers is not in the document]
chrishawldev: Yeah I think if we get the user story in for the s78 submission we can add them as subtasks to that, will ask for a ticket now.

packages/appeals-service-api/src/routes/responses.js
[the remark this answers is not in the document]
chrishawldev: In the body we only post the answers we want to update at the moment, everything else cones from the params.

We could refactor that but I don't think there's a great deal of value in that as for other user types we might not care about the LPA code in the future so I don't want to bind that to the payload as it might be more straight forward down the line to say user role x send patch requests here user role sends patch requests there, a bit of a service discovery pattern based on a users role. 

In that respect from a metric and observability point of view  you can monitor requests from different user roles based on the url which is easier to query and observe than a request body.

We shouldn't need the LPA code to query the data on the API because in this instance the LPACode doesn't actually help us query the data, we've got a unique lookup based on the journey id and the appeal reference, in future for other user types we might have a different synthetic id that we use to query journey responses.

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
bryn-pins: I think there's at least three things here:
1: only sending through answers to questions that are conditionally valid for the journey based on the response. This should always be the same as we control what questions are valid to have been answered
2: lookup fieldnames that are hardcoded strings here
3: modifying the fieldnames/values/structure/metadata based on data model/requirements from back office

moving the journey to shared business rule can solve 1/2
this mapper can then still handle any requirements of 3

Happy for it to be a refactor job though

packages/appeals-service-api/src/routes/responses.js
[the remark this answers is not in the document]
bryn-pins: think it just feels a little off to me as I would normally attempt to make apis at least somewhat restful, so seeing different params for the same resource seems odd, but it's not a blocker and easy enough to change later on as we don't have external consumers

bryn-pins (approved):
Happy to get this over the line for now

packages/appeals-service-api/src/mappers/questionnaire-submission/has-mapper.js
chrishawldev: Yeah I think it would have to be a refactor and we'd need to get some requirements from s78 as the journey there is more complex in comparison to HAS so right now I don't think we'd capture everything, you're still creating a coupling between the forms web app, the api and the pins data model it's just a case of identifying that risk and how to mitigate it. I'm imagining potential for the journey to become increasingly complex to serve both sub systems, but that's based purely on the fact of only having seen the HAS questionnaire and the WIP in progress for S78 and not knowing a great deal yet about other journeys and or journeys for other users, plus with the data model still in development there's a lot of unknowns for me that makes a refactor in the short term less valuable because it's likely to be refactored again.

PR #3121, opened 2023-10-06
feat(aapd-370): add http listed building trigger function

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-370

## Description of change

- may still need to add a role to write messages for the function app in infra repo
- expecting to potentially share the service bus client added from back-office in aapd 313, kept it simple for now
- Added http trigger function to add listed building messages in batches to the listed building service bus topic
- Using stream-json to handle the array data from the file as a stream, JSON.parse uses a lot of memory
- Currently batching logic is done twice, once during parsing the json and once when sending topic messages
Moving to only the servicebus client level would require loading the entire file into memory
Moving to only topic level risks going over any potential maxBatchSize/defaults set in the function configuration

## Checklist

- [?] Requires infrastructure changes
- [x] New features have tests

## Important


packages/integration-functions/common/src/azure-service-bus/service-bus.js
```diff
@@ -0,0 +1,45 @@
...
+const { DefaultAzureCredential } = require('@azure/identity');
+
+/**
+ * todo: potentially extend/reuse class from back-office/aapd 313
+ * however will listed building messages need the message transformations used there?
+ * @param {String} serviceBusHostname
+ * @param {String} topicName
+ * @param {Array.<Object>} messages
+ */
+async function sendMessageBatch(serviceBusHostname, topicName, messages) {
```
chrishawldev, line 11: I know the documentation is awful, but possibly this might work across different types of queue output function context.extraOutputs.set(queueOutput, ['message 1', 'message 2']); If not this code looks fine to me.

chrishawldev (approved):
Just one query around whether we need the custom service bus client but seeing as this is a one off I'm happy to go with it as it looks good regardless

packages/integration-functions/common/src/azure-service-bus/service-bus.js
bryn-pins: if I get some time I'll look into how/if this works

PR #3171, opened 2023-10-19
feat(aapd-402): add date component and community levy adopted date qu…

edpackard-work:
…estion

## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-402

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/validator/date-validator.js
+++ b/packages/forms-web-app/src/dynamic-forms/validator/date-validator.js
@@ -1,0 +1,158 @@
+const { body } = require('express-validator');
+const { endOfDay, isAfter, isValid, parse } = require('date-fns');
+const { enGB } = require('date-fns/locale');
+
+const BaseValidator = require('./base-validator.js');
+const { dateInputsToDate } = require('../dynamic-components/utils/date-inputs-to-date.js');
+
+/**
+ * @typedef {import('../question.js')} Question
+ */
+
+/**
+ * enforces a user has entered a valid date
+ * @class
+ */
+class DateValidator extends BaseValidator {
+	/**
+	 * creates an instance of a DateValidator
+	 * @param {string} [inputLabel] - string representing the field as displayed on the UI as part of an error message
+	 * @param {Object} [errorMessage] - object containing custom error messages to show on validation failure
+	 */
+	constructor(inputLabel, errorMessages) {
+		super();
+
+		const defaultErrorMessages = this.#getDefaultErrorMessages(inputLabel);
+
+		this.emptyErrorMessage =
+			errorMessages?.emptyErrorMessage ?? defaultErrorMessages.emptyErrorMessage;
+		this.noDayErrorMessage =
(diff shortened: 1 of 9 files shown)


packages/forms-web-app/src/dynamic-forms/validator/date-validator.js
```diff
@@ -0,0 +1,153 @@
...
+				}),
+			body(yearInput)
+				.notEmpty()
+				.withMessage((_, { req }) => {
+					if (req.body[dayInput] && req.body[monthInput]) return this.noYearErrorMessage;
+				}),
+
+			// check date values entered are within valid ranges
+			body(dayInput)
+				.isInt({ min: 1, max: 31 })
```
chrishawldev, line 94: I'd consider a utility method like so
`var isValid = require('date-fns/isValid')
isValidWrapper(year = 2000, month = 1, day = 1)
{
isValid(new Date(year, month, 1)) : true ? throw InvalidMonthError("Month is invalid month " + month);
}
// Then we can do, which will work on any day, we can do the same to check the month and year on their own above
if (isValidWrapper(day = value) {
						const yearNum = parseInt(req.body[yearInput], 10);
						return yearNum && isLeapYear(new Date(yearNum, 1, 1)) && value === 29;
					}

`

packages/forms-web-app/src/dynamic-forms/validator/date-validator.js
```diff
@@ -0,0 +1,153 @@
...
+					}
+
+					return true;
+				})
+				.withMessage(this.invalidDateErrorMessage),
+			body(monthInput).isInt({ min: 1, max: 12 }).withMessage(this.invalidDateErrorMessage),
+			body(yearInput).isInt({ min: 1000, max: 9999 }).withMessage(this.invalidYearErrorMessage),
+
+			//check date is not in the future
+			body(dayInput).custom((value, { req }) => {
```
chrishawldev, line 114: We might need to be able to configure if the validation is a check on the future or the past, I think there is another question where it checks the date is allowed to be in the past but not the future

packages/forms-web-app/src/dynamic-forms/validator/date-validator.js
```diff
@@ -0,0 +1,153 @@
...
+				const day = `0${value}`.slice(-2);
+				const month = `0${req.body[monthInput]}`.slice(-2);
+				const year = req.body[yearInput];
+
+				const dateInput = `${year}-${month}-${day}`;
+				const today = endOfDay(new Date());
+
```
chrishawldev, line 121: Neat

packages/forms-web-app/src/dynamic-forms/validator/date-validator.js
edpackard-work: as discussed, we do not need to implement this functionality for this ticket - the existing functionality can be adapted in a future ticket when requirements demand.

chrishawldev (approved):

PR #3218, opened 2023-11-01
feat(aapd-570): added list add more for nearby appeals

ms-ek:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-570

## Description of change

Added component to add several nearby appeals

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/config.js
+++ b/packages/forms-web-app/src/config.js
@@ -94,7 +94,7 @@ module.exports = {
 		stringValidation: {
 			listedBuildingNumber: {
 				regex: {
-					regex: '[0-9]',
+					regex: '^\\d+$',
 					regexMessage: 'Enter a list entry number using numbers 0 to 9'
 				},
 				maxLength: {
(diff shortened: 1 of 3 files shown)


packages/forms-web-app/src/config.js
```diff
@@ -105,6 +105,20 @@ module.exports = {
 					minLength: 7,
 					minLengthMessage: 'List entry number must be 7 digits'
 				}
+			},
+			appealReferenceNumber: {
+				regex: {
+					regex: '[0-9]',
```
Si3rr4wow, line 111: Is this supposed to be just one digit?

Si3rr4wow: Oh wait does the min max thing like iterate through and check each char with this regex?

ms-ek: No it should be several digits, good spot, just changed it.

packages/forms-web-app/src/config.js
```diff
@@ -94,7 +94,7 @@ module.exports = {
 		stringValidation: {
 			listedBuildingNumber: {
 				regex: {
-					regex: '[0-9]',
+					regex: '^[0-9]+$',
```
Si3rr4wow, line 97: I'm gunna seem really picky now but Sonar recommends using the character classes rather than sets in cases like this, ie `/^\d+$/` 

Screenshot 2023-11-01 at 12 38 15

Si3rr4wow (approved):
LGTM ⭐

bryn-pins (approved):

PR #3225, opened 2023-11-02
Feature/aapd 670

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-670

## Description of change

- added new api route
- added /case into get route by caseref as it felt like it should be possible to do GET: /{documentid} at some point in future

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/document-metadata-repository.js
+++ b/packages/appeals-service-api/src/repositories/document-metadata-repository.js
@@ -1,18 +1,21 @@
 const { MongoRepository } = require('./mongo-repository');
 
+/**
+ * @typedef {import('@pins/common/schema/documentMetadata.js').DocumentMetadata} DocumentMetadata
+ * @typedef {import('@pins/common/schema/documentMetadata.js').DocumentType} DocumentType
+ * @typedef {import('./mongo-repository').UpdateResult} UpdateResult
+ */
+
 class DocumentMetadataRepository extends MongoRepository {
 	constructor() {
 		super('documentMetadata');
 	}
-	/**
-	 * @typedef {'Supporting Documents' | 'Appeal Statement' | 'Planning application form' | 'Decision notice'} DocumentType
-	 */
 
 	/**
 	 *
 	 * @param {string} caseRef
 	 * @param {DocumentType} documentType
-	 * @return {Promise<any>}
+	 * @return {Promise<Array.<DocumentMetadata>|DocumentMetadata|undefined>}
 	 */
 	async getDocumentMetadata(caseRef, documentType, returnMultipleDocuments) {
 		if (returnMultipleDocuments) {
@@ -50,6 +53,22 @@ class DocumentMetadataRepository extends MongoRepository {
 				console.log(e);
 			}
 		}
+	}
+
+	/**
+	 * Upsert document metadata
+	 * @param {DocumentMetadata} documentMetadata
+	 * @returns {Promise<UpdateResult>}
+	 */
+	async createUpdateDocumentMetadata(documentMetadata) {
+		const filter = {
+			caseRef: `${documentMetadata.caseRef}`,
+			documentId: `${documentMetadata.documentId}`
+		};
+
+		return await this.updateOne(filter, {
+			$set: documentMetadata
+		});
 	}
 }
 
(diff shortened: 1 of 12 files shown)


packages/appeals-service-api/src/repositories/document-metadata-repository.js
```diff
@@ -51,6 +51,21 @@ class DocumentMetadataRepository extends MongoRepository {
 			}
 		}
 	}
+
+	/**
+	 * @param {*} documentMetadata
+	 * @returns
+	 */
```
Si3rr4wow, line 58: idk how these work exactly but something like this?

```
/**
 * @typedef {Object} DocumentMetadata
 * idk tbh
 */


/**
 * @typedef {Object} UpdateResult
 * @property {number} matchedCount
 * @property {number} modifiedCount
 */

/**
 * @param {DocumentMetadata} documentMetadata
 * @returns UpdateResult
 */
```

packages/appeals-service-api/src/services/document-metadata.service.js
```diff
@@ -19,6 +25,17 @@ const getDocumentMetadata = async (caseRef, documentType, returnMultipleDocument
...
 };
 
+const createOrUpdateDocumentMetadata = async (documentMetadata) => {
+	// check schema with joi
+	const validationResult = documentMetaDataSchema.validate(documentMetadata);
+	if (validationResult.error) {
+		throw ApiError.badRequest();
+	}
+
+	await documentMetadataRepository.createUpdateDocumentMetadata(documentMetadata);
```
Si3rr4wow, line 35: If you just return this function call then this function won't need any async/await sugar but will still work the same way

Si3rr4wow (dismissed):
slicc 👍

packages/appeals-service-api/src/repositories/document-metadata-repository.js
bryn-pins: Obviously forgot about this, have moved the doc type into common and also imported the mongo type

packages/appeals-service-api/src/routes/documentMetadata.js
```diff
@@ -12,11 +16,33 @@ router.get('/:caseref', async (req, res) => {
...
+
+	res.status(statusCode).send(body);
+});
+
+router.put('/:id', async (req, res) => {
+	let statusCode = 200;
+	let body = {};
+	try {
+		req.body.documentId = req.params.id;
+		await createOrUpdateDocumentMetadata(req.body);
```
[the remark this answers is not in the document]
bryn-pins: have added a mapper to do this in case we want to standardise/reuse with other endpoints

packages/appeals-service-api/src/services/document-metadata.service.js
```diff
@@ -1,6 +1,17 @@
 const logger = require('../lib/logger');
 const { DocumentMetadataRepository } = require('../repositories/document-metadata-repository');
 const ApiError = require('../errors/apiError');
+const Joi = require('joi');
+
+/**
+ * @typedef {import('../repositories/document-metadata-repository.js').DocumentMetadata} DocumentMetadata
+ * @typedef {import('../repositories/document-metadata-repository').UpdateResult} UpdateResult
```
Si3rr4wow, line 8: how come one of these has `.js` and the other doesn't?

packages/integration-functions/document-metadata-fo-integration/index.js
```diff
@@ -3,43 +3,12 @@ const VALID_SCAN_STATUSES = ['scanned'];
 const VALID_PUBLISHED_STATUSES = ['published', 'archived'];
 
 /**
- * @typedef {Object} ServiceBusMessage
- * @property {DocumentPayload} body
+ * @typedef {import('@pins/common/schema/documentMetadata.js').DocumentMetadata} DocumentMetadata
```
Si3rr4wow, line 6: Does fo stand front office in this file name?

Si3rr4wow: Talking about "packages/integration-functions/document-metadata-fo-integration/index.js", not the target of this import btw, didn't know where else to attach this

packages/appeals-service-api/src/services/document-metadata.service.js
bryn-pins: makes no difference, just wasn't consistent

packages/integration-functions/document-metadata-fo-integration/index.js
bryn-pins: yes, was just following the naming from appeals-case-data, not sure it's really needed as it lives in a function app with that name already: pins-func-appeals-fo-integration

didn't do it with listed building ones as there is no equivalent in Back Office

Si3rr4wow (approved):

PR #3227, opened 2023-11-03
ci(as-5891): add pr pipeline, without build steps

beejjacobs:
## Ticket Number

[AS-5891](https://pins-ds.atlassian.net/browse/AS-5891)

## Description of change

Add in a PR pipeline which does linting & tests, to avoid building on every PR when the artefacts are not used.

**Note: this will require a rules change as Build checks won't run on PR**

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/azure-pipelines-pr.yml
+++ b/azure-pipelines-pr.yml
@@ -1,0 +1,42 @@
+trigger: none
+
+pr:
+  branches:
+    include:
+      - '*'
+
+resources:
+  repositories:
+    - repository: templates
+      type: github
+      endpoint: Planning-Inspectorate
+      name: Planning-Inspectorate/common-pipeline-templates
+      ref: refs/tags/release/3.5.0
+
+extends:
+  template: stages/wrapper_ci.yml@templates
+  parameters:
+    globalVariables:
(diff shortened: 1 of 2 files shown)


edpackard-work (approved):

azure-pipelines-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+              nodeVersion: 18
+          - template: ../steps/node_script.yml
+            parameters:
+              condition: succeededOrFailed()
+              script: npm run test
+              nodeVersion: 18
+          - template: ../steps/node_script.yml
+            parameters:
+              condition: succeededOrFailed()
+              script: npm run integration-test
```
bryn-pins, line 41: So previously we had integration tests only running on main
They used to be flaky so were causing us to need to rebuild to get through a PR
Did spend some time resolving that and haven't seen it fail on main for a while but worth noting we may need to take this out or spend some more time looking at them

beejjacobs: That's good to know, I mis-read the condition!

I checked a few runs at 'random' and they've all run OK. Let's keep it in, and review.

PR #3252, opened 2023-11-10
dev: switch to npm workspaces

beejjacobs:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-809

## Description of change

Moving to npm workspaces:

- remove individual package-lock files
- update Dockerfiles, pipelines, and builds to support workspaces
- update docs
- some fixes to forms-web-app to use `require.resolve`
- added placeholder `web-comment` package (fixed issue with `webpack` being hoisted by `webpack-cli` not being hoisted

Important packages with changes (yes I reviewed all/most package-lock changes...):

- `@azure/storage-blob` - 12.16.0 (bump from 12.10.0 for docs-api)
- `express`  - 4.18.2 (bump from 4.18.1 for some packages)
- `mongodb` - api uses 3.7.4, forms uses 3.6.12, docs uses 3.7.3 - unchanged
- `notifications-node-client` - 5.2.3 (bump from 5.1.1)
- `nunjucks` - 3.2.4 (bump from 3.2.3)
- `webpack` - 5.89.0 (bump from 5.76.1)

Tested locally. Need to check functions build + deploy OK - script(s) for build copied from back-office.

## Checklist

- [ ] New features have tests

## Important


beejjacobs:
Need to fix tests - something not working with `jest-fetch-mock`.

chrishawldev (dismissed):
Looks good to me. Would be interested to see if there is much difference in size in docker images, e.g is it more or less efficient or no real difference.

beejjacobs:
> Looks good to me. Would be interested to see if there is much difference in size in docker images, e.g is it more or less efficient or no real difference.

Good question. Will be interesting to see. Here's what looks like a typical output currently: 
image

docker-compose.yml
```diff
@@ -281,6 +289,8 @@ services:
     depends_on:
```
bryn-pins, line 289: appeals-service-api-data:
    build:
      context: ./data
      
 Given we're in ./data
 I don't think the adding of package-lock works

beejjacobs: Yes good spot, I had build from root not checked docker-compose. Sorted in next commit.

beejjacobs:
Fixed `data` and `mocked-services` Docker builds now, I think!

beejjacobs:
First successful build: 
image

Looks like forms-web-app is quite a bit large. Might be worth some time to check that.

chrishawldev:
> First successful build: image
> 
> Looks like forms-web-app is quite a bit large. Might be worth some time to check that.

Is this because we're now copying in the package lock from the root workspace like
COPY package.json /opt/
COPY package-lock.json /opt/
Which in theory will contain dependencies for other modules as well?
I don't think this is that big of an issue, best practise says keep your containers as small as possible but in reality having an extra 150Mb on the container size to have a pretty quick win with using npm workspaces is a good trade off IMHO and it's something we can come back to with a separate task in the future.

chrishawldev (approved):

docker-compose.yml
bryn-pins, line 1: Did we want to add the azurite version?
image: mcr.microsoft.com/azure-storage/azurite:3.26.0

bryn-pins (approved):

PR #3271, opened 2023-11-15
feat(aapd-617): prisma set up added

ms-ek:
## Ticket Number

## Description of change

Added Prisma set up with placeholder database schema.
Prisma is responsible for:
- Creating the database
- Creating the database tables
- Populating the tables with some initial data (see example below)

Screenshot 2023-11-15 at 14 59 48

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,23 @@ npm install <new-package>
 
 A SQL Server database server has been set up to run when the appeals service api is running.
 
+First, make sure you have a `.env` file in `./packages/appeals-service-api` (you can copy the `.env.example`) and it has a `DATABASE_URL` environment variable defined with details pointing to your local database server (`mssql` Docker container).
+
+To set up the SQL Server with tables and some data, you will need to run the following commands (whilst the SQL Server Docker container is running. You can run it by running the appeals service api or forms web app. For example: `make run SERVICE=appeals-service-api`. Alternatively, you can run the Docker container called 'mssql' manually using the Docker interface):
+
+```shell
+npm run db:generate
+npm run db:migrate:dev
+npm run db:seed 
+```
+**npm run db:generate:** To generate the client from the schema
+**run db:migrate:dev:** To apply changes → Creating the database and tables
+**run db:seed :** Populating the database with some initial data
+
+The ORM used by the application to access SQL Server is [Prisma](https://www.prisma.io/). The schema is defined in [schema.prisma](./packages/appeals-service-api/prisma/schema.prisma). 
+
+**Note:** If the `prisma.schema` file has been updated, don't forget to run `npm run db:migrate:dev` to apply the changes.
+
 ##### Azure Data Studio
 
 [Azure Data Studio](https://learn.microsoft.com/en-us/sql/azure-data-studio/download-azure-data-studio) or the [MSSQL VS code extension](https://marketplace.visualstudio.com/items?itemName=ms-mssql.mssql) can be used as a database client, to create and monitor the database contents.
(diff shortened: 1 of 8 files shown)


bryn-pins (commented):
Should this have a migration file committed?
And I guess we may want to leave merging until we have the schema ready
Only other bit is I guess the pipelines running the migrate script but those should probably also be done separately as it requires my infra PR going in as well

packages/appeals-service-api/src/db/prisma/schema.prisma
beejjacobs, line 1: Can we drop the `prisma` folder? I'm not a huge fan of naming things after a particular library etc... - I think the `src/db` folder is clear enough; I can't think what else we'd have in there.

README.md
```diff
@@ -44,6 +44,25 @@ npm install <new-package>
...
+Create a `.env` file in `./packages/appeals-service-api`. Copy the values from `.env.development`.
+
+To set up the SQL Server with tables and some data, you will need to run the following commands (whilst the SQL Server Docker container is running. You can run it by running the appeals service api or forms web app. For example: `make run SERVICE=appeals-service-api`. Alternatively, you can run the Docker container called 'mssql' manually using the Docker interface):
+
+```shell
+npm run db:generate
+npm run db:migrate:dev
+npm run db:seed 
+```
+**npm run db:generate:** To register the schema
```
beejjacobs, line 58: Perhaps "To generate the client from the schema" is more descriptive?

README.md
```diff
@@ -44,6 +44,25 @@ npm install <new-package>
 
 A SQL Server database server has been set up to run when the appeals service api is running.
 
+First, make sure you have a `.env.development` file in `./packages/appeals-service-api` and it has a `DATABASE_URL` environment variable defined with details pointing to your local database server (`mssql` Docker container).
```
beejjacobs, line 47: Do we need `.env.development` or just the `.env` one? If that is just an example, let's rename it `.env.example` to make that clear, and we could edit this paragraph?

> First, make sure you have a `.env` file in... (you can copy the `.env.example`)...

packages/appeals-service-api/package.json
```diff
@@ -7,7 +7,14 @@
...
+		"schema": "src/db/prisma/schema.prisma",
+		"seed": "node src/db/prisma/seed/seed.js"
+	},
 	"scripts": {
+		"db:migrate:dev": "npx prisma migrate dev",
+		"db:generate": "npx prisma generate",
+		"db:seed": "npx prisma db seed",
 		"lint": "eslint ./",
 		"lint:fix": "eslint ./ --fix",
 		"start": "node ./",
```
beejjacobs, line 20: We need to run generate before  `start:dev` else we won't have a generated client to use. Just `npm run db:generate && ...` should do it.

We probably need to add `npm run db:generate` to the Dockerfile for the API to do this at build time as well.

beejjacobs (changes requested):
Looks good! I've got a few points, open to comments and thoughts.

Also, on back office we used a prettier plugin to ensure the schema was formatted on commit: https://github.com/Planning-Inspectorate/back-office/commit/5b46dd6eb0a66f0d0d2bb034bf6a2610bada0ae6 - be nice to add that in.

packages/appeals-service-api/package.json
ms-ek: Or can we just add `npm run db:generate` before the `start` script since this is what Docker calls?

beejjacobs: Sure, either way works!

beejjacobs (dismissed):
Great work! I'll do a separate PR with a real schema after this.

beejjacobs (approved):

PR #3274, opened 2023-11-16
feat(add local docker file): add a local docker file

Si3rr4wow:
add a local docker file so I don't have to keep commenting out the pdf thingy

This has bad down behaviour which I can't think of a fix for off the top of my head. I'll convert to draft

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,10 @@ serve:
 	docker-compose up
 .PHONY: serve
 
+serve-local:
+	docker-compose -f docker-compose-local.yml up
+.PHONY: serve
+
 uninstall:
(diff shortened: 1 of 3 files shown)


beejjacobs:
@Si3rr4wow I had a quick look into this.

Some info here: https://github.com/docker/compose/issues/1294

This seems to work for me: `docker-compose up --scale pdf-service-api=0`

We could also look at profiles: https://docs.docker.com/compose/profiles/

beejjacobs:
Should be sorted by: https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/3365

PR #3275, opened 2023-11-16
Feature/aapd 617 sql schema

beejjacobs:
## Ticket Number

## Description of change

Add an initial SQL schema with `AppealUser` as a focus. Updated the seed script too.

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/.env.example
```diff
@@ -4,4 +4,4 @@
 # Prisma supports the native connection string format for PostgreSQL, MySQL, SQLite, SQL Server, MongoDB and CockroachDB.
 # See the documentation for all the connection string options: https://pris.ly/d/connection-strings
 
-DATABASE_URL="sqlserver://0.0.0.0:1433;database=pins_front_office_development;user=sa;password=DockerDatabaseP@22word!;trustServerCertificate=true"
\ No newline at end of file
+DATABASE_URL="sqlserver://localhost:1433;database=pins_front_office_development;user=sa;password=DockerDatabaseP@22word!;trustServerCertificate=true"
```
beejjacobs, line 7: I needed this for it to work for me.

ms-ek (dismissed):

packages/appeals-service-api/src/db/schema.prisma
```diff
@@ -10,8 +7,119 @@ datasource db {
...
+
+  // link to an AppealUser, who is the appellant
+  // this is optional because we may get an appeal from back office before the associated (service) user entity
+  appellantUserId String?     @db.UniqueIdentifier
+  Appellant       AppealUser? @relation("appellant", fields: [appellantUserId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+
+  // link to an AppealUser, who is the agent
+  // this is optional because not all appeals have an agent
+  agentUserId String?     @db.UniqueIdentifier
+  Agent       AppealUser? @relation("agent", fields: [agentUserId], references: [id], onDelete: NoAction, onUpdate: NoAction)
```
PaulGruffyddAmaze, line 95: According to the PINS data model, `agent` is an array rather than a single entity so I wonder whether we should create a cross-reference table to allow the linking of multiple agents to an appeal. 

If we were to also add a field to the cross reference table for `role`, that would allow us to remove references to all types of users from the appeal table and manage the relationships between entities independently. That would also allow us to manage other types of relation between those entities e.g. `interestedParty`. Obviously the trade-off there is we'd need a table join to retrieve the data rather than querying a single table.

packages/appeals-service-api/.env.example
bryn-pins: Should have caught this in last PR, this isn't the naming I've used for Azure
Can change here or there
Maybe why dev is down?

https://github.com/Planning-Inspectorate/infrastructure-environments/blob/3134210d37501df66e0b20d4ef64704854093db8/app/components/appeals-app-services/locals.tf#L94

beejjacobs: Good catch - no the current issue on dev - but it will be!

We specify the env var to use in the schema file (datasource.url). I don't mind much which we use, but we need to align.

packages/appeals-service-api/src/db/schema.prisma
beejjacobs: Great point! I think a cross reference table is the right solution. I've implemented that.

packages/appeals-service-api/src/db/schema.prisma
```diff
@@ -10,8 +7,120 @@ datasource db {
...
+  AppealCase AppealCase?
+
+  // todo: in the future link to appeal submission and other data
+}
+
+// AppealToUser represents the relationships between AppealUser and Appeal
+// This is a many-to-many relationship.
+model AppealToUser {
+  // relation to an Appeal
+  appealId   String     @unique @db.UniqueIdentifier
```
PaulGruffyddAmaze, line 101: I don't think `appealId` would be unique in this instance as we can have multiple users (e.g. agent & appellant) linked to the same appeal.

packages/appeals-service-api/src/db/schema.prisma
```diff
@@ -10,8 +7,120 @@ datasource db {
...
+  Appeal     Appeal     @relation(fields: [appealId], references: [id])
+  // relation to a User
+  userId     String     @unique @db.UniqueIdentifier
```
PaulGruffyddAmaze, line 104: Similar to my comment above, a given user (particularly an agent) may be associated with multiple appeals so I don't think `userId` would be unique in this table.

beejjacobs: Definitely a copy-paste error!

packages/appeals-service-api/src/db/schema.prisma
beejjacobs: Definitely a copy-paste error!

packages/appeals-service-api/src/db/schema.prisma
```diff
@@ -10,8 +7,120 @@ datasource db {
...
+// while we don't rely on that for security, it adds an extra layer 
+// not everything needs this, but easier to make them all consistent and the increase in size (vs int) is negligible
+
+// AppealUser represents a user of our service
+// this may be an appellant, agent, or LPA employee
+model AppealUser {
+  id         String  @id @default(dbgenerated("newid()")) @db.UniqueIdentifier
+  email      String  @unique
+  // role may only be needed for lpaUser, and may not be populated otherwise
+  role       String? // appellant | agent | lpaUser
```
bryn-pins, line 23: Thinking again I guess this is correct and we do want a role on both AppealUser and AppealToUser as lpa users are not linked to appeals directly

However I guess this will just be lpauser | null 
at least to start with as an appeal user could be linked as either an appellant or an agent at the same time

packages/appeals-service-api/src/db/schema.prisma
```diff
@@ -10,8 +7,120 @@ datasource db {
...
+model AppealToUser {
+  // relation to an Appeal
+  appealId   String     @db.UniqueIdentifier
+  Appeal     Appeal     @relation(fields: [appealId], references: [id])
+  // relation to a User
+  userId     String     @db.UniqueIdentifier
+  AppealUser AppealUser @relation(fields: [userId], references: [id])
+
+  // the role the user has in relation to the appeal
+  role String // appellant | agent | interestedParty
```
bryn-pins, line 108: potentially use this a foreign key to a role table for clarity?

packages/appeals-service-api/src/db/schema.prisma
```diff
@@ -10,8 +7,120 @@ datasource db {
...
+
+  // link to case (from Back Office)
+  AppealCase AppealCase?
+
+  // todo: in the future link to appeal submission and other data
+}
+
+// AppealToUser represents the relationships between AppealUser and Appeal
+// This is a many-to-many relationship.
+model AppealToUser {
```
bryn-pins, line 99: I feel as if maybe AppealUser should just be called User?
That way this table name makes more sense (should be AppealUserToAppeal at the moment)
And LPA users aren't directly linked to an Appeal
Not that important though

packages/appeals-service-api/.env.example
beejjacobs: Going to update this in this PR to align with infra, which has a more specific name.

packages/appeals-service-api/src/db/schema.prisma
beejjacobs: Replaced in favour of `isLpaUser`

packages/appeals-service-api/src/db/schema.prisma
beejjacobs: Done.

beejjacobs:
Made some changes:

* Added `AppealToUserRole` table as an enum for user role
* Added `seed-dev` and `seed-prod` scripts, with some reference data
* use `SQL_CONNECTION_STRING` to match infra

bryn-pins (commented):
minor comment

packages/appeals-service-api/src/db/seed/data-dev.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * @type {import('@prisma/client').Prisma.AppealUserCreateInput}
+ */
+const users = [
```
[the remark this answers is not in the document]
beejjacobs: d'oh - yes.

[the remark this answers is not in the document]
beejjacobs: fixed

bryn-pins (approved):

PR #3285, opened 2023-11-20
feat(516): fix journey to be submittable

Si3rr4wow:
Tweaks the S78 questionnaire journey so it's submittable


packages/forms-web-app/src/dynamic-forms/s78-questionnaire/journey.js
```diff
@@ -74,8 +80,7 @@ class S78Journey extends Journey {
 					response.answers &&
 						response.answers[questions.screeningOpinionEnvironmentalStatement.fieldName] === 'yes'
 				)
-				.addQuestion(questions.developmentDescription)
-				.addQuestion(questions.sensitiveArea),
```
Si3rr4wow, line 78: This was in this section twice, causing you to get stuck in a loop

beejjacobs: Good catch. Probably wouldn't get caught until next sprint when they do navigation tickets.

beejjacobs (approved):
LGTM.

PR #3288, opened 2023-11-20
Feature/aapd 617 pipelines

bryn-pins:
## Ticket Number

## Description of change

Have added a migrate step into both deploy and release pipelines
Have set up a new seeding pipeline

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/azure-pipelines-db-seed.yml
+++ b/azure-pipelines-db-seed.yml
@@ -1,0 +1,62 @@
+parameters:
+  - name: environment
+    displayName: Environment
+    type: string
+    default: Dev
+    values:
+      - Dev
+      - Test
+      - Prod
+
+pool: pins-odt-agent-pool
+
+pr: none
+trigger: none
+
+resources:
+  repositories:
+    - repository: templates
+      type: github
+      endpoint: Planning-Inspectorate
+      name: Planning-Inspectorate/common-pipeline-templates
+      ref: refs/tags/release/3.4.0
+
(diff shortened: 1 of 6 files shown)


azure-pipelines-release.yml
beejjacobs, line 1: I don't think this is used anymore, but no harm having the change.

azure-pipelines-db-seed.yml
beejjacobs, line 1: I think it's useful to have a separate seed pipeline, in case we have some test data to seed or something later on without a code change.

I do wonder though if seed should also be part of deploy (possibly default false)? Since if we rely on any static data, it should be part of deployment.

beejjacobs (commented):
Looks good. One question below, but may not need changes.

azure-pipelines-db-seed.yml
bryn-pins: Think I've added this, hopefully without repeating the checkout and secret fetching steps

azure-pipelines-release.yml
bryn-pins: Part of me thinks we should aim to start using them again as currently we don't keep the last prod deployment, so reverting requires a complete rebuild and deploy if it has fallen off the list in the deploy pipeline

beejjacobs: OK, let's pick that up separately.

beejjacobs (approved):

PR #3290, opened 2023-11-21
fix(pipeline): deploy to test automatically and fix pipeline

bryn-pins:
## Description of change

- fix env var name in pipeline
- auto deploy to test

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/azure-pipelines-deploy.yml
+++ b/azure-pipelines-deploy.yml
@@ -78,13 +78,13 @@ extends:
                     parameters:
                       secrets:
                         - name: appeals-sql-server-connection-string-admin
-                          variable: DATABASE_URL
+                          variable: SQL_CONNECTION_STRING_ADMIN
                   - ${{ if eq(parameters.schemaMigration, true) }}:
                       - template: ../steps/node_script.yml
                         parameters:
                           nodeVersion: 18
                           environmentVariables:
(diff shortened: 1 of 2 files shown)


azure-pipelines-deploy.yml
```diff
@@ -60,6 +60,12 @@ resources:
 extends:
   template: stages/wrapper_cd.yml@templates
   parameters:
+    automaticDeployments:
```
beejjacobs, line 63: How does this work alongside the pipeline trigger? At the moment on main this pipeline is triggered after build.

```
  pipelines:
    - pipeline: build
      source: Appeals Service Build
      trigger:
        branches:
          include:
            - main
 ```

beejjacobs (commented):
Small question, but only based on my lack of understanding of these triggers.

azure-pipelines-deploy.yml
bryn-pins: So my understanding is:

This object gets passed through to wrapper_cd in common as a param and overwrites the default there which only does dev:
https://github.com/Planning-Inspectorate/common-pipeline-templates/blob/main/stages/wrapper_cd.yml

Then it loops over the auto deployment envs to deploy on CI builds

beejjacobs (dismissed):

beejjacobs (approved):

PR #3294, opened 2023-11-22
feat(aapd-645): restructured journey and refactored conditional code

Maby0:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/s78-questionnaire/journey.js
```diff
@@ -30,154 +33,131 @@ class S78Journey extends Journey {
...
+
 		this.sections.push(
 			new Section('Constraints, designations and other issues', 'constraints')
 				.addQuestion(questions.appealTypeAppropriate)
 				.addQuestion(questions.changesListedBuilding)
 				.addQuestion(questions.changedListedBuildingNumber)
-				.withCondition(
-					response.answers && response.answers[questions.changesListedBuilding.fieldName] == 'yes'
-				)
+				.withCondition(meetsCondition('changesListedBuilding', 'yes'))
```
beejjacobs, line 44: Looks great - my only thought/question.
Did you consider passing in the question rather than just a string? Makes it slightly longer, but we also (in theory) get autocomplete. 

e.g. `meetsCondition(questions.changesListedBuilding, 'yes')`

packages/forms-web-app/src/dynamic-forms/s78-questionnaire/journey.js
```diff
@@ -30,154 +33,131 @@ class S78Journey extends Journey {
 			journeyTitle
 		);
 
+		const meetsCondition = meetsConditionBuilder(response, questions);
```
PaulGruffyddAmaze, line 36: This may be slightly pedantic but I wonder whether the naming should be changed from `meetsCondition` to `questionHasAnswer` as this doesn't evaluate an arbitrary condition. As far as I can see, it always checks whether a given question has a given answer. That would mean that, when it's used, it would look something like this (also incorporating Ben's change):
`.withCondition(questionHasAnswer(questions.changesListedBuilding, 'yes'))`
To me, that makes it slightly more obvious what it's doing.

packages/forms-web-app/src/dynamic-forms/s78-questionnaire/journey.js
Maby0: thanks - have tweaked it to match your suggestion!

packages/forms-web-app/src/dynamic-forms/s78-questionnaire/journey.js
Maby0: sounds good - have renamed the functions

beejjacobs (approved):
Looks good. Makes the journey + conditions nice and readable too.

PR #3307, opened 2023-11-29
docs: use swagger-jsdoc for api specs

beejjacobs:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-171

## Description of change

Use `swagger-jsdoc` to generate an API spec for separate parts. I've have tried two approaches to get feedback on:

- `appeals-service-api` has comments inline in the router with `@openapi` comments
- `document-service-api` has route specs in a separate yaml file per router

I found writing yaml in a comment not particularly intuitive, which is why I tried the second approach. Feedback appreciated.

## Checklist

- [ ] New features have tests

## Important


bryn-pins (commented):
I think one of the reasons I like auto generated is that the definition lives next to code, which is a reminder to both coder and reviewer to update it with any changes made

if we had feature folders with the routes/controller next to the spec file think it would be ok

Also understand managing yml whitespace in a comment with no syntax highlighting could be a pain

Guess we could add as a checkbox to the PR template although it's gotten to the point where I gloss over those

beejjacobs:
> I think one of the reasons I like auto generated is that the definition lives next to code, which is a reminder to both coder and reviewer to update it with any changes made
> 
> If we had feature folders with the routes/controller next to the spec file think it would be ok as a spec file
> 
> Also understand managing yml whitespace in a comment with no syntax highlighting could be a pain
> 
> Guess we could add as a checkbox to the PR template although it's gotten to the point where I gloss over those

Yeah I agree. Unfortunately this setup doesn't "autogenerate" anything, just combines it. So aside from being near to the route definition it doesn't provide huge benefit being in a jsdoc comment.

Let's try with yaml files and see how it goes.

bryn-pins (approved):

PR #3308, opened 2023-11-29
Feature/aapd 854

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-854

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/src/controllers/token.js
```diff
@@ -31,32 +44,147 @@ async function tokenPut(req, res) {
...
+ */
 async function tokenPost(req, res) {
 	const { id, token } = req.body;
 
 	const document = await getTokenDocumentIfExists(id);
 
 	if (!document) {
-		return res.status(200).send({});
+		res.status(200).send({});
+		return;
```
ms-ek, line 116: Curious, is there an advantage of doing the return on a different line or just to make sure we don't return any values?

packages/appeals-service-api/src/db/db-client.js
```diff
@@ -1,11 +1,27 @@
 const { PrismaClient } = require('@prisma/client');
 const config = require('../configuration/config');
 
-const dbClient = new PrismaClient({
+let primsaConfig = {
```
ms-ek, line 4: small typo "prismaConfig"

ms-ek (approved):

packages/appeals-service-api/src/controllers/token.js
bryn-pins: Since adding the jsdoc for express handler it gives me red squiggles:
Type 'Response<any, Record<string, any>, number>' is not assignable to type 'void'.

I don't think it actually matters though

packages/appeals-service-api/src/routes/token-v2.spec.yaml
```diff
@@ -0,0 +1,83 @@
...
+      responses:
+        200:
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                type: object
+        404:
+          description: Appeal/User not found
+          schema:
```
beejjacobs, line 38: Sorry this is my fault as this is probably copied, this doesn't work and needs the content + type bit.

beejjacobs (approved):
Looks good! The new pattern seems to work well here, as does the transaction for the create/update of the token - neatly in a repo method seems clear.

PR #3315, opened 2023-12-04
feat(api): appeal-case data

beejjacobs:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-954

## Description of change

- Added in fields to AppealCase from the Data Model
- Created a /v2/appeal-cases endpoint
- fixed integration tests (failures weren't reported)
- updated Jest to support subpath imports
- to do: tests for this feature!

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/db/seed/util.js
+++ b/packages/appeals-service-api/src/db/seed/util.js
@@ -1,0 +1,64 @@
+const now = new Date();
+const year = now.getFullYear();
+const days = Array.from({ length: 28 }, (_, i) => i + 1);
+
+/**
+ * @param {number} year
+ * @param {number} month 0-11
+ * @param {number} day 1-28
+ * @returns {Date}
+ */
+function dateFromYMD(year, month, day) {
+	const yearStr = year.toString();
+	month++; // one-index
+	// wrap around
+	if (month < 1) {
+		month += 12;
+	} else if (month > 12) {
+		month -= 12;
+	}
+	const monthStr = month.toString().padStart(2, '0');
+	const dayStr = day.toString().padStart(2, '0');
+	return new Date(`${yearStr}-${monthStr}-${dayStr}T09:00:00Z`);
+}
+
+module.exports = {
+	now,
+	/**
+	 * Generate some dates in the previous month
+	 *
+	 * @returns {Date[]}
+	 */
+	datesLastMonth() {
+		const month = now.getMonth() - 1;
+		return days.map((day) => dateFromYMD(year, month, day));
+	},
+	/**
+	 * Generate some dates in the next month
+	 *
+	 * @returns {Date[]}
+	 */
+	datesNextMonth() {
+		const month = now.getMonth() + 1; // next month
+		return days.map((day) => dateFromYMD(year, month, day));
+	},
+	/**
+	 *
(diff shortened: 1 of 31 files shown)


packages/appeals-service-api/src/db/seed/util.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+function dateFromYMD(year, month, day) {
+	const yearStr = year.toString();
+	month++; // one-index
+	// wrap around
+	if (month < 1) {
+		month += 12;
+	} else if (month > 12) {
+		month -= 12;
+	}
```
Si3rr4wow, line 19: `Date.setMonth` expects zero index and wraps ;)

packages/appeals-service-api/src/routes/v2/appeal-cases/controller.js
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * @type {import('express').RequestHandler}
+ */
+async function getCount(req, res, next) {
+	if ('lpa-code' in req.query) {
+		countByLpaCode(req, res, next);
+		return;
+	}
+	throw ApiError.withMessage(400, 'lpa-code is required');
+}
```
Si3rr4wow, line 90: Maybe invert this so the error is thrown by a guard clause?

packages/appeals-service-api/src/routes/v2/appeal-cases/repo.js
```diff
@@ -0,0 +1,107 @@
...
+	 * @param {string} caseReference
+	 * @returns {Promise<AppealCase|null>}
+	 */
+	async getByCaseReference(caseReference) {
+		return await dbClient.appealCase.findUnique({
+			where: {
+				caseReference
+			}
+		});
+	}
```
Si3rr4wow, line 23: is async await necessary here if the goal is to return a promise?

packages/appeals-service-api/src/routes/v2/appeal-cases/controller.js
beejjacobs: I was using the same pattern as `list`, though it does look odd with just one option in there.

packages/appeals-service-api/src/routes/v2/appeal-cases/spec.yaml
```diff
@@ -0,0 +1,94 @@
+paths:
```
bryn-pins, line 1: thoughts on using tags, little annoying needing to add to main spec and here, but it looks nicer to me

bryn-pins (commented):
Apart from tests, think it looks good

I don't think we need to require tests on this PR as it was already missed, can do separately if needed

packages/appeals-service-api/src/routes/v2/appeal-cases/spec.yaml
beejjacobs: I hadn't appreciated what they did in the UI. added in.

You don't have to add to the root, but if you do you can control order in the UI which is handy.

beejjacobs:
This work is required by other tickets going forward, so going to merge as-is and add tests after.

bryn-pins (approved):

PR #3318, opened 2023-12-04
feat(api): v2 folder structure

beejjacobs:
## Ticket Number

N/A

## Description of change

Move the new /v2/users route to a new folder structure. Based off the work done in `web-comment` to load routes dynamically from the file structure. Moves controllers, services, routers, and spec into a single folder per route.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,0 +1,12 @@
+{
+	"compilerOptions": {
+		"baseUrl": ".",
+		"allowSyntheticDefaultImports": true,
+		"checkJs": true,
+		"esModuleInterop": true,
+		"moduleResolution": "node",
+		"target": "esnext",
+		"module": "CommonJS",
+		"strict": true
+	}
+}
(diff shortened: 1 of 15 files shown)


jsconfig.json
beejjacobs, line 1: So vscode enables code checking.

packages/appeals-service-api/jsconfig.json
beejjacobs, line 1: specific config to support subpath imports. 

https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-imports

packages/appeals-service-api/src/routes/v2/users/controller.js
```diff
@@ -30,6 +30,9 @@ async function userGet(req, res) {
  * @type {import('express').Handler}
  */
 async function userPost(req, res) {
+	if ('id' in req.body) {
```
beejjacobs, line 33: We don't need a user ID to be settable.

bryn-pins (approved):

PR #3326, opened 2023-12-07
feat(AAPD-855): make sql appeal when mongo appeal is made

Si3rr4wow:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/sql/appeals-repository.js
+++ b/packages/appeals-service-api/src/repositories/sql/appeals-repository.js
@@ -28,7 +28,7 @@ class AppealsRepository {
 			});
 		} catch (err) {
 			logger.error(err);
-			if (err instanceof Prisma.PrismaClientKnownRequestError) {
+			if (err instanceof Prisma.PrismaClientKnownRequestError && err.code === 'P2002') {
 				throw ApiError.appealDuplicate();
 			}
 			throw err;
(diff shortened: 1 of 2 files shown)


packages/appeals-service-api/src/repositories/sql/appeals-repository.js
```diff
@@ -0,0 +1,76 @@
...
+	 * @returns {Promise<Appeal>}
+	 */
+	async createAppeal(appeal) {
+		try {
+			return await this.dbClient.appeal.create({
+				data: appeal
+			});
+		} catch (err) {
+			logger.error(err);
+			if (err instanceof Prisma.PrismaClientKnownRequestError) {
```
beejjacobs, line 31: This check isn't enough to know that it's a duplicate (this was perhaps copied I missed it elsewhere).

See: https://www.prisma.io/docs/orm/reference/error-reference#prismaclientknownrequesterror - we can check error codes instead.

beejjacobs (commented):
Small point on errors, but otherwise looks good to me.

packages/appeals-service-api/src/repositories/sql/appeals-repository.js
Si3rr4wow: interesting 🤔 I actually inherited this code by copy and pasting the user repo so I'll have a look at the link and fix it there too

beejjacobs (approved):
Looks good!

PR #3329, opened 2023-12-07
Feature/aapd 861

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-861

## Description of change
This PR links users to appeals in sql at two separate occasions
When doing enter code to confirm an email for an appeal, it sets user as an appellant
When setting isOriginalApplicant on the appeal update either agent/appellant role

There may probably need to be an addition to this in future to for users creating a new appeal when already logged in

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/sql/appeals-repository.js
+++ b/packages/appeals-service-api/src/repositories/sql/appeals-repository.js
@@ -1,0 +1,90 @@
+const { createPrismaClient } = require('../../db/db-client');
+const { Prisma } = require('@prisma/client');
+const logger = require('../../lib/logger');
+const ApiError = require('../../errors/apiError');
+
+/**
+ * @typedef { import("@prisma/client").Appeal } Appeal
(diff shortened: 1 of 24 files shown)


packages/forms-web-app/src/controllers/common/enter-code.js
bryn-pins, line 1: If we use a new controller going forward will have to included this logic there

packages/appeals-service-api/src/repositories/sql/appeals-repository.js
```diff
@@ -0,0 +1,90 @@
...
+		});
+	}
+
+	/**
+	 * Get by legacy appeal id
+	 *
+	 * @param {string} id
+	 * @returns {Promise<Appeal|null>}
+	 */
+	async getByLegacyId(id) {
```
bryn-pins, line 81: This is the only addition, rest copied from another PR

PR #3330, opened 2023-12-07
feat(aapd-728): enrol users with no appeals associated

ms-ek:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-728

## Description of change

enrol users with no appeals associated

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/lib/is-token-valid.js
```diff
@@ -8,15 +8,20 @@ const testConfirmEmailToken = '12345';
 const isTestEnvironment = () => config.server.allowTestingOverrides;
 
 /**
- * Check if LPA is System Test Borough Council and token is test token
```
bryn-pins, line 11: should this still be here

packages/appeals-service-api/src/repositories/appeals-repository.js
```diff
@@ -15,6 +15,19 @@ class AppealsRepository extends MongoRepository {
 		return await this.findOneByQuery({ _id: id });
 	}
 
+	/**
+	 *
+	 * @param {any} email
+	 * @return {Promise<any>}
+	 */
+	async getByEmail(email) {
```
bryn-pins, line 23: Not sure on this, will message in teams

packages/appeals-service-api/src/controllers/appeals.js
```diff
@@ -0,0 +1,27 @@
+const { getAppealsByEmail } = require('../services/appeal.service');
+const logger = require('../lib/logger');
+const ApiError = require('#errors/apiError');
+
+const getUserAppealsByEmail = async (req, res) => {
```
bryn-pins, line 5: think this should probably be a new v2 endpoint, that checks for linked appeals in sql rather than mongo

bryn-pins (commented):
Some initial thoughts

packages/appeals-service-api/src/repositories/appeals-repository.js
```diff
@@ -15,6 +15,19 @@ class AppealsRepository extends MongoRepository {
...
+		const result = await this.getAllDocumentsThatMatchQuery({
+			'appeal.email': email,
+			'appeal.lpaCode': { $exists: true } // only return if an actual appeal has been started
```
bryn-pins, line 26: is appeal.email not enough here, isn't it always set after lpaCode?

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -173,8 +182,14 @@ const postEnterCode = (views) => {
 			});
 		}
 
+		const isReturningUser = req.session.appeal.newOrSavedAppeal === 'return';
```
bryn-pins, line 185: could we create a separate file (appeal-user-service.js or something along those lines) that handles appeal user based session values, being able to see in one place would help understanding in future of what can be set on a session

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -208,6 +223,18 @@ const postEnterCode = (views) => {
 		// if handling an email confirmation
 		// session will be in browser so can redirect here and consider email confirmed
 		if (tokenValid.action === enterCodeConfig.actions.confirmEmail) {
+			if (isReturningUser) {
+				const user = await getUserByEmailV2(req.session.appeal.email);
+				if (!user.isEnrolled) {
```
bryn-pins, line 228: is this required? I added an auto enrol to the token endpoint in api?

packages/appeals-service-api/src/repositories/appeals-repository.js
ms-ek: yeah on the return flow it doesn't have it

packages/appeals-service-api/src/controllers/appeals.js
ms-ek, line 5: This is to get all the appeals for a given user email. I used it to check a user has no associated appeals before enrolling them.

packages/appeals-service-api/src/repositories/appeals-repository.js
bryn-pins: I guess my question would be is it possible to avoid creating an appeal in session if they've clicked

packages/appeals-service-api/src/routes/v2/users/controller.js
```diff
@@ -53,7 +53,32 @@ async function userPost(req, res) {
 	}
 }
 
+/**
+ * @type {import('express').Handler}
+ */
+async function userPut(req, res) {
```
ms-ek, line 59: This is to update a user in the SQL database

packages/forms-web-app/src/controllers/appeal/view-my-appeals.js
```diff
@@ -0,0 +1,13 @@
+const { VIEW } = require('../../lib/views');
+
+exports.get = async (req, res) => {
+	const returningUserWithNoAppeals = req.session.returningUserWithNoAppeals;
```
ms-ek, line 4: This is a placeholder page for the 'view my appeals page'

packages/forms-web-app/src/controllers/common/enter-code.js
ms-ek: Good idea, will add that. So something like:

exports.setNewOrSavedAppeal = (session, value) => {
 session.appeal.newOrSavedAppeal === value;
};

packages/forms-web-app/src/controllers/common/enter-code.js
ms-ek: Oh okay, that makes sense, that didn't work locally because we accept the 12345 token without calling the check token endpoint

packages/appeals-service-api/src/repositories/appeals-repository.js
ms-ek: Oh okay I see what you mean, probably possible, I'll try it

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -208,6 +220,15 @@ const postEnterCode = (views) => {
 		// if handling an email confirmation
 		// session will be in browser so can redirect here and consider email confirmed
 		if (tokenValid.action === enterCodeConfig.actions.confirmEmail) {
+			if (isReturningUser) {
+				const userEmail = req.session.appeal.email;
+				const associatedAppeals = await getAppealsByEmail(userEmail);
```
ms-ek, line 225: Need to check in SQL rather than MongoDB

beejjacobs: You can use the new endpoint:

`/api/v2/users/{user-id}/appeals`

packages/forms-web-app/src/views/appeal/enter-code.njk
beejjacobs, line 1: Do we need these nunjucks files, can't we just render the common file?

packages/forms-web-app/src/lib/appeals-api-wrapper.js
```diff
@@ -97,6 +97,10 @@ exports.getExistingAppeal = async (appealId) => {
 	return handler(`/api/v1/appeals/${appealId}`);
 };
 
+exports.getAppealsByEmail = async (email) => {
+	return handler(`/api/v1/appeals/email/${email}`, 'GET');
+};
+
```
beejjacobs, line 103: We don't need this any more do we? Let's remove it and the backend code. It could be a very expensive query so we don't want to call it accidentally!

packages/forms-web-app/src/lib/appeals-api-wrapper.js
```diff
@@ -268,8 +229,66 @@ exports.errorMessages = {
 	}
 };
 
+/**
+ * V2 endpoints
+ */
+
 exports.getUserAppealsById = async (id) => {
 	return handler(`/api/v2/users/${id}/appeals`, 'GET');
+}
```
beejjacobs, line 238: We should move this to the new `appeals-api-client` in common - same with `updateUser`.

packages/appeals-service-api/src/routes/v2/token/controller.js
```diff
@@ -65,6 +65,7 @@ async function tokenPostV2(req, res) {
 	// user has confirmed email for first time so set enrolled flag
 	if (!user.isEnrolled) {
 		user.isEnrolled = true;
+		sendConfirmRegistrationEmailToAppellant(user.email, user.id, user.email);
```
beejjacobs, line 68: We only want to send this is `enrol-users` flag is active - is that checked somewhere?

Also I think you're missing an `await` here.

beejjacobs (commented):
Looks good - quite a few changes to make this work!

Can we remove the only list-appeals-by-email v1 endpoint please?

Also a few other comments to check over.

packages/forms-web-app/src/controllers/appeals/your-appeals.js
```diff
@@ -1,17 +1,23 @@
-const { getUserAppealsById, getUserByEmail } = require('../../lib/appeals-api-wrapper');
 const { mapToAppellantDashboardDisplayData } = require('../../lib/dashboard-functions');
 const { VIEW } = require('../../lib/views');
 const logger = require('../../lib/logger');
+const { AppealsApiClient } = require('@pins/common/src/client/appeals-api-client');
+const config = require('../../config');
 
 exports.get = async (req, res) => {
-	const email = req.session.appeal?.email;
+	const apiClient = new AppealsApiClient(config.appeals.url, config.appeals.timeout);
```
beejjacobs, line 8: You can import this from `src/lib/appeals-api-client.js` which contains a shared instance of AppealsApiClient

packages/common/src/client/appeals-api-client.js
```diff
@@ -62,7 +62,16 @@ class AppealsApiClient {
 		const endpoint = urlParams.toString()
 			? '/api/v2/appeal-cases?' + urlParams.toString()
 			: '/api/v2/appeal-cases';
+		const response = await this.#makeGetRequest(endpoint);
+		return response.json();
+	}
 
+	/**
+	 * @param {string} id
+	 * @returns {Promise<import('appeals-service-api').Api.AppealUser>}
```
beejjacobs, line 71: I think this is actually `(AppealCase|AppealSubmission)[]` ?

With
```
/**
 * @typedef {import('appeals-service-api').Api.AppealCase} AppealCase
 * @typedef {import('appeals-service-api').Api.AppealSubmission} AppealSubmission
 */
 ```

beejjacobs (commented):
Two more minor ones!

bryn-pins (approved):

PR #3334, opened 2023-12-12
feat(aapd-837): add dynamic routing and tables with test data

Maby0:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-

## Description of change

Add placeholder index files for all routes with placeholder data in the tables

## Checklist

- [ ] Requires infrastructure changes
- [ ] If adding or remove environment variables (e.g. in `docker-compose.yaml`) then I have updated the appropriate Helm chart
- [ ] I have updated the documentation accordingly
- [ ] My commit history in this PR is linear
- [ ] New features have tests
- [ ] Breaking change (team conversation required)

## Important

Please do not merge from `main` (please only [rebase](https://github.com/foundry4/appeal-planning-decision/wiki/An-intro-to-Git-Rebase)). This keeps the history linear and easier to debug.

PR #3351, opened 2023-12-18
feat(aapd-861): link sqluser to appeal

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-861

## Description of change

Link user and appeal when feature flag is on and user confirms email

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/common/src/client/appeals-api-client.js
+++ b/packages/common/src/client/appeals-api-client.js
@@ -1,38 +1,20 @@
 const { default: fetch, AbortError } = require('node-fetch');
 const uuid = require('uuid');
+const AppealsApiError = require('./appeals-api-error');
 
 const config = require('../config');
-const parentLogger = require('./logger');
+const parentLogger = require('../lib/logger');
 
 const v2 = '/api/v2';
 
 /**
  * @class Api Client for v2 urls in appeals-service-api
  */
-class AppealsApiError extends Error {
-	/**
-	 * @param {string} message
-	 * @param {string|number} code
-	 * @param {Array.<string>} [errors]
-	 */
-	constructor(message, code, errors) {
-		super(message);
-		this.name = 'AppealsApiError';
-		/** @type {string|number} */
-		this.code = code;
-		/** @type {Array.<string>|undefined} */
-		this.errors = errors;
-	}
-}
-
-/**
- * @class Api Client for v2 urls in appeals-service-api
- */
 class AppealsApiClient {
 	/**
-	 * @param {string|undefined} baseUrl - defaults to config.appeals.url
+	 * @param {string|undefined} baseUrl - defaults to config.apis.appealsApi.url
 	 */
-	constructor(baseUrl = config.appeals.url) {
+	constructor(baseUrl = config.apis.appealsApi.url) {
 		if (!baseUrl) {
 			throw new Error('baseUrl is required');
 		}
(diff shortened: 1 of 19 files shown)


packages/forms-web-app/src/lib/appeals-api-wrapper.js
```diff
@@ -276,6 +277,38 @@ exports.getUserAppealsById = async (id) => {
...
+ * @typedef {import('appeals-service-api').Api.AppealCaseWithAppellant} AppealCaseWithAppellant
+ */
+
+/**
+ * @param {string} email
+ * @param {string} appealSqlId
+ * @param {string} [role]
+ * @returns {Promise<AppealCaseWithAppellant>}
+ */
+exports.linkUserToV2Appeal = async (email, appealSqlId, role) => {
```
beejjacobs, line 290: Thoughts on a class-based approach for v2 API calls?

e.g. https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/3334/files#diff-9727715fc7c722a6ac0258c1eddf1a7e9445e6bad6047711d142f81066bc8b6c

beejjacobs (commented):
Looks good and thorough to me!

My only real question is: will this update the role if it changes during the journey? I thought we might need something on other parts of the journey or to check appeal submission update data?

bryn-pins:
> Looks good and thorough to me!
> 
> My only real question is: will this update the role if it changes during the journey? I thought we might need something on other parts of the journey or to check appeal submission update data?

No, will do that in part two

packages/forms-web-app/src/lib/appeals-api-wrapper.js
bryn-pins: Definitely preferable, want me to wait for that?

bryn-pins: I'll bring a copy in

beejjacobs (approved):

packages/common/src/client/appeals-api-client.js
```diff
@@ -0,0 +1,172 @@
+const { default: fetch, AbortError } = require('node-fetch');
+const uuid = require('uuid');
```
beejjacobs, line 2: don't need a package for this, we can use `crypto.randomUUID()`

packages/common/src/client/appeals-api-client.js
```diff
@@ -0,0 +1,172 @@
...
+		const endpoint = `${v2}/users/${email}`;
+		const response = await this.#makeGetRequest(endpoint);
+		return response.json();
+	}
+
+	/**
+	 * Handles error responses and timeouts from calls to appeals api
+	 * @param {string} path endpoint to call e.g. /api/v2/users
+	 * @param {'GET'|'POST'|'PUT'|'DELETE'} [method] - request method, defaults to 'GET'
+	 * @param {object|undefined} [opts] - options to pass to fetch can include request body
```
beejjacobs, line 62: FYI `[]` (optional) implies `|undefined`

packages/common/src/config.js
beejjacobs, line 1: Appreciate it's already a pattern but I don't like using `process.env` in packages/libraries. Better to explicitly pass in any parameters that are required - which I think the ApiClient supports for `baseUrl`, we should pass in `timeout` too.

beejjacobs (commented):
api client looks great! Couple of comments.

beejjacobs (approved):

PR #3365, opened 2023-12-22
chore(local-dev): setup docker profiles and mock services

bryn-pins:
## Description of change

**Profiles:**
Setup 3 docker profiles and a make script to run them
- all
- appeals (forms web app and dependencies)
- comment  (web comment and dependencies, guessed with these having not worked on it)

**Service Mocks**
Mock clam av and pdf service to have slimmer option for running locally
Use these mocks with docker-compose.slim.yml

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,25 @@ run:
 		${CMD}
 .PHONY: run
 
+# all services
 serve:
-	docker-compose up
+	docker compose --profile all -f docker-compose.yml -f docker-compose.full.yml up
 .PHONY: serve
+
+# all services slimmed down by mocking some more services
+slim:
+	docker compose --profile all -f docker-compose.yml -f docker-compose.slim.yml up
+.PHONY: slim
+
+# only services required for web-comment
+web-comment:
+	docker compose --profile comment -f docker-compose.yml -f docker-compose.slim.yml up
+.PHONY: web-comment
+
+# only services required for appeals/forms-web-app
+appeals:
+	docker compose --profile appeals -f docker-compose.yml -f docker-compose.slim.yml up
+.PHONY: appeals
 
 uninstall:
 	rm -Rf node_modules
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -19,10 +19,26 @@ run:
...
+# only services required for web-comment
+web-comment:
+	docker compose --profile comment -f docker-compose.yml -f docker-compose.slim.yml up
+.PHONY: web-comment
+
+# only services required for appeals/forms-web-app
+appeals:
+	docker compose --profile appeals -f docker-compose.yml -f docker-compose.slim.yml up
+.PHONY: appeals
+
```
beejjacobs, line 41: Should we make these npm scripts instead/as well?

Makefile
```diff
@@ -19,10 +19,26 @@ run:
 		${CMD}
 .PHONY: run
 
+# all services
 serve:
-	docker-compose up
+	docker compose --profile all -f docker-compose.yml -f docker-compose.full.yml up
```
beejjacobs, line 24: Does the order matter? do we need slim/full first, then the main one?

beejjacobs (commented):
I like it! Couple of questions.

Makefile
bryn-pins: have added and documented in readme

Makefile
bryn-pins: my understanding is they're combined into one file when ran so order does matter due to overrides but I think this is the correct way round and currently doesn't matter?
https://docs.docker.com/engine/reference/commandline/compose/#specifying-multiple-compose-files

beejjacobs: I was thinking more with the order things start in, but if this works that's all good.

beejjacobs (approved):
Looks good!

PR #3379, opened 2024-01-04
feat(aapd-807): add appeals migration endpoint

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-807

## Description of change

Added an endpoint to migrate existing appeals/users from mongo to sql
Have left all db logic in controller as expecting it to be a one off, but could move to repo functions

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/src/routes/v2/migrate/controller.js
```diff
@@ -0,0 +1,197 @@
...
+				legacyAppealSubmissionDecisionDate: doc.appeal.decisionDate,
+				legacyAppealSubmissionState: doc.appeal.state
+			}
+		});
+
+		await sqlClient.appealToUser.create({
+			data: {
+				appealId: appeal.id,
+				userId: userId,
+				role: 'appellant'
```
beejjacobs, line 179: Do we not need to set this from the data? could be agent?

beejjacobs (commented):
Looks pretty good, one minor comment.

I assume it works locally? I didn't try it!

My only other thought is whether it's useful to have the positives as well as errors: e.g. do we return all appeal ids processed rather than just counts?

packages/appeals-service-api/src/routes/v2/migrate/controller.js
bryn-pins: ah yes, will look into that

bryn-pins: Have resolved that and wrapped in transaction
Have also added a few other checks/options a separate commit

beejjacobs (approved):

PR #3387, opened 2024-01-09
feat(aapd-863): save and return functions

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-863

## Description of change

This PR adds
- Ensures all journeys pre/post enrol users flag are functioning
- Refactors enter code logic for clarity
- update document links to handle login, different route for lpas
- discovered some missing/broken routes so have fixed those

- [x] New features have tests

## Important


packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -107,81 +101,112 @@ async function sendTokenToLpaUser(req) {
 	}
 }
 
-const getEnterCode = (views, appealInSession) => {
+/**
+ * @param {{EMAIL_ADDRESS: string, ENTER_CODE: string, REQUEST_NEW_CODE: string}} views
+ * @param {boolean} [isGeneralLogin]
```
beejjacobs, line 106: Could we explain the meaning of this a little? Also wondering if an options object with that as a property would be clearer in usage: `getEnterCode(views, {isGeneralLogin: true})` ?

packages/forms-web-app/src/routes/appeal-householder-decision/index.js
```diff
@@ -4,14 +4,17 @@ const router = express.Router();
...
+} = require('../../controllers/full-appeal/submit-appeal/appeal-already-submitted');
 
-// pre login
-router.use(checkAppealExists, checkDecisionDateDeadline, require('./planning-application-number'));
 router.use(require('./cannot-appeal'));
-
-// login
 router.use(require('./login'));
 
+router.get('/appeal-already-submitted', [fetchExistingAppealMiddleware], getAppealAlreadySubmitted);
```
beejjacobs, line 15: Is this a new route?

packages/forms-web-app/src/middleware/check-logged-in.js
```diff
@@ -21,17 +28,54 @@ const isLoggedIn = async (req, res, next) => {
 		return next();
 	}
 
-	const pathRequiresCustomRedirect = requiresCustomRedirect(req.originalUrl);
+	if (requiresCustomRedirect(req.originalUrl)) {
+		req.session.loginRedirect = req.originalUrl; // will redirect after enter code
 
-	req.session.newOrSavedAppeal = NEW_OR_SAVED_APPEAL_OPTION.RETURN;
+		const appealId = req.params?.appealOrQuestionnaireId;
```
beejjacobs, line 34: Can we note which scenario this is handling?

packages/forms-web-app/src/lib/is-token-valid.js
```diff
@@ -73,9 +73,17 @@ const getToken = async (id, token, emailAddress, session) => {
  * @param {string} token
  * @param {string} [emailAddress]
  * @param {any} [session] Express request session data
+ * @param {boolean} [isTestScenario] is test scenario
  * @returns {Promise<TokenValidResult>}
  */
-const isTokenValid = async (id, token, emailAddress, session) => {
+const isTokenValid = async (id, token, emailAddress, session, isTestScenario) => {
+	if (isTestScenario) {
```
beejjacobs, line 80: I don't love this code path. Do we need some config for production? i.e. so we don't hit this code path if configured for production.

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -107,81 +101,112 @@ async function sendTokenToLpaUser(req) {
...
 			body: { errors = {} }
 		} = req;
 
-		const action = req.session?.enterCode?.action ?? enterCodeConfig.actions.saveAndReturn;
-
-		if (action === enterCodeConfig.actions.saveAndReturn) {
-			req.session.enterCode = req.session.enterCode || {};
-			req.session.enterCode.action = enterCodeConfig.actions.saveAndReturn;
+		/** @type {string|undefined} */
+		const enterCodeId = req.params.id;
```
beejjacobs, line 116: Should we make this param name `enterCodeId` for consistency?

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -107,81 +101,112 @@ async function sendTokenToLpaUser(req) {
...
-		//this handles old save & return url (without id params)
-		if (appealInSession && req.session.appeal?.id && !req.params.id) {
-			res.redirect(`/${views.ENTER_CODE_URL}/${req.session.appeal.id}`);
-			return;
+			const email = getSessionEmail(req.session, false);
+
+			try {
+				await sendToken(undefined, action, email);
+			} catch (e) {
+				logger.error(e, 'failed to send token to general login user');
```
beejjacobs, line 141: Should this not show an error to the user?

beejjacobs (commented):
Generally looks good. Some duplication with the other PR but I've just ignored those bits.

I haven't tested it, but the logic appears sound. We'll do extensive testing when this is all merged. Some comments/thoughts to look at.

packages/forms-web-app/src/lib/is-token-valid.js
bryn-pins: So was a refactor of the existing logic, allowing 12345 for test scenarios, was previously all in entercode and there's an env variable set to false for prod.

I'll look at moving the rest of the logic into here as well, which definitely makes more sense

packages/forms-web-app/src/routes/appeal-householder-decision/index.js
bryn-pins: yes I hit a 404 on this path for householder while testing, so just added the same page from full planning
Could take out of this PR and make a separate ticket

beejjacobs: I'm OK with it being in here.

packages/forms-web-app/src/controllers/common/enter-code.js
beejjacobs: Yes I see. I guess I was thinking about "we failed to send the email" type errors - which perhaps we handle differently so they can retry? Possibly for another ticket/PR.

beejjacobs (approved):
LGTM

PR #3388, opened 2024-01-10
fix(aapd-724): fix routes

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-724

## Description of change

- removed an unused route
- added a missing return to the login middleware, not sure how I missed that
- where possible simplified the middleware logic to rely on the order of definition rather than lists of urls

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/__tests__/unit/middleware/check-appeal-exists.test.js
```diff
@@ -23,51 +23,14 @@ describe('middleware/check-appeal-exists', () => {
...
-	});
-
-	it('should call next() for the `/before-you-start/use-a-different-service` page', () => {
-		req.originalUrl = '/before-you-start/use-a-different-service';
-		checkAppealExists(req, res, next);
-		expect(next).toBeCalled();
-		expect(res.redirect).not.toBeCalled();
-	});
-
-	it('should call next() for the `/appellant-submission/submission-information/6e1195ad-176d-4ca3-a944-525218780a7e` page', () => {
```
bryn-pins, line 47: slightly worried about these two submission-information routes but as far as I can tell they are called when submitting an appeal so the appeal will be in session

packages/forms-web-app/src/middleware/check-decision-date-deadline.js
```diff
@@ -10,16 +10,7 @@ const {
 		APPELLANT_SUBMISSION: { CANNOT_APPEAL: cannotAppealHAS }
 	}
 } = require('../lib/views');
-const validationExclusionPages = [
```
bryn-pins, line 13: felt safe to just rely on /before-you-start to simplify this
user will be redirected to a cannot appeal page if they enter a date beyond the deadline in decision-date
So think the only way this would impact someone is if they changed the appeal type after setting date, and typed the urls rather than followed the journey, at which point they would get told after completing before you start

packages/forms-web-app/src/routes/full-appeal/index.js
```diff
@@ -13,6 +13,11 @@ router.use(require('./login'));
 /// as well as appeal objects
 router.use('/submit-final-comment', require('./submit-final-comment'));
 
+router.use(
+	checkAppealExists,
+	checkDecisionDateDeadline,
+	require('./submit-appeal/planning-application-number')
+);
```
beejjacobs, line 20: I take it this one doesn't need to be logged in to access?

packages/forms-web-app/src/middleware/check-appeal-exists.js
beejjacobs, line 1: Nice to remove this list of URLs - felt strange this was in "checkAppealExists" too!

packages/forms-web-app/src/lib/views.js
```diff
@@ -36,7 +36,6 @@ const VIEW = {
 	APPELLANT_SUBMISSION: {
 		TASK_LIST: 'appeal-householder-decision/task-list',
 		APPEAL_STATEMENT: 'appellant-submission/appeal-statement',
-		APPLICATION_NUMBER: 'appellant-submission/application-number',
```
beejjacobs, line 39: Sorry all the route changes are hard to follow - why is this not needed? Just unused?

beejjacobs (approved):
The changes look good to me, I like the approach. Can't pretend to follow it all exactly though!

We probably need to make sure there's some decent test criteria on this ticket.

packages/forms-web-app/src/routes/full-appeal/index.js
bryn-pins: Yes it goes:
- before you start
- planning application number
- email

PR #3392, opened 2024-01-12
feat(aapd-654): increase max file size

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-654

## Checklist

- [x] New features have tests

## Important


beejjacobs (dismissed):

packages/common/src/client/clamav-rest-client.js
```diff
@@ -0,0 +1,53 @@
...
+	 * @param {number} [maxBodyLength] - maxBodySize in bytes used by axios
+	 */
+	constructor(host, maxBodyLength) {
+		if (!host) {
+			throw new Error('host is required');
+		}
+
+		/** @type {string} */
+		this.host = host.replace(trailingSlashRegex, '');
+		this.maxBodyLength = maxBodyLength ? maxBodyLength * 1.1 : maxBodyLength; // add headroom
```
bryn-pins, line 19: something was larger in the forwarded request than the incoming one so added some headroom, feels hacky if someone can spot what would likely be increasing the request size could make more specific

packages/common/src/client/clamav-rest-client.js
```diff
@@ -0,0 +1,53 @@
...
+
+	/**
+	 * sends file to clamav rest server
+	 * @param {import('express-fileupload').UploadedFile} fileInformation
+	 * @param {string} fileName
+	 * @returns {Promise<boolean>}
+	 */
+	async scan(fileInformation, fileName) {
+		if (typeof fileInformation?.tempFilePath !== 'undefined') {
+			const readableStream = fs.createReadStream(fileInformation?.tempFilePath);
```
bryn-pins, line 30: have modified this from believe this should be streams all the way down now?

    const fileBuffer = fs.readFileSync(fileInformation?.tempFilePath);
    const readableStream = Readable.from(fileBuffer.toString());

packages/clamav-rest-server/src/lib/clamd.js
```diff
@@ -5,30 +5,31 @@ const config = require('./config');
 const logger = require('./logger');
 
 const sendFile = async (file) => {
-	console.log(file);
```
bryn-pins, line 8: this seemed problematic

packages/forms-web-app/src/validators/common/schemas/file-upload.js
```diff
@@ -46,7 +47,8 @@ const schema = (noFilesError) => ({
 
 				//check file for virus
 				const { name } = req.files[path];
-				await validAV(req.files['file-upload'], name);
+				const clamAVClient = new ClamAVClient(clamAVHost, uploadApplicationMaxFileSize);
```
Si3rr4wow, line 50: Could this be constructed outside the schema function?

Si3rr4wow (dismissed):

packages/forms-web-app/src/validators/common/schemas/file-upload.js
bryn-pins: yes, I guess the max request size should probably move to a param for scan as well, will do that now

bryn-pins: So I tried this but it led to 30+ unrelated test suites breaking which required either a load of mocking or a workaround, so have undone this change
might look into tests another time
I don't think creating this class on each request is that inefficient, or are we potentially losing connection pooling somewhere along the line by not reusing the same instance?

beejjacobs: Given we just call axios in the class function, I don't think we miss much. 

We could have a function that returns the singleton/instance, and only call that where it's needed - shouldn't break other tests then? May not be worth it!

bryn-pins: Have done the require within a function call

beejjacobs (approved):

PR #3401, opened 2024-01-18
feat(dev): add script to update ticket statuses

beejjacobs:
## Ticket Number

N/A

## Description of change

Script to help updating ticket statuses after release

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/dev/release-util/src/update-released-issues.js
+++ b/dev/release-util/src/update-released-issues.js
@@ -1,0 +1,65 @@
+const dotenv = require('dotenv');
+const { docopt } = require('docopt');
+const { JiraApi } = require('./jira-api');
+
+dotenv.config(); // load from .env
+const username = process.env.JIRA_USERNAME;
+const apiKey = process.env.JIRA_API_KEY;
+const baseUrl = process.env.JIRA_BASE_URL;
+
+const docs = `
+Usage:
+    update-released-issues --release-link=<release-link> --release-name=<release-name> <issues>...
+
+        <issues> - issue ids to update
+`;
+
+/**
+ * a cli tool to aid updating ticket statuses to released
+ */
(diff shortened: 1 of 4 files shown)


dev/release-util/src/update-released-issues.js
```diff
@@ -0,0 +1,47 @@
...
+			console.log('no transition for', issue);
+			notTransitioned.push(issue);
+			continue;
+		}
+		await jira.transitionIssue(
+			issue,
+			released.id,
+			`Released as part of [${releaseName}|${releaseLink}]`
+		);
+	}
```
Si3rr4wow, line 43:
```
const transitionIssue = async (issue) => {
	const transitions = await jira.fetchIssueTransitions(issue);
	const released = transitions.find((t) => t.name === 'Released');
	if (!released) {
		console.log('no transition for', issue);
		notTransitioned.push(issue);
		return Promise.resolve();
	}
	return await jira.transitionIssue(
		issue,
		released.id,
		`Released as part of [${releaseName}|${releaseLink}]`
	);
}

await Promise.allSettled(issues.map(transitionIssue))
```

Not super important in this case because there won't be that many issues but generally it's faster to spool up all the promises synchronously then await them all. That way it takes as long as the slowest promise rather than the sum of all promises

Si3rr4wow (dismissed):

dev/release-util/src/update-released-issues.js
Si3rr4wow: Quick scan of their docs, the rate limit is... confusing 😵‍💫 Looks like they pass back a retry-after header on a 429 though so that's nice. probably best to stick with crude and functional

Si3rr4wow (approved):

PR #3417, opened 2024-01-26
Aapd 706/slow uploads

Si3rr4wow:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/package.json
```diff
@@ -49,6 +49,7 @@
 		"jsonpath": "^1.1.1",
 		"jspdf": "^2.5.1",
 		"lusca": "^1.6.1",
+		"multer": "^1.4.5-lts.1",
```
Si3rr4wow, line 52: remove

packages/forms-web-app/src/app.js
```diff
@@ -32,6 +32,11 @@ const routes = require('./routes');
 
 const app = express();
 
+app.post('/*', (req, _res, next) => {
+	console.log('😭😭😭😭😭😭 Start request');
+	next();
+});
+
```
Si3rr4wow, line 39: lool  remove

packages/forms-web-app/src/app.js
```diff
@@ -101,7 +106,17 @@ app.use(
...
+app.use((req, res, next) => {
+	console.log('before fileUpload: ', new Date());
+	next();
+});
+app.use(fileUpload({ ...config.fileUpload /*useTempFiles: true*/ }));
+app.use((req, res, next) => {
+	console.log('after fileUpload: ', new Date());
+	next();
+});
```
Si3rr4wow, line 118: remove

packages/forms-web-app/src/controllers/common/appeal-multi-file-upload.js
```diff
@@ -36,14 +37,16 @@ const postAppealMultiFileUpload = (
 			if ('files' in body && 'file-upload' in body.files) {
 				const validFiles = getValidFiles(errors, body.files['file-upload']);
 
-				// eslint-disable-next-line no-restricted-syntax
-				for await (const file of validFiles) {
-					const document = await createDocument(appeal, file, file.name, documentType, sectionTag);
+				const resolutions = await conjoinedPromises(validFiles, createDocument, {
+					asyncDepMapPredicate: (file) => [appeal, file, file.name, documentType, sectionTag],
+					applyMode: true
+				});
```
beejjacobs, line 43: If I follow correctly, `conjoinedPromises` waits for each promise in turn - so this isn't much better than a `for await` loop. To me this would be easier to read as:

```
const resolutions = await conjoinedPromises(validFiles, (file) => createDocument(appeal, file, file.name, documentType, sectionTag));
```

Do we need the apply/call + predicate options?

Si3rr4wow: It's should be as performant as promise all because it spawns the promises synchronously, so that for loop only gets held up as long as the slowest promise. I'll add a test that demonstrates this.

The apply vs call thing was a way to get it to pass multiple arguments to the promise. The map predicate is so the resulting `Map`'s keys are the object you want to work with, rather than the arguments of the promise. You could do that mapping up front but then you'd have a map like [[appeal, file, file.name, documentType, sectionTag], createDocumentResult],...], rather than [[file, createDocumentResult],...]

Si3rr4wow: ooooh I see what you mean on the second point now, yeah I like that and it would mean the options wouldn't be needed. I'll make that change

beejjacobs: Also, understood on the promise part, I realise now it is starting all the promises up front.

Si3rr4wow: done this rework, much cleaner, thanks for the suggestion :)

beejjacobs (approved):
Looks good, I find it much easier to read with the new signature of `conjoinedPromises`, thanks!

PR #3431, opened 2024-02-13
Feature/aapd 1195 auth

bryn-pins:
## Ticket Number

## Description of change

Done:

- Moved prisma logic to standalone database package to allow it to be used across multiple apps
- set up auth server
- added custom ropc grant type to auth server
- handle user/token lookup in auth server
- add auth check to BO docs api

Todo:

- figure out access from business logic rules (who can access which doctype and when)
- logging
- tests

## Checklist

- [ ] New features have tests

## Important


docker-compose.yml
```diff
@@ -256,8 +309,9 @@ services:
     #       memory: 250M
 
   mssql:
-    profiles: ['all', 'api', 'appeals', 'comment']
+    profiles: ['all', 'api', 'appeals', 'auth', 'comment']
     image: mcr.microsoft.com/azure-sql-edge:latest
+    init: true
```
bryn-pins, line 314: this seems to make stopping the server much faster, not sure what other impact this has, but seems like a nice QOL improvement if nothing else

PR #3436, opened 2024-02-15
feat(aapd-945): Appelant and LPA appeal pages

Si3rr4wow:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/selected-appeal/lpa-user-sections.js
+++ b/packages/forms-web-app/src/controllers/selected-appeal/lpa-user-sections.js
@@ -1,0 +1,91 @@
+/**
+ * @typedef {import('appeals-service-api').Api.AppealCaseWithAppellant} AppealCaseWithAppellant
+ */
+
+/**
+ * @type {import('./section').Section}
+ */
+
+exports.sections = [
+	{
+		heading: 'Appeal details',
+		links: [
+			{
+				url: 'anything',
+				text: 'View appeal details',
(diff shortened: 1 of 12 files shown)


packages/forms-web-app/src/controllers/selected-appeal/section.d.ts
```diff
@@ -0,0 +1,10 @@
+import { AppealCaseWithAppellant } from '@pins/common';
```
topickering, line 1: Do we want to import from here or the typedef way from appeals-service-api?  Not sure if it makes a difference, this current way is probably neater so feel free to ignore

Si3rr4wow: Good point! It might not actually be working importing like this but I thought I might get away with it since it's a ts file. I'll try to make sure it's actually getting this definition into Sections

Si3rr4wow: this import was _not_ working but it is now

packages/forms-web-app/src/controllers/selected-appeal/lpa-user-sections.js
beejjacobs, line 1: Not sure why the indentation is different to the appellant one?

beejjacobs (approved):
Part of me wishes we could share the config between appellant and LPA, but I guess we end up with more complexity because of the title changes.

Generally looks good! One minor question...

PR #3439, opened 2024-02-16
fix(aapd-1046): displays text in the other field for designated sites…

lauren-PINS:
… question

## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1046

## Description of change

Display the text from the 'other' field for the designated fields question and other checkbox questions from the same dynamic template.

## Checklist

- [ ] New features have tests

## Important


beejjacobs (commented):
Looks like there are some `formatAnswerForSummary` tests in `question.test.js`, can we add some tests to handle this new case?

It's not obvious to me what the code is actually doing; perhaps a few comments and test cases might help?

Si3rr4wow:
I've been having a look at this but my brain is a bit fried. I'd be happy to pair on this to try to refactor and add some tests if you think that would be helpful :)

PR #3454, opened 2024-02-22
feat(aapd-1033): BO mapping

Si3rr4wow:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/questionnaire-submission.d.ts
+++ b/packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/questionnaire-submission.d.ts
@@ -12,5 +12,3 @@ export type LPAQuestionnaireSubmission = Prisma.LPAQuestionnaireSubmissionGetPay
 		SubmissionNeighbourAddress: true;
 	};
 }>;
-
-type thing = LPAQuestionnaireSubmission[''];


packages/appeals-service-api/src/services/appeal.service.js
```diff
@@ -57,6 +61,10 @@ async function createAppeal(req, res) {
 	res.status(500).send(appeal);
 }
 
+/**
+ * @param {string} id
+ * @returns {Promise<AppealSubmission['appeal']>}
```
Si3rr4wow, line 66: I _think_ this is correct

packages/appeals-service-api/src/services/back-office-v2/formatters/has/questionnaire.js
```diff
@@ -0,0 +1,121 @@
+const { initContainerClient } = require('@pins/common');
+const { conjoinedPromises } = require('@pins/common/src/utils');
+const { blobMetaGetter } = require('../../../../services/object-store');
+
+const getBlobMeta = blobMetaGetter(initContainerClient);
+
+/**
+ * @param {*} questionnaireResponse
+ * @returns {Promise<*>}
+ */
```
Si3rr4wow, line 10: I want to tighten up the `*` use in this

packages/appeals-service-api/src/services/back-office-v2/formatters/index.js
```diff
@@ -0,0 +1,32 @@
...
+ * @typedef {import('src/spec/api-types').AppealSubmission} AppealSubmission
+ */
+
+const appealTypeToDirName = {
+	[APPEAL_ID.HOUSEHOLDER]: 'has'
+};
+
+/**
+ * this does "work" (the type is visible in autocomplete etc)
+ * we should probably just tighten up what we expect to be in the db
```
Si3rr4wow, line 12: move this comment

packages/appeals-service-api/src/services/responses.service.js
```diff
@@ -47,36 +39,7 @@ const getResponse = async (journeyId, referenceId, projection) => {
...
-};
-
-const submitResponse = submitResponseFactory(mapQuestionnaireDataForBackOffice, broadcast);
-
 module.exports = {
 	getResponse,
-	mapQuestionnaireDataForBackOffice,
-	patchResponse,
-	submitResponse,
-	submitResponseFactory
```
Si3rr4wow, line 81: There's tests for this you can bin I think

packages/appeals-service-api/src/services/responses.service.js
```diff
@@ -47,36 +39,7 @@ const getResponse = async (journeyId, referenceId, projection) => {
...
-	}, []);
-
-	const getBlobMeta = blobMetaGetter(initContainerClient);
-
-	const uploadedFilesAndBlobMeta = await conjoinedPromises(uploadedFiles, (uploadedFile) =>
-		getBlobMeta(uploadedFile.location)
-	);
-
-	return questionnaireMapper.mapToPINSDataModel(questionnaireResponse, uploadedFilesAndBlobMeta);
-};
```
Si3rr4wow, line 63: The guts of this moved into the formatter

packages/appeals-service-api/src/services/responses.service.js
```diff
@@ -47,36 +39,7 @@ const getResponse = async (journeyId, referenceId, projection) => {
...
-
-const submitResponseFactory = (mapper, callback) => async (questionnaireResponse) => {
-	try {
-		return await callback(await mapper(questionnaireResponse));
-	} catch (err) {
-		logger.error(err);
-		throw ApiError.unableToSubmitResponse();
```
Si3rr4wow, line 70: Something else is gunna have to be responsible for doing this now

packages/appeals-service-api/src/services/back-office-v2/index.js
```diff
@@ -0,0 +1,43 @@
...
+			throw new Error(`Appeal type could not be determined on appeal ${appealId}`);
+
+		return await broadcast(formatters.appeal[appeal.appealType](appeal));
+	}
+
+	/**
+	 * @param {*} questionnaireResponse
+	 * @returns {Promise<Array<*> | void>}
+	 */
+	async submitQuestionnaire(questionnaireResponse) {
```
Si3rr4wow, line 31: `submitAppeal` just expects an id and is responsible for getting the data itself, but the controller where the existing `submitQuestionnaire` function was used is responsible for getting the data. Consider picking one of these paradigms and aligning these methods.

Si3rr4wow: I did this it was good

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/spec.yaml
```diff
@@ -84,3 +84,25 @@ paths:
             application/json:
               schema:
                  $ref: '#/components/schemas/ErrorBody'
+  /api/v2/appeal-cases/{caseReference}/lpa-questionnaire-submission/submit:
```
Si3rr4wow, line 87: Would there be opposition to dropping `submission` from the `lpa-questionnaire-submission`?

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/submit/controller.js
```diff
@@ -0,0 +1,10 @@
+const BackOfficeV2Service = require('../../../../../../services/back-office-v2');
+
+const backOfficeV2Service = new BackOfficeV2Service();
+
+/** @type {import('express').Handler} */
+exports.post = async (req, res) => {
+	await backOfficeV2Service.submitQuestionnaire(req.params.caseReference);
+
+	res.sendStatus(200);
```
Si3rr4wow, line 9: There didn't seem to be any need to respond to the front end with the mapped data, which is what the old thing used to do. I can add that in if we'd prefer though

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/submit/controller.js
```diff
@@ -0,0 +1,10 @@
+const BackOfficeV2Service = require('../../../../../../services/back-office-v2');
+
+const backOfficeV2Service = new BackOfficeV2Service();
+
+/** @type {import('express').Handler} */
+exports.post = async (req, res) => {
+	await backOfficeV2Service.submitQuestionnaire(req.params.caseReference);
```
Si3rr4wow, line 7: oh, you should probably catch an error here and throw that unable to submit one that used to be in response.service or wherever I left a note for myself

packages/appeals-service-api/src/services/back-office-v2/formatters/has/appeal.js
```diff
@@ -0,0 +1,77 @@
+const LpaService = require('../../../lpa.service');
+const lpaService = new LpaService();
+const ApiError = require('../../../../errors/apiError');
+
+/**
+ *
+ * @param {import('src/spec/api-types').AppealSubmission['appeal']} appeal
+ * @returns
+ */
+exports.formatter = async (appeal) => {
```
Si3rr4wow, line 10: This one is still grizzly, I just moved it here so we had a jumping off point

packages/appeals-service-api/src/services/back-office-v2/formatters/index.js
```diff
@@ -0,0 +1,33 @@
...
+ * @type {{
+ *   appeal: Record<AppealSubmission['appeal']['appealType'], (appeal: AppealSubmission['appeal']) => *>,
+ *   questionnaire: Record<AppealSubmission['appeal']['appealType'], (caseReference: string, questionnaireResponse: LPAQuestionnaireSubmission) => *>
+ * }}
+ */
+module.exports = ['appeal', 'questionnaire'].reduce(
```
Si3rr4wow, line 19: this reduces the adjacent directories to an object the `BackOfficeThingy` class can navigate like `formatter['questionnaire']['1001']` to get a function to format the relevant data. That's gunna mean that as we build out more types we don't need conditions in the `BackOfficeThingy` because the data will tell us that number and the method being used tells us what kind thing we need to format.

packages/appeals-service-api/src/services/back-office-v2/forwarders.js
```diff
@@ -0,0 +1,17 @@
...
+const config = require('../../configuration/config');
+
+/**
+ * @param {string} [topic]
+ * @returns {(data: Array<unknown>) => Promise<Array<unknown>>}
+ */
+const forward = (topic) => {
+	if (!topic) throw new Error('Service Bus topic has not been configured in this environment');
+	return (data) => eventClient.sendEvents(topic, data, EventType.Create);
+};
```
Si3rr4wow, line 12: This is based on the old `broadcast functions

packages/common/src/is-feature-active.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * @param {{ featureFlagging: { endpoint?: string, timeToLiveInMinutes: string | number } }} config
+ * @returns {(featureFlagName: string, localPlanningAuthorityCode?: string) => Promise<boolean>}
+ */
+exports.isFeatureActive = (config) => async (featureFlagName, localPlanningAuthorityCode) => {
```
Si3rr4wow, line 5: This code was in 3 packages, they now init their own instance from this by passing their configs

packages/common/src/utils.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
 	 * @template TResolution
 	 * @param {Arg[]} argArr
 	 * @param {(arg0: Arg) => Promise<TResolution>} asyncFunc
-	 * @returns
+	 * @returns {Promise<Map<Arg, TResolution>>}
```
Si3rr4wow, line 95: detail I'd overlooked, makes this much nicer to use

packages/forms-web-app/src/dynamic-forms/options-question.js
```diff
@@ -129,31 +138,61 @@ class OptionsQuestion extends Question {
 	/**
 	 * returns the data to send to the DB
 	 * side effect: modifies journeyResponse with the new answers
-	 * @param {ExpressRequest} req
+	 * @param {import('express').Request} req
 	 * @param {JourneyResponse} journeyResponse - current journey response, modified with the new answers
-	 * @returns {Promise.<Object>}
+	 * @returns {Promise<{ answers: Record<string, unknown> }>}
 	 */
 	async getDataToSave(req, journeyResponse) {
```
Si3rr4wow, line 145: This is the bit that "fixes" not saving irrelevant data so we don't need all the nested ternaries in the BO formatters. In essence it just nullifies answers to conditional questions that the user should not have answered

packages/forms-web-app/src/dynamic-forms/questions.js
```diff
@@ -40,6 +40,8 @@ const DateQuestion = require('./dynamic-components/date/question');
 const TextEntryQuestion = require('./dynamic-components/text-entry/question');
 const { documentTypes } = require('@pins/common');
 
+inputMaxCharacters = Math.min(Number(inputMaxCharacters), 32500);
```
Si3rr4wow, line 43: This was for a codeQL issue, it didn't like the regex being based off an unsanitised env var, this sets a max value of 32500 chars, the number we have as the default in our config

packages/forms-web-app/src/dynamic-forms/questions.js
```diff
@@ -358,7 +360,7 @@ exports.questions = {
 			new ConditionalRequiredValidator('Enter the reason'),
 			new StringValidator({
 				regex: {
-					regex: new RegExp(`^[0-9a-z- '()]{1,${inputMaxCharacters}}$`, 'gi'),
+					regex: new RegExp(`^[0-9a-z- '()]{0,${inputMaxCharacters}}$`, 'gi'),
```
Si3rr4wow, line 363: This was causing a bug that resulted in the user seeing the associated error even if they did not need to answer this conditional question. The bug precedes the other changes I've made

packages/forms-web-app/src/dynamic-forms/options-question.js
bryn-pins: Could this be simplified to the following or would that miss something?

foreach opt in this.options
if opt in req.body -> set value
else -> set null

packages/common/src/is-feature-active.js
Si3rr4wow: good thinking 🧠

packages/appeals-service-api/src/services/back-office-v2/formatters/index.js
```diff
@@ -0,0 +1,33 @@
...
+	(mod, submittable) => ({
+		...mod,
+		[submittable]: {
+			...Object.entries(appealTypeToDirName).reduce(
+				(formatters, [appealType, dirName]) => ({
```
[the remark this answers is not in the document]
Si3rr4wow: fair point, it does drill quite deep, I'll do some extraction 🚁

packages/forms-web-app/src/dynamic-forms/options-question.js
bryn-pins: Ah think I thought it was an issue because it's still in the DB, presumably it's an issue because it's still in the request.
Reckon it could still be simplified by doing null by default and then setting only the valid ones, rather than creating two arrays

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/submit/controller.js
Si3rr4wow: did that

packages/appeals-service-api/src/services/back-office-v2/formatters/index.js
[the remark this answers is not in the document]
Si3rr4wow: I've rearranged this and added some more typing, lemme know what you think

packages/common/src/is-feature-active.js
Si3rr4wow: I've done this now

packages/forms-web-app/src/dynamic-forms/options-question.js
Si3rr4wow: I can't think of a way to do this without making 2 arrays I'm afraid. To nullify everything I'd still need to construct an array of _all_ the conditionals so I can loop through that an nullify them. Then I'd loop through the valid ones. Am I missing something?

Si3rr4wow:
```
		// sort conditional subsections they should be able to answer form those they shouldn't
		const [validConditionalFieldNames, allConditionalFieldNames] = this.options.reduce(
			(/** @type {[string[], string[]]} */ acc, option) => {
				if (!option.conditional) return acc;
				const optionIsSelectedOption = selectedOptions.some(
					(selectedOption) =>
						option.text === selectedOption.text && option.value === selectedOption.value
				);
				if (optionIsSelectedOption) acc[0].push(option.conditional.fieldName)
				return [acc[0], [...acc[1], option.conditional.fieldName]]
			},
			[[], []]
		);

		// nullify all fields
		allConditionalFieldNames.forEach((conditionalFieldName) => {
			const key = getConditionalFieldName(this.fieldName, conditionalFieldName);
			responseToSave.answers[key] = null;
			journeyResponse.answers[key] = null;
		});

		// add data from each valid conditional answer to data to be saved
		validConditionalFieldNames.forEach((validConditionalFieldName) => {
			const key = getConditionalFieldName(this.fieldName, validConditionalFieldName);
			const conditionalAnswer = req.body[key];
			if (!conditionalAnswer) return;
			responseToSave.answers[key] = req.body[key];
			journeyResponse.answers[key] = req.body[key];
		});
```

bryn-pins: was thinking something along the lines of a single loop instead of 3.

untested example:
```js
this.options.forEach((opt) => {
  if (!opt.conditional) return;
  
  const optionIsSelectedOption = selectedOptions.some(
    (selected) => opt.text === selected.text && opt.value === selected.value
  );
  
  const key = getConditionalFieldName(this.fieldName, opt.conditional.fieldName);
  const conditionalAnswer = req.body[key];
  
  if (!optionIsSelectedOption || !conditionalAnswer) {
    responseToSave.answers[key] = null;
    journeyResponse.answers[key] = null;
    return;
  }
  
  responseToSave.answers[key] = conditionalAnswer;
  journeyResponse.answers[key] = conditionalAnswer;
});
```

Shorter but potentially less explicit, happy to have it as is

Si3rr4wow: 🤦‍♀️  right yeah sorry I get it, done

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/questionnaire-submission.d.ts
```diff
@@ -0,0 +1,16 @@
...
+				LPACode: true;
+				appealTypeCode: true;
+			};
+		};
+		SubmissionDocumentUpload: true;
+		SubmissionNeighbourAddress: true;
+	};
+}>;
+
+type thing = LPAQuestionnaireSubmission[''];
```
bryn-pins, line 16: should this be here

bryn-pins (approved):

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/lpa-questionnaire-submission/questionnaire-submission.d.ts
Si3rr4wow: oh lol no, whoops

bryn-pins (approved):

PR #3459, opened 2024-02-27
feat(aapd-1195): add auth server

bryn-pins:
## Ticket Number

## Description of change

Done:

- set up auth server
- added custom otp grant type to generate one time tokens
- added custom ropc grant type to auth server to handle logins

## Checklist

- [x] New features have tests

## Important


packages/auth-server/src/account/account.js
```diff
@@ -0,0 +1,85 @@
...
+				user = await repo.getById(id);
+			}
+
+			if (!user) throw new UnknownUserId(id);
+
+			return new Account(user.id, user);
+		}
+
+		return store.get(id);
+	}
```
Si3rr4wow, line 82: I think this is denestable? 
```suggestion
	static async fA(ctx, id, token) {
		if (store.has(id)) return store.get(id);

		let user;

		if (isEmailLike(id)) {
			id = id.trim();
			user = await repo.getByEmail(id);
		} else {
			user = await repo.getById(id);
		}

		if (!user) throw new UnknownUserId(id);

		return new Account(user.id, user);
	}
```

packages/auth-server/src/grants/otp-grant-handler.js
```diff
@@ -0,0 +1,76 @@
+import { InvalidClient, UnknownUserId } from 'oidc-provider/lib/helpers/errors.js';
```
bryn-pins, line 1: these grants are the core of the logic

packages/auth-server/src/configuration/oidc/index.js
```diff
@@ -0,0 +1,71 @@
...
+		features: {
+			resourceIndicators: resourceFeature,
+			clientCredentials: {
+				enabled: true
+			},
+			devInteractions: {
+				enabled: false
+			}
+		},
+		ttl: {
```
bryn-pins, line 56: these time to live settings may need some thought, we can't have an access token that has session length, so users need to log in again once it expires.
Have gone with a day for now

packages/auth-server/src/configuration/featureFlag.js
```diff
@@ -0,0 +1,76 @@
+// todo use version in common when in /pull/3454
```
bryn-pins, line 1: see todo, will use shared version once in

packages/auth-server/src/account/repository.js
```diff
@@ -0,0 +1,91 @@
+import { Prisma } from '@prisma/client';
```
bryn-pins, line 1: this has come from appeals api, can potentially remove some of it once in/move all user logic to this api

packages/auth-server/dev/gen-client.js
```diff
@@ -0,0 +1,19 @@
+import crypto from 'crypto';
+
+/**
+ * @returns a client id + secret for local dev use
+ */
+async function generate() {
+	return {
+		client_id: crypto.randomUUID(),
```
bryn-pins, line 8: these are just for local, will use terraform generated

packages/auth-server/__tests__/tests.spec.js
bryn-pins, line 1: majority of testing is here, set up in a similar fashion to appeals api

packages/auth-server/__tests__/globalSetup.js
```diff
@@ -0,0 +1,23 @@
...
+			if (err) {
+				reject(err);
+			} else {
+				resolve(stdout);
+			}
+		});
+	});
+}
+
+export default async () => {
```
bryn-pins, line 18: slight difference between appeals api here, in that setup/teardown is done once for notify mock, rather than on every test

packages/auth-server/src/account/account.js
```diff
@@ -0,0 +1,85 @@
...
+	 *   based on this detail or not return them in id tokens but only userinfo and so on.
+	 * @param {{ [key: string]: null | import('oidc-provider').ClaimsParameterMember}} claims
+	 *   the part of the claims authorization parameter for either "id_token" or "userinfo" ("use" param)
+	 * @param {string[]} rejected claim names that were rejected by the end-user, you might want to skip
+	 *   loading some claims from external resources or through db projection
+	 * @returns {Promise<import('oidc-provider').UnknownObject>}
+	 */ // eslint-disable-next-line no-unused-vars
+	async claims(use, scope, claims, rejected) {
+		return {
+			sub: this.accountId, // it is essential to always return a sub claim
```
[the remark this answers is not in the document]
beejjacobs: Yes I think for our initial use case we can keep this minimal and review.

packages/auth-server/src/adapter/prisma-adapter.js
```diff
@@ -0,0 +1,200 @@
...
+	'ReplayDetection',
+	'PushedAuthorizationRequest',
+	'Grant',
+	'BackchannelAuthenticationRequest'
+].reduce((map, name, i) => ({ ...map, [name]: i + 1 }), {});
+
+/**
+ * @param { import("@prisma/client").Oidc } doc
+ */
+const prepare = (doc) => {
```
beejjacobs, line 30: Could this have a return type? Oidc update input is it?

beejjacobs (changes requested):
Got some suggestions, but generally minor. The docs and layout help make this follow-able. Looks very thorough, with some details to finalise like scopes.

packages/auth-server/src/adapter/prisma-client.js
```diff
@@ -0,0 +1,15 @@
+import createClient from '@pins/database/src/create-client.js';
+
+/** @type {import('@pins/database/src/create-client').prismaConfig} */
+const prismaConfig = {
+	datasourceUrl: process.env.SQL_CONNECTION_STRING
```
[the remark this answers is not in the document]
Si3rr4wow: Ooh here's a scab to pick. I have liked having the config objects because you get the inferred type and autocomplete _but_ we could just type the process by having a declaration file like this

```
/** Defines the typings for `process.env.VARIABLES` */
export interface Climate360ProcessEnv {
  NEXT_PUBLIC_BACKEND_BASE_URL: string
}

declare global {
  // eslint-disable-next-line @typescript-eslint/no-namespace
  namespace NodeJS {
    // https://github.com/typescript-eslint/typescript-eslint/issues/1640
    // eslint-disable-next-line @typescript-eslint/no-empty-interface
    interface ProcessEnv extends Climate360ProcessEnv {
      // add members to Climate360ProcessEnv, not here
    }
  }
}
```

I think it's also worth introducing something like this 

```
/** Checks whether the environment variables conform to expectations */
export function checkEnv(): void {
  const boolStringAllowedValues = ["true", "false"];
  const nonEmptyRegex = /.+/;
  const pathRegex = /^\/.*/;

  checkRegex("API_KEY", nonEmptyRegex);
  checkString("ASSET_PREFIX", true);
  checkRegex("CI_VERSION", nonEmptyRegex);
  checkType("CONTENT_SECURITY_POLICY", ["local", "secure"]);
  checkRegex("HOST_ENV", nonEmptyRegex);
  // dev-only env vars
  checkRegex("API_PROXY_PATH", pathRegex, true);
  checkURL("API_PROXY_URL", true);
  checkType("DISABLE_APPD", boolStringAllowedValues, true);
  checkType("ENABLE_API_PROXY", boolStringAllowedValues, true);
  checkType("USE_LOCAL_SSL", boolStringAllowedValues, true);
  checkType("SILENCE_WINSTON", boolStringAllowedValues, true);
  checkString("APPDYNAMICS_ACCESS_KEY");
}
```

to run as the app boots so we know it's configured correctlyish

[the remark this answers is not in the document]
beejjacobs: We use `joi` on the back office to check config values:
https://github.com/Planning-Inspectorate/back-office/blob/main/apps/api/src/server/config/schema.js#L3

packages/auth-server/dev/gen-keys.js
```diff
@@ -0,0 +1,24 @@
...
+async function generate() {
+	// supported values: https://github.com/panva/node-oidc-provider/blob/270af1da83dda4c49edb4aaab48908f737d73379/lib/consts/jwa.js#L1
+	// 'HS256', 'HS384', 'HS512',
+	// 'RS256', 'RS384', 'RS512',
+	// 'PS256', 'PS384', 'PS512',
+	// 'ES256', 'ES256K', 'ES384', 'ES512',
+	// 'EdDSA'
+	const keyPair = await jose.generateKeyPair('specify an algorithm');
+	const privateJwk = await jose.exportJWK(keyPair.privateKey);
+	return JSON.stringify(privateJwk);
```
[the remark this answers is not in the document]
bryn-pins: can potentially add something to do this, do we have an example somewhere of updating a key vault secret outside of terraform in a pipeline? Would need to be be able to handle rotation, so wouldn't be a simple pipeline (i.e. needs to have option to wait for all existing tokens to expire then remove old key from array)

packages/auth-server/src/adapter/prisma-client.js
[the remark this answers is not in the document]
bryn-pins: looking into the joi implementation, seems to log redacted secrets on failure which probably isn't ideal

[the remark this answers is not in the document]
bryn-pins: this is done

packages/auth-server/dev/gen-keys.js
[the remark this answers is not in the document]
beejjacobs: BO have a basic pipeline running a npm script, which runs on a schedule. We can do a separate PR for that if we go down that route?

https://github.com/Planning-Inspectorate/back-office/blob/main/azure-pipelines-regenerate-api-key.yml

beejjacobs (approved):
Changes addressed. LGTM

PR #3463, opened 2024-02-29
feat(aapd-1195): use oauth for enter code login

bryn-pins:
## Ticket Number

## Description of change

Enter code to use oauth and keep tokens in session for use in api calls

## Checklist

- [x] New features have tests

## Important


packages/appeals-service-api/src/lib/notify.js
```diff
@@ -219,7 +181,7 @@ const sendSecurityCodeEmail = async (recipientEmail, code, identifier, action =
 			'Sending secure code email to appellant'
 		);
 		await NotifyBuilder.reset()
-			.setTemplateId(mapActionToTemplate(action))
+			.setTemplateId(templates.SAVE_AND_RETURN.enterCodeIntoServiceEmailToAppellant)
```
bryn-pins, line 184: there is no separate lpa email so have removed this logic

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -397,7 +402,15 @@ const postEnterCodeLPA = (views) => {
 		}
 
 		// check token
-		const tokenResult = await isTokenValid(emailCode, id, user.email, req.session, user.lpaCode);
+		const enrolUsers = true;
```
beejjacobs, line 405: Is this not meant to be the flag value? `enrolUsersFlag`?

packages/forms-web-app/src/lib/is-token-valid.js
```diff
@@ -63,21 +69,68 @@ const getToken = async (token, id, emailAddress, action) => {
 
 /**
  * @param {string} token
- * @param {string} [id] - appealId or userid
+ * @param {string} [emailAddress]
+ * @param {string} [action]
+ * @returns {Promise<TokenValidResult>}
+ */
+const authToken = async (token, emailAddress, action) => {
```
beejjacobs, line 76: Not sure why, I find the name of this incomplete, needs a verb?

packages/forms-web-app/src/middleware/check-logged-in.js
```diff
@@ -24,7 +24,7 @@ const checkLoggedIn = async (req, res, next) => {
 
 	const user = getAppealUserSession(req);
 
-	if (user && user?.id) {
+	if (user && user?.expiry.getTime() > Date.now()) {
```
beejjacobs, line 27: Feels a much better check!

beejjacobs (commented):
Code looks OK. Feels like this is important enough I should run it and check properly!

packages/forms-web-app/src/controllers/common/enter-code.js
bryn-pins: this is for lpas
the entire process relies on the sql user so this should always be true
perhaps it gets reworked if we split out appellant-dashboard and sql users into separate flags

beejjacobs (approved):

PR #3473, opened 2024-03-05
feat(aapd-1195): check documents access via auth tokens

bryn-pins:
## Ticket Number

## Description of change

Use tokens to check document access

## Checklist

- [x] New features have tests

## Important


packages/auth-server/src/account/account.js
```diff
@@ -46,11 +46,19 @@ class Account {
...
 	 */ // eslint-disable-next-line no-unused-vars
 	async claims(use, scope, claims, rejected) {
-		return {
+		const details = {
 			sub: this.accountId, // it is essential to always return a sub claim
 			email: this.user.email
 		};
 
+		if (this.user.isLpaUser) {
+			details.lpaCode = this.user.lpaCode;
```
bryn-pins, line 55: add lpa details to id token

packages/document-service-api/test/developer/developer.test.js
```diff
@@ -1,15 +1,74 @@
...
 	isFeatureActive: () => true
 }));
 
+jest.mock('express-oauth2-jwt-bearer', () => {
+	return {
+		auth: jest.fn(() => {
+			return (req, res, next) => {
+				req.auth = {
+					payload: {
+						sub: '29670d0f-c4b4-4047-8ee0-d62b93e91a11'
```
bryn-pins, line 28: couldn't figure out how to vary this by test

packages/auth-server/src/account/account.js
```diff
@@ -46,11 +46,19 @@ class Account {
...
+			details.isLpaAdmin = this.user.isLpaAdmin;
+			details.lpaStatus = this.user.lpaStatus;
```
Si3rr4wow, line 57: what is lpaStatus?

bryn-pins: added|confirmed|removed
can't login with removed
might not be needed for anything but thought I'd just include the lpa specific user details

packages/document-service-api/test/developer/developer.test.js
Si3rr4wow: I think the options are to either remock the entire function or you could try setting the auth header on the supertest request, but that might not work

Si3rr4wow: Oh or we could have a common func got getting the sub from a request then we mock that function, like
```
exports.getSub = (req) => req.auth.payload.sub
```
```
jest.mock('../getSub', () => ({
  getSub: () => <some guid>
}))
```

then you can just mock the return value

```
getSub.mockReturnValueOnce(<some other guid>)
```

Si3rr4wow (approved):
I'd like to use the middleware you've made here so shall we get this in, I'll rebase 1502 and improve your test if I can find a nicer way to mock?

PR #3474, opened 2024-03-06
Fix/aapd 516

lauren-PINS:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-516

## Description of change

Added journey type to the html path to match copydeck behaviour.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/controller.js
+++ b/packages/forms-web-app/src/dynamic-forms/controller.js
@@ -235,13 +235,25 @@ exports.submit = async (req, res) => {
 	const { referenceId } = req.params;
 	const journeyResponse = res.locals.journeyResponse;
 	const journey = getJourney(journeyResponse);
+
+	const journeyUrl = (journeyId) => {
+		if (journeyId === 'has-questionnaire') {
+			return 'householder/';
+		} else if (journeyId === 's78-questionnaire') {
+			return 'full-planning/';
+		} else return '';
+	};
+
 	if (!journey.isComplete()) {
 		res.sendStatus(400);
 		return;
 	}
 	await apiClient.submitLPAQuestionnaire(referenceId);
 	return res.redirect(
-		'/manage-appeals/' + encodeURIComponent(referenceId) + '/questionnaire-submitted/'
+		'/manage-appeals/' +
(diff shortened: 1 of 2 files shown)


packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -238,10 +238,23 @@ exports.submit = async (req, res) => {
 	const { referenceId } = req.params;
 	const journeyResponse = res.locals.journeyResponse;
 	const journey = getJourney(journeyResponse);
+
+	const journeyUrl = (journeyId) => {
```
Si3rr4wow, line 242: In situations like this (where you've got a 1:1 mapping from one string to another and you're likely to add more things to it later) it's often preferable to use a dictionary style approach rather than having many bits of branching logic. If you made an object like

```
const journeyIdToUrl = {
  'has-questionnaire': 'householder/',
  's78-questionnaire': 'full-planning'
}
```

then, so long as you verify that the given id is valid, you can unconditionally return the matching element from the dictionary 

```
'/manage-appeals/' +
				journeyIdToUrl[journeyResponse.journeyId] +
				encodeURIComponent(referenceId) +
				'/questionnaire-submitted/'
```

packages/forms-web-app/src/dynamic-forms/route.js
```diff
@@ -27,12 +27,20 @@ router.post(
 router.post('/questionnaire/:referenceId/', getJourneyResponse(), validationErrorHandler, submit);
 
 router.get(
```
Si3rr4wow, line 29: oh! and if you did make that dictionary I mentioned then you could loop through its keys to make these routes!

Si3rr4wow (dismissed):
lgtm, one suggestion but feel free to leave it

packages/forms-web-app/src/dynamic-forms/controller.js
lauren-PINS: Just going to get this through for now, but thank you, this is a great point! I'll look into fixing it at a later date

bryn-pins (approved):

bryn-pins (approved):

Si3rr4wow (approved):

PR #3476, opened 2024-03-07
feat(aapd-1400): lpa routes to use sql

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1400

## Description of change

Ensures all paths with lpa pages uses sql where possible (some data is still in mongo such as lpas)
Added user management endpoints and used the user-appealcase to get appeal details

## Checklist

- [x] New features have tests

## Important


packages/appeals-service-api/src/routes/v2/users/index.js
```diff
@@ -1,12 +1,16 @@
 const express = require('express');
-const { userGet, userPost, userLink } = require('./controller');
+const { userPost, userSearch, userGet, userUpdate, userDelete, userLink } = require('./controller');
 const router = express.Router();
 const asyncHandler = require('@pins/common/src/middleware/async-handler');
+const { openApiValidatorMiddleware } = require('../../../validators/validate-open-api');
 
-// Would be nice to untangle this at some point, we're using
```
bryn-pins, line 6: do we want to allow either like I have done here

packages/appeals-service-api/src/routes/v2/users/_id/appeal-cases/_caseReference/service.js
```diff
@@ -11,7 +11,9 @@ exports.get = async ({ caseReference, userId, role }) => {
 	if (role === LPA_USER_ROLE) {
 		// handle LPA user, no explicit appeal <-> user link
 		const data = await repo.getForLpaUser({ caseReference, userId });
-		if (data) return data;
+		if (data) {
+			return caseService.appendAppellant(data);
```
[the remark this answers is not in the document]
beejjacobs: Seems OK!

packages/appeals-service-api/src/routes/v2/users/controller.js
```diff
@@ -1,76 +1,110 @@
...
-		}
-	} finally {
-		res.status(statusCode).send(body);
-	}
+async function userSearch(req, res) {
+	const { lpaCode } = req.query;
+
+	const body = await searchUsers({
+		lpaCode: lpaCode?.trim()
+	});
```
beejjacobs, line 31: Worth a 400 return here if not defined? Saves a DB call.

packages/appeals-service-api/src/routes/v2/users/controller.js
```diff
@@ -1,76 +1,110 @@
...
+
+/**
+ * @type {import('express').RequestHandler}
+ */
+async function userUpdate(req, res) {
+	const user = await resolveUser(req.params.userLookup);
+
+	const { isEnrolled, lpaStatus } = req.body;
+	if (isEnrolled) user.isEnrolled = isEnrolled;
+	if (lpaStatus) user.lpaStatus = lpaStatus;
```
beejjacobs, line 52: Do we need to check for `!== undefined` rather than truthy?

packages/appeals-service-api/src/routes/v2/users/controller.js
```diff
@@ -1,76 +1,110 @@
...
+ */
+async function resolveUser(userLookup) {
+	let user;
+
+	userLookup = userLookup.trim();
+
+	if (userLookup.includes('@')) {
+		user = await getUserByEmail(userLookup);
+	} else {
+		user = await getUserById(userLookup);
```
beejjacobs, line 97: Can just return?

packages/appeals-service-api/src/routes/v2/users/index.js
beejjacobs: Works for me!

beejjacobs (changes requested):
This was more involved than first thought... good to unpick this.

A few minor points, but generally looks good and thorough.

packages/appeals-service-api/src/routes/v2/users/controller.js
bryn-pins: So I added the open api format checker that validates the type with 400
lpaStatus is a string so think that should be fine

isEnrolled is boolean, but I think this should only work 1 way, i.e. can't unenroll

bryn-pins: Will add a comment

beejjacobs (approved):
Comments addressed. LGTM.

PR #3484, opened 2024-03-13
fix(aapd-1344): refactor web comment no results page to distinguish b…

DeanH1990:
…etween type of search

## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1344

## Description of change

correct web comment no search result page depending on if the user searched by post code or appeal reference number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/web-comment/src/routes/appeal-search-no-results/format-no-results-page.js
+++ b/packages/web-comment/src/routes/appeal-search-no-results/format-no-results-page.js
@@ -1,22 +1,4 @@
 /**
- * @param {string} searchQuery
- * @returns {string}
- */
-exports.isPostcodeOrReferenceNumber = (searchQuery) => {
-	const fullPostcodeRegex =
-		/^([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([AZa-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9]?[A-Za-z])))) [0-9][A-Za-z]{2})$/;
-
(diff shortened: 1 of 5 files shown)


packages/web-comment/src/routes/appeal-search-no-results/format-no-results-page.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * @param {string} searchQuery
+ * @returns {string}
+ */
+exports.isPostcodeOrReferenceNumber = (searchQuery) => {
```
bryn-pins, line 5: before either of these regexes went in I'd want a lot of unit tests for them
however rather than managing a complex regex, aren't we just passing through a query string telling us this on redirect?

```
res.redirect(`appeal-search-no-results?search=
```

bryn-pins: I've noticed this is also used in the enter-postcode controller, I don't think it should be

DeanH1990: I removed the regex and added the search type as a param in the redirect. This wasnt already done if thats what you meant

bryn-pins: ah yes sorry was being dumb, but yes I think this is preferable

DeanH1990: yea it definitely seems like a smoother approach.

Think changing or amending the regex in the other controller should probably be a separate ticket?

bryn-pins: yes, will attempt to think of a simpler approach

PR #3488, opened 2024-03-14
Feature/aapd 1195 oauth send token

bryn-pins:
## Ticket Number

## Description of change

Send token via auth server
v2 token code removed as this is now fully replaced with auth server
use a shared constant for grant type string

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -52,15 +76,17 @@ const getEnterCode = (views, { isGeneralLogin = true }) => {
 			delete req.session?.enterCode?.newCode;
 		}
 
+		const sqlUser = await isFeatureActive(FLAG.ENROL_USERS);
```
beejjacobs, line 79: I think making it clear this is a feature flag in the variable name will help with clarity in the rest of the code.

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -317,13 +352,13 @@ const lpaTokenVerification = (res, token, views, id) => {
  * @param {import('express').Request} req
  * @returns {Promise<void>}
  */
-async function sendTokenToLpaUser(req) {
+const sendTokenToLpaUser = async (req) => {
```
beejjacobs, line 355: Intrigued by this change! I personally prefer explicit functions unless there's a need for arrow function, but we rarely reference `this` where it would make a difference. It seems there's a general preference to use arrow functions though.

beejjacobs (commented):
Looks pretty clear. Should we remove the token table too?

packages/forms-web-app/src/controllers/common/enter-code.js
bryn-pins: I'm not that fussed but largely done here for consistency with every other function in this file.
But I'm sold on the idea of const arrow functions by default, can't reassign, no hidden hoisting, no this confusion
less sold on multiple nested anonymous arrow functions

packages/forms-web-app/src/controllers/common/enter-code.js
bryn-pins: agreed yes, I did this before I started on the split of the feature flag, but this was a poor choice

bryn-pins:
> Looks pretty clear. Should we remove the token table too?

It still uses the token table under the hood, haven't changed that logic
There was probably a way to avoid it but this was simplest

packages/forms-web-app/src/controllers/common/enter-code.js
beejjacobs: Good enough reasons for me!

beejjacobs:
> > Looks pretty clear. Should we remove the token table too?
> 
> It still uses the token table under the hood, haven't changed that logic There was probably a way to avoid it but this was simplest

That's just me not thinking...

beejjacobs (approved):

PR #3527, opened 2024-04-08
feat: add authorized api clients via middleware

bryn-pins:
## Description of change

Adds api clients to all requests, authorized if user is present otherwise will use client credentials

Need to replace lib/appeals-api-client.js going forward, will do in a separate PR

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/middleware/create-api-clients.js
+++ b/packages/forms-web-app/src/middleware/create-api-clients.js
@@ -1,0 +1,63 @@
+const { AppealsApiClient } = require('@pins/common/src/client/appeals-api-client');
+const { DocumentsApiClient } = require('@pins/common/src/client/documents-api-client');
+const getAuthClient = require('@pins/common/src/client/auth-client');
+const { AUTH } = require('@pins/common/src/constants');
+
+const { getAppealUserSession } = require('../services/appeal-user.service');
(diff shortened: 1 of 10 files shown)


packages/forms-web-app/src/middleware/create-api-clients.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @type {import('express').Handler}
+ */
+const createApiClients = async (req, res, next) => {
+	const lpaUser = getLPAUserFromSession(req);
+	const appellantUser = getAppealUserSession(req);
+
+	/** @type {import('@pins/common/src/client/appeals-api-client').AuthTokens} */
+	const auth = {
+		access_token: lpaUser?.access_token ?? appellantUser?.access_token,
```
bryn-pins, line 48: unsure what's best here, thinking it was easiest if we use the same property name so we can reuse controllers but perhaps that is a mistake

bryn-pins: could be a bit buggy if an lpa user ever logs in then tries to submit an appeal
more likely to cause issues in testing than real use, perhaps we remove existing users when logging in

Si3rr4wow (approved):

PR #3619, opened 2024-05-22
Feat/aapd 1656 multifield tasklist display

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1656

## Description of change

Amends the way that contact details questions are displayed on the task list - adds a formatType param to the MultiFieldInputQuestion type which determines the task list display method.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-field-input/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-field-input/question.js
@@ -1,4 +1,6 @@
 const Question = require('../../question');
+const escape = require('escape-html');
+const { nl2br } = require('@pins/common/src/utils');
 
 /**
  * @typedef {import('../../question').QuestionViewModel} QuestionViewModel
@@ -12,6 +14,7 @@ const Question = require('../../question');
  * @typedef {Object} InputField
  * @property {string} fieldName
  * @property {string} label
+ * @property {string} formatJoinString
  */
 
 /**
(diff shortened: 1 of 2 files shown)


packages/forms-web-app/src/dynamic-forms/questions.js
```diff
@@ -1602,6 +1602,7 @@ exports.questions = {
 		html: 'resources/your-details/applicant-name.html',
 		fieldName: 'applicantName',
 		url: 'applicant-name',
+		formatType: 'contactDetails',
```
bryn-pins, line 1605: can we move the 3 uses of this string into a const
maybe in multi-field-input/question.js

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-field-input/question.js
```diff
@@ -117,6 +137,45 @@ class MultiFieldInputQuestion extends Question {
...
+		const companyNameField =
+			inputFields.find((inputField) => inputField.fieldName.includes('CompanyName'))?.fieldName ||
+			'';
+
+		if (!journey.response.answers[firstNameField]) return this.NOT_STARTED;
+
+		const contactName = `${journey.response.answers[firstNameField]} ${journey.response.answers[lastNameField]}`;
+		const companyName = journey.response.answers[companyNameField];
+
+		return contactName + (companyName ? `<br>${companyName}` : '');
```
bryn-pins, line 161: we need this to be escaped for if a user enters html into one of the fields
may want to use the escape function: const escape = require('escape-html');

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-field-input/question.js
```diff
@@ -117,6 +137,45 @@ class MultiFieldInputQuestion extends Question {
...
 	}
+
+	/**
+	 * returns the formatted answers values to be used to build task list elements
+	 * @param {Journey} journey
+	 * @param {InputField[]} inputFields
+	 * @returns {String}
+	 */
+	formatContactDetails(journey, inputFields) {
+		const firstNameField =
```
bryn-pins, line 148: could this logic be held in input fields somehow so that it is generic
something like, or is that too complex?

foreach inputfield
format += inputField.formatForSummary(answers)

packages/forms-web-app/src/dynamic-forms/questions.js
topickering: No need for this now, I've set a formatJoinString field on the InputField which handles formatting - default is for each input to appear on a new line.

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-field-input/question.js
topickering: I've updated to use escape and nl2br

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-field-input/question.js
topickering: See above - I've amended to use a formatJoinString field on InputField, with a '\n' default.

packages/forms-web-app/src/dynamic-forms/dynamic-components/multi-field-input/question.js
```diff
@@ -39,7 +49,8 @@ class MultiFieldInputQuestion extends Question {
 		html,
 		label,
 		inputAttributes = {},
-		inputFields
+		inputFields,
+		formatType
```
bryn-pins, line 53: think references to formatType can go now? Otherwise looks good

topickering: Yep, good spot!

bryn-pins (approved):

PR #3637, opened 2024-05-31
fix: delete v1 appeal after creating a v2

bryn-pins:
## Description of change

when creating a v2 appeal we need to delete the v1 created during before you start to avoid duplicates in the dashboard

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/appeals-repository.js
+++ b/packages/appeals-service-api/src/repositories/appeals-repository.js
@@ -71,6 +71,14 @@ class AppealsRepository extends MongoRepository {
 				{ returnDocument: 'after' }
 			);
 	}
+
+	/**
+	 * @param {string} appealId
+	 * @return {Promise<any>}
+	 */
+	async delete(appealId) {
+		return await mongodb.get().collection(this.collectionName).deleteOne({ _id: appealId });
+	}
 }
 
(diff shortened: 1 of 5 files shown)


packages/appeals-service-api/src/repositories/appeals-repository.js
```diff
@@ -71,6 +71,14 @@ class AppealsRepository extends MongoRepository {
 				{ returnDocument: 'after' }
 			);
 	}
+
+	/**
+	 * @param {string} appealId
+	 * @return {Promise<any>}
+	 */
+	async delete(appealId) {
+		return await mongodb.get().collection(this.collectionName).deleteOne({ _id: appealId });
```
DeanH1990, line 80: guessing the findOneAndDelete or whatever it was before was the issue?

DeanH1990 (approved):

packages/appeals-service-api/src/repositories/appeals-repository.js
bryn-pins: I was doing appeal.id here but only had the id passed through
looked into the docs a little harder before realising that and I believe findOneAndDelete returns the deleted document which we didn't need so switched to the simpler function

PR #3645, opened 2024-06-05
chore: update functions version to v4

bryn-pins:
## Description of change

- update functions to v4 programming model
- update extension version to v4 (unrelated)
- add event function
- add service user function
- add basic tests for has appeal function

## Checklist

- [x] New features have tests

## Important


packages/integration-functions/src/functions/appeal-document.js
```diff
@@ -8,21 +18,16 @@ const VALID_REDACTED_STATUSES = ['redacted'];
...
- * @param {ServiceBusMessage} msg
+ * @type {import('@azure/functions').ServiceBusTopicHandler}
  */
-module.exports = async function (context, msg) {
-	context.log('Handle document metadata message', msg);
-	await processDocumentMetadata(context, msg.body);
+const handler = async (message, context) => {
+	context.log('Handle document metadata message', message);
+	await processDocumentMetadata(context, message.body);
+	return {};
```
Si3rr4wow, line 26: Why this? Is it just a service bus thing?

Si3rr4wow (approved):

packages/integration-functions/src/functions/appeal-document.js
bryn-pins: I think I'm just wrong, took from the docs regarding migrating:
https://learn.microsoft.com/en-gb/azure/azure-functions/functions-node-upgrade-v4?tabs=v4&pivots=programming-language-javascript#set-the-primary-output-as-your-return-value

But think you're right I didn't need to add it:
https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=python-v2%2Cisolated-process%2Cnodejs-v4%2Cextensionv5&pivots=programming-language-javascript#example

I'll remove

bryn-pins: ah I get a ts error when I don't return anything...

PR #3650, opened 2024-06-06
feat(aapd-1663): move web comment into forms-web-app

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1663

## Description of change

Moves web comment into forms-web-app under /comment-planning-appeal
## Checklist

- [ ] New features have tests

## Important


packages/common/src/regex.js
```diff
@@ -0,0 +1,11 @@
+// https://github.com/ideal-postcodes/postcode MIT
+
+/**
+ * Tests for a valid postcode
+ */
+module.exports.fullPostcodeRegex = /^[a-z]{1,2}\d[a-z\d]?\s*\d[a-z]{2}$/i;
```
bryn-pins, line 6: we need to handle ignoring whitespace in postcodes with the sql query at some point

packages/web-comment/package.json
beejjacobs, line 1: If we're removing this package there should be some `package-lock` changes too.

packages/common/src/regex.js
topickering: Do we have a bug / feat ticket for this?

bryn-pins: I've reverted the postcode regex to enforce spaces for now and will get a ticket in

topickering (dismissed):
Looks good to me, good luck with conflicts!

topickering (approved):

PR #3652, opened 2024-06-10
feat(1755): pdf download for appeal form v2

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1755

## Description of change

Adds steps to create pdf of appeal and create link for download on submission of v2 appeal

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -282,6 +310,58 @@ exports.appellantSubmissionDeclaration = async (req, res) => {
 	});
 };
 
+/**
+ * @param {import('express').Request} req
+ * @param {import('express').Response} res
+ */
+exports.appellantSubmissionInformation = async (req, res) => {
```
bryn-pins, line 317: can we put the shared logic into a function and just not use the status bits?

bryn-pins (commented):
Think this looks good, minor comment around code reuse but nice work

packages/forms-web-app/src/dynamic-forms/controller.js
topickering: Good call, I've made the status param optional (with a default of an empty string) in the buildSectionViewModel function

bryn-pins: I meant the shared logic between list and appellantSubmissionInformation but it's not a major issue so not to worry

bryn-pins (approved):

PR #3665, opened 2024-06-14
Task/update pino

bryn-pins:
## Description of change

- removes headers from logs as they contain auth details (required before sql users go live)
- skips certain requests from being logged to clean up logs
- updates pino
- moves to pino-http over deprecated express-pino-logger
- updates pino-pretty

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/src/lib/logger.js
```diff
@@ -5,11 +5,18 @@
...
 	redact: config.logger.redact,
-	prettyPrint: config.logger.prettyPrint
+	transport: config.logger.prettyPrint
+		? {
+				target: 'pino-pretty',
+				options: {
+					colorize: true
+				}
+		  }
+		: undefined
```
beejjacobs, line 21: Will this log as JSON if `prettyPrint==false`?

beejjacobs (commented):
Bit of repetition between the apps, but I don't think that's worth tackling as part of this.

Also, you can specify `"*"` as a package version and just include a specific version in the root `package.json`, I think. Again, not something that needs to change.

Just one question about JSON logging....

packages/appeals-service-api/src/lib/logger.js
beejjacobs: As agreed, to deploy and check in DEV.

beejjacobs (approved):

PR #3667, opened 2024-06-17
feat(ASI-58): add service-user API

Si3rr4wow:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/sql/service-user-repository.js
+++ b/packages/appeals-service-api/src/repositories/sql/service-user-repository.js
@@ -89,8 +89,19 @@ class ServiceUserRepository {
 				});
 
 				if (appealCase) {
-					await tx.appealToUser.create({
-						data: {
+					await tx.appealToUser.upsert({
+						where: {
+							appealId_userId: {
+								appealId: appealCase.Appeal.id,
+								userId: appealUser.id
+							}
+						},
+						create: {
+							appealId: appealCase.Appeal.id,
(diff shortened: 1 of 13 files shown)


bryn-pins (commented):
Could you look into the points from ticket regarding linking the service user to appeals/appeal users please:

- Find AppealUser by email (service-user#emailAddress), and add serviceUserId if exists
- If no AppealUser found and service-user#emailAddress is set, add a new AppealUser
- If an AppealCase exists with the service-user#caseReference 
   then add AppealToUser relation based on serviceUserType and caseReference

Si3rr4wow:
@bryn-pins oops sorry done

packages/appeals-service-api/src/repositories/sql/service-user-repository.js
```diff
@@ -26,6 +38,88 @@ class ServiceUserRepository {
...
+						Appeal: {
+							select: {
+								id: true
+							}
+						}
+					}
+				});
+
+				if (appealCase) {
+					await tx.appealToUser.create({
```
bryn-pins, line 92: am I right in thinking this will create a duplicate appealToUser record if the service user is updated and this already exists? thinking it needs to be an upsert on appealUser.id

Si3rr4wow: ah yep good point, will change

bryn-pins (approved):
think this looks good

PR #3683, opened 2024-06-21
Feat/aapd 1778 why inquiry

DeanH1990:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1778
https://pins-ds.atlassian.net/browse/AAPD-1779
https://pins-ds.atlassian.net/browse/AAPD-1780

## Description of change

adds inquiry questions to s78 journey

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/number-entry/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/number-entry/question.js
@@ -1,4 +1,5 @@
 const Question = require('../../question');
+const { getPersistedNumberAnswer } = require('../utils/persisted-number-answer');
 
 /**
  * @typedef {import('../../journey').Journey} Journey
@@ -40,12 +41,7 @@ class NumberEntryQuestion extends Question {
 		let viewModel = super.prepQuestionForRendering(section, journey, customViewData, payload);
 
 		const answer = journey.response.answers[this.fieldName];
-		const persistedAnswer =
-			answer !== undefined && answer !== null && answer === 0
-				? `${answer}`
-				: answer !== undefined && answer !== null
-				? answer
-				: '';
+		const persistedAnswer = getPersistedNumberAnswer(answer);
 
 		viewModel.question.value = persistedAnswer;
 		viewModel.answer = persistedAnswer;
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/number-entry/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/number-entry/question.js
@@ -39,6 +39,17 @@ class NumberEntryQuestion extends Question {
 	prepQuestionForRendering(section, journey, customViewData, payload) {
 		let viewModel = super.prepQuestionForRendering(section, journey, customViewData, payload);
 
+		const answer = journey.response.answers[this.fieldName];
+		const persistedAnswer =
+			answer !== undefined && answer !== null && answer === 0
+				? `${answer}`
+				: answer !== undefined && answer !== null
+				? answer
+				: '';
+
+		viewModel.question.value = persistedAnswer;
+		viewModel.answer = persistedAnswer;
+
 		viewModel.question.label = this.label;
 		viewModel.question.suffix = this.suffix;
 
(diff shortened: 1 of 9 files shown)


packages/forms-web-app/src/dynamic-forms/dynamic-components/number-entry/question.js
```diff
@@ -39,6 +39,17 @@ class NumberEntryQuestion extends Question {
 	prepQuestionForRendering(section, journey, customViewData, payload) {
 		let viewModel = super.prepQuestionForRendering(section, journey, customViewData, payload);
 
+		const answer = journey.response.answers[this.fieldName];
+		const persistedAnswer =
+			answer !== undefined && answer !== null && answer === 0
+				? `${answer}`
+				: answer !== undefined && answer !== null
+				? answer
+				: '';
```
DeanH1990, line 48: as it was agreed to let the number of witnesses question have a minimum of 0 this had a lot of strange behaviours everywhere as 0 is being treated as falsy. 

I couldnt figure out how to get the persistence on the question page to work for 0 without making it a string and all other numbers stay as numbers..

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -318,9 +318,12 @@ class Question {
 	 * }>}
 	 */
 	formatAnswerForSummary(sectionSegment, journey, answer, capitals = true) {
+		const isAnswerProvided = answer !== null && answer !== undefined && answer !== '';
 		const formattedAnswer = capitals
 			? capitalize(answer ?? this.NOT_STARTED)
-			: answer || this.NOT_STARTED;
+			: isAnswerProvided
+			? answer
+			: this.NOT_STARTED;
```
DeanH1990, line 326: Both this and the getAction function were treating 0 as a falsy value. 

0 was being saved in the db and the viewModel/ journeyResponse but the client side was seeing it as falsy so was leaving the 'Not started' text on the task list page and the action button as 'Answer'

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -318,9 +318,12 @@ class Question {
 	 * }>}
 	 */
 	formatAnswerForSummary(sectionSegment, journey, answer, capitals = true) {
+		const isAnswerProvided = answer !== null && answer !== undefined && answer !== '';
```
Si3rr4wow, line 321: empty strings are also falsy, couldn't you just handle zero like this and store a number?

DeanH1990: Not quite sure what you mean here. 0 is/ was being passed/ saved as a number. It was being saved in to the journey and db as 0 as expected but on the task list and question page 0 wouldnt be shown. The task list would show 'Not started' and question page would have an empty field as if it hadnt been answered but looking in the logs it was in the response

In dynamic-components/number-entry/question.js 0 was being passed in as a number but in order for it to display the only way I found was converting it to string:

> const persistedAnswer =
			answer !== undefined && answer !== null && answer === 0
				? `${answer}`
				: answer !== undefined && answer !== null
				? answer
				: '';

packages/forms-web-app/src/dynamic-forms/dynamic-components/number-entry/question.js
beejjacobs: If this is the logic we need, we should consider putting it into a simple utility function. That way you can use `if` and `return` instead of nested ternaries. 

e.g.

```
// or some sensible name
const asNumberAnswer(answer) => {
  if (answer === 0) {
    return answer.toString();
  }
  // etc...
}'
```

DeanH1990: > Wondering if it needs some more fundamental changes, is something in the dynamic display meant to be checking for null/undefined, but just checks for falsey?

Im thinking there is a check somewhere thats meant to be checking for null/ undefined but is checking falsey but I can find anything if it is this. 

Even if changing the question.value to be 0 which is what the nunjucks template is using it still doesnt display unless making it string. Again this makes me think its to do with a falsy check hiding somewhere.

Also realised the section wasnt being marked as complete in the section.js file as it was checking if there was no answer which 0 evaluated to, meaning the submit button wouldnt appear if all other answers were completed.

I'll add a util function

Si3rr4wow (approved):

PR #3696, opened 2024-07-02
Feat/aapd 1809 s78 site area

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/AAPD-1809

## Description of change

Creates a new dynamic component question type - unitOptionEntry.  This is similar to radio and option questions (although it extends the base Question class) and allows for radio buttons to select a unit type, with a conditional then appearing to enter the quantity.  For example, entering a site area but having different unit options (square metres, hectares etc).

Also creates a new validator - existing conditional validators were not applicable given the fact that there are multiple conditionals on this page type.

## Checklist

- X New features have tests

## Important

### DIFF
--- a/packages/database/src/schema.prisma
+++ b/packages/database/src/schema.prisma
@@ -375,6 +375,7 @@ model AppealCase {
   appellantGreenBelt Boolean?
 
   siteAreaSquareMetres Decimal?
+  siteAreaUnits        String?
 
   agriculturalHolding                Boolean?
   tenantAgriculturalHolding          Boolean?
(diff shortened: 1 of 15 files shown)


packages/database/src/schema.prisma
```diff
@@ -375,6 +375,7 @@ model AppealCase {
   appellantGreenBelt Boolean?
 
   siteAreaSquareMetres Decimal?
+  siteAreaUnits        String?
```
bryn-pins, line 378: do we need this on appeal case for now? Should probably wait until defined in data model to make things easier going forward

topickering: Fair point, removed it for now

bryn-pins (approved):

PR #3728, opened 2024-07-16
Feat/aapd 1096 display ip comment data

DeanH1990:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-480

## Description of change

shows ip comments for appellant and lpa. 

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/src/lib/client-side/comment-actions.js
```diff
@@ -0,0 +1,38 @@
+document.addEventListener('DOMContentLoaded', () => {
+	const readMoreButtons = document.querySelectorAll('.read-more');
+	const closeButtons = document.querySelectorAll('.close-button');
+	readMoreButtons.forEach((button) => {
```
bryn-pins, line 4: It might not be worth doing but it would be preferable to make this reusable and work with a njk macro

In my head that would look something like:
```
attach on click event to page body
  check if target has .read-more class
  lookup text via a parent element by classname e.g. read-more-container
  toggle text content
  toggle button text content
```

bryn-pins (commented):
ok a number of comments but mostly very minor, nice work
need to run it locally with nvda still as worried about accessibility

packages/forms-web-app/src/sass/expand-collapse-comments.scss
```diff
@@ -0,0 +1,12 @@
...
+    display: none;
+}
+
+.visible {
+    display: inline;
+}
+
+.expand-collapse-button {
+    color: #1D70B8;
+    cursor: pointer;
```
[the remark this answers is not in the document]
DeanH1990: When i was testing what was working using <a> without the href doesnt automatically make the cursor a pointer. However when adding href it reloads the page or wherever its linked to. Having an empty string reloads the page and doesnt leave the comment expanded, having a # reloads the page and does leave the comment expanded. 

Can use javascript to stop the default behaviour of the page reloading as it goes back to the top of the page otherwise. 

Happy to remove the cursor from here and add the href and js to stop the default, not sure which way is preferable?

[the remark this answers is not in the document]
bryn-pins: ah ok, my mistake can leave as is

packages/forms-web-app/src/sass/expand-collapse-comments.scss
```diff
@@ -0,0 +1,12 @@
+.toggle-button {
+    display: none;
+}
+
+.js-enabled .toggle-button {
+    display: inline;
```
DeanH1990, line 6: I did try using a `<noscript>` in the nunjucks to hide buttons if JS wasnt enabled instead of doing this but couldnt get it to work. The buttons were keen to stay, this way removes them if the default full text is shown and no JS enabled

bryn-pins: So BO seem to have something similar:
https://github.com/Planning-Inspectorate/appeals-back-office/blob/main/appeals/web/src/client/components/excerpts/excerpt.js

Havne't gone over it myself yet, might be worth seeing if there's bits in there we can reuse, e.g. tests + fallbacks

packages/forms-web-app/src/views/selected-appeal/ip-comment-details.njk
```diff
@@ -31,7 +32,11 @@
...
-			
-		</div>
-	</div>
+			{% for comment in appeal.formattedComments %}
+                <h2 class="govuk-heading-l">{{ comment.key.text }}</h2>
+                {{renderComment(comment.value, loop.index, "Comments") }}
+            {% endfor %}
+        </div>
+    </div>
+    <script src="/public/javascripts/commentActions.bundle.js"></script>
```
bryn-pins, line 41: this should probably be
{% block pageScripts %}
    <script src="/public/javascripts/commentActions.bundle.js"></script>
{% endblock %}

or use defer

packages/forms-web-app/src/lib/client-side/comment-actions.js
```diff
@@ -0,0 +1,31 @@
+document.addEventListener('DOMContentLoaded', () => {
+	document.body.classList.add('js-enabled');
```
bryn-pins, line 2: this can go

packages/forms-web-app/src/views/macros/expand-collapse-comment.njk
```diff
@@ -0,0 +1,15 @@
+{% macro renderComment(comment, index, key) %}
+    <dl class="govuk-summary-list">
+        <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key">{{ key }}</dt>
+            <dd class="govuk-summary-list__value">
+                <span id="comment-text-{{ index }}" class="comment-text" data-truncated-text="{{ comment.truncatedText }}" data-full-text="{{ comment.text }}">
+                    {{ comment.text }}
+                </span>
+                {% if comment.truncated %}
+                    <a id="toggle-button-{{ index }}" class="toggle-button expand-collapse-button">Read more</a>
```
bryn-pins, line 10: needs href="#" to be accessible from keyboard

bryn-pins: use event.preventDefault to stop page jumping

packages/forms-web-app/src/views/selected-appeal/ip-comment-details.njk
DeanH1990: buttons stopped working using the block so went with defer

bryn-pins: it would have had to have been outside the block it's already in, was using locally so deffo works, this is good though

bryn-pins (approved):

packages/forms-web-app/src/views/selected-appeal/ip-comment-details.njk
DeanH1990: Ah ok, good to know thank you!

PR #3734, opened 2024-07-19
fix(489): add css to page

topickering:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/controller.js
+++ b/packages/forms-web-app/src/dynamic-forms/controller.js
@@ -8,6 +8,10 @@ const ListAddMoreQuestion = require('./dynamic-components/list-add-more/question
 const questionUtils = require('./dynamic-components/utils/question-utils');
 const { storePdfAppellantSubmission } = require('../../src/services/pdf.service');
 const { CONSTS } = require('../../src/consts');
+const {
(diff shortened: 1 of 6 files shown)


packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -368,10 +370,14 @@ exports.appellantSubmissionInformation = async (req, res) => {
 		summaryListData.sections.push(sectionView);
 	}
 
+	const css = fs.readFileSync(path.resolve(__dirname, '../public/stylesheets/main.css'), 'utf8');
```
Si3rr4wow, line 373: Might be nicer to do https://expressjs.com/en/starter/static-files.html rather than dealing with fs yourself (this should shut codeql up)

topickering: This particular file of css is being passed to njk as a variable, so not sure how to use express.static in this case - isn't it used at root with app.use?  Or can you use express.static in the controller?

bryn-pins: and linking to it doesn't work? Does it need to be inline?

bryn-pins (approved):

PR #3742, opened 2024-07-26
Feat/a2 178 submitted page

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-178

## Description of change

Add submitted page, move interested party session get/set etc to service

## Checklist

- [ ] New features have tests

## Important


topickering:
We should consider where and when (if at all) to scrub the interestedParty session data - at the moment it gets overwritten when viewing an open appeal but query whether we should wipe it on submission.

PR #3749, opened 2024-07-31
fix(a2-603): use simpler check for doc ref as not all are uuids

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-603

## Description of change

These unfortunately aren't all uuids, have reverted to a path traversal in the handler

## Checklist

- [ ] New features have tests

## Important


topickering (approved):
It was too good to be true

PR #3774, opened 2024-08-17
Refactor/improve journey oop

Si3rr4wow:
This has been bugging me for a while. Each journey is it's own class which extends the base `Journey` class. I think this was probably done to bind together all the journey specific params so `getJourney` in `journey-factory.js` could create a journey with just a `journeyId`. Here I've preserved this functionality but exported functions for building the params based on a `JourneyResponse` instance. 

Once this is done the next thing is to generalise the journey factory concept, the mapping from `journeyId` to a journey param builder needs to be pass-in-able from outside the module

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,5 @@
     }
   },
   "jestrunner.jestCommand": "node --experimental-vm-modules node_modules/jest/bin/jest.js",
-  "files.exclude": {
-    "**/*.test.js": true
-  }
+  "files.exclude": {}
 }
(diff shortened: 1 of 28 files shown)


Si3rr4wow:
Beware the base of this includes my comments on the state of dynamic-forms

Si3rr4wow:
I think the individual journey tests might not be needed as they're just less robust versions of the journey class tests

bryn-pins (commented):
Think this all seems sensible, but think we have a new journey now that will need refactoring before approval

bryn-pins (approved):

PR #3775, opened 2024-08-18
Refactor/make sections declarative

Si3rr4wow:
Makes journey sections a fixed parameter by pushing conditional rendering knowledge into `Question`s. `withCondition` now takes a function of response returning a boolean. The functions are evaluated when rendering the task list or moving to the next question.

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/controller.js
+++ b/packages/forms-web-app/src/dynamic-forms/controller.js
@@ -363,6 +363,10 @@ exports.appellantSubmissionInformation = async (req, res) => {
 			if (question.taskList === false) {
 				continue;
 			}
+			// don't show question on tasklist if it's hidden from journey
+			if (!question.shouldDisplay(journeyResponse)) {
+				continue;
+			}
 
 			const answers = journey.response?.answers;
 			let answer = answers[question.fieldName];
--- a/packages/forms-web-app/src/dynamic-forms/controller.js
+++ b/packages/forms-web-app/src/dynamic-forms/controller.js
@@ -129,6 +129,10 @@ exports.list = async (req, res, pageCaption, viewData) => {
 				continue;
 			}
 
+			if (!question.shouldDisplay(journeyResponse)) {
+				continue;
+			}
+
 			const answers = journey.response?.answers;
 			let answer = answers[question.fieldName];
 			const conditionalAnswer = questionUtils.getConditionalAnswer(answers, question, answer);
--- a/packages/forms-web-app/src/dynamic-forms/dynamic-components/list-add-more/question.js
+++ b/packages/forms-web-app/src/dynamic-forms/dynamic-components/list-add-more/question.js
@@ -1,3 +1,4 @@
+// const { getJourney } = require('../../journey-factory');
 const Question = require('../../question');
 const AddMoreQuestion = require('../add-more/question');
 
@@ -303,8 +304,7 @@ class ListAddMoreQuestion extends Question {
 		}
 
 		// move to the next question
-		const updatedJourney = new journey.constructor(journeyResponse);
-		return this.#handleNextQuestion(res, updatedJourney, section.segment, this.fieldName);
+		return this.#handleNextQuestion(res, journey, section.segment, this.fieldName);
 	};
 
 	/**
--- a/packages/forms-web-app/src/dynamic-forms/journey.js
+++ b/packages/forms-web-app/src/dynamic-forms/journey.js
@@ -218,7 +218,7 @@ class Journey {
 						);
 					}
 
-					if (!takeNextQuestion && question.fieldName === questionSegment) {
+					if (question.fieldName === questionSegment) {
 						takeNextQuestion = true;
 					}
 				}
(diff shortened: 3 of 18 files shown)


Si3rr4wow:
Don't merge me till https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/3774 is in

packages/forms-web-app/src/dynamic-forms/journey.js
```diff
@@ -206,14 +206,14 @@ class Journey {
 					reverse ? j-- : j++
 				) {
 					const question = currentSection.questions[j];
-					if (takeNextQuestion) {
+					if (takeNextQuestion && question.shouldDisplay(this.response)) {
```
bryn-pins, line 209: getStatus of section will probably need to handle question.shouldDisplay as well I think?
trying to think of anywhere else it may be needed

packages/forms-web-app/src/dynamic-forms/journey.js
```diff
@@ -206,14 +206,14 @@ class Journey {
...
 						return this.#buildQuestionUrl(
 							currentSection.segment,
 							question.url ? question.url : question.fieldName
 						);
 					}
 
-					if (question.fieldName === questionSegment) {
+					if (!takeNextQuestion && question.fieldName === questionSegment) {
```
bryn-pins, line 216: is this needed? I feel like it still takes the next question without this

Also now we'll have a consistent set of sections/questions it feels like it should be possible to simplify this next question logic, for another time though

packages/forms-web-app/src/dynamic-forms/dynamic-components/list-add-more/question.js
```diff
@@ -303,8 +304,7 @@ class ListAddMoreQuestion extends Question {
 		}
 
 		// move to the next question
-		const updatedJourney = new journey.constructor(journeyResponse);
```
bryn-pins, line 306: well I'm glad these can go

packages/forms-web-app/src/dynamic-forms/question.js
```diff
@@ -63,6 +63,8 @@ class Question {
 	html;
 	/** @type {string|undefined} optional question type */
 	type;
+	/** @type {(response: JourneyResponse) => boolean} */
+	shouldDisplay = () => true;
```
bryn-pins, line 67: minor point but shouldDisplay feels little frontend/display focussed, can't currently think of a good enough alternative name to suggest though so happy to keep

bryn-pins (commented):
think this is all seems sensible and is nicely constrained

packages/forms-web-app/src/dynamic-forms/journey.js
Si3rr4wow: IIRC this didn't work without that when I tested it but I might be wrong. I'll experiment

packages/forms-web-app/src/dynamic-forms/dynamic-components/list-add-more/question.js
Si3rr4wow: I'm not sure these were ever needed? the journey coming in was always made with the latest response

packages/forms-web-app/src/dynamic-forms/question.js
Si3rr4wow: That is a fair point. Happy to rename in future if we think of something better

packages/forms-web-app/src/dynamic-forms/journey.js
Si3rr4wow: Ok that is a new concept I wasn't aware of

packages/forms-web-app/src/dynamic-forms/controller.js
```diff
@@ -129,6 +129,10 @@ exports.list = async (req, res, pageCaption, viewData) => {
 				continue;
 			}
 
+			if (!question.shouldDisplay(journeyResponse)) {
```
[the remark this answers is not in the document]
Si3rr4wow: I think it slotted in pretty easy but I might've misunderstood what that function is doing

packages/forms-web-app/src/dynamic-forms/journey.js
Si3rr4wow: I've stuck a condition in the getStatus method, lemme know if you think I've done the right thing there. WRT the looping I'm wondering if either sections or journeys should have a method to get all the relevant questions ie omitting ones that aren't displayed

packages/forms-web-app/src/dynamic-forms/journey.js
Si3rr4wow: you were right it didn't need this, removed

packages/forms-web-app/src/dynamic-forms/controller.js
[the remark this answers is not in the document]
bryn-pins: yeah should be fine, it's the same idea

packages/forms-web-app/src/dynamic-forms/journey.js
bryn-pins: yes it should be that simple and the section completion logic should now be correct
A single method could certainly centralize that logic rather than needing to remember to do it everywhere!

bryn-pins (dismissed):
nice stuff

topickering (approved):

PR #3776, opened 2024-08-19
Fix/a2 772 bfs

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-722
https://pins-ds.atlassian.net/browse/A2-780
https://pins-ds.atlassian.net/browse/A2-782

## Description of change

Fixes bugs relating to Before You start journey, particularly adding display of 'back' button on several pages.  Note that the changes require the cessation of scrubbing session.appeal data if not eligible for an appeal.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/full-appeal/you-cannot-appeal.js
+++ b/packages/forms-web-app/src/controllers/full-appeal/you-cannot-appeal.js
@@ -7,7 +7,7 @@ const getYouCannotAppeal = async (req, res) => {
 	const { appealDeadline, appealPeriod } = appeal.eligibility;
 
 	// Clear appeal object
-	req.session.appeal = null;
+	// req.session.appeal = null;
 
 	return res.render(VIEW.YOU_CANNOT_APPEAL, {
 		appealDeadline,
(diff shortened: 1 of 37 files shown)


packages/forms-web-app/src/controllers/full-appeal/you-cannot-appeal.js
```diff
@@ -7,7 +7,7 @@ const getYouCannotAppeal = async (req, res) => {
 	const { appealDeadline, appealPeriod } = appeal.eligibility;
 
 	// Clear appeal object
-	req.session.appeal = null;
+	// req.session.appeal = null;
```
bryn-pins, line 10: unsure on the implications of this?
will any questions filled in remain set
does this impact the check on deadline that kicks you out? i.e. can you change your answer to deadline or will it block users from progressing

topickering: I have remvoed the back link from this page and reinstated the appeal scrubbing

bryn-pins: did it break things?

topickering: It did cause issues, especially in terms of trying to navigate to the decision date page!   It would be necessary to just wipe the decision date props but that could have had problems if the decisions date wasn't the issue, so would have required a few further checks and so fell outside the timebox we set.

bryn-pins (approved):
can't see any issues here

PR #3846, opened 2024-09-26
fix: use data model for caseType/procedure in mapping

bryn-pins:
## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/services/back-office-v2/formatters/s78/appeal.js
+++ b/packages/appeals-service-api/src/services/back-office-v2/formatters/s78/appeal.js
@@ -9,6 +9,7 @@ const {
 } = require('../utils');
 const deadlineDate = require('@pins/business-rules/src/rules/appeal/deadline-date');
 const { APPEAL_ID } = require('@pins/business-rules/src/constants');
+const { APPEAL_CASE_PROCEDURE, APPEAL_CASE_TYPE } = require('pins-data-model');
 
 /**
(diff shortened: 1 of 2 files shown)


packages/appeals-service-api/src/services/back-office-v2/formatters/s78/appeal.js
```diff
@@ -57,8 +58,8 @@ exports.formatter = async (appellantSubmission) => {
 	return {
 		casedata: {
 			submissionId: appellantSubmission.appealId,
-			caseType: 'D',
-			caseProcedure: 'written',
+			caseType: APPEAL_CASE_TYPE.W,
```
topickering, line 61: Almost certainly my ignorance of business logic, but how come the Case Type is W for s78 and D for HAS?

topickering: Does caseType correspond directly to HAS, S78 etc?

bryn-pins: D is always HAS, so was just copied from HAS incorrectly

I'm not sure yet if we need to change the case type for S78 based on the answers provided, but W should be S78 written

topickering: Thanks!

topickering (approved):

PR #3874, opened 2024-10-11
Feat/a2 1191 rule 6 login email

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-1191
https://pins-ds.atlassian.net/browse/A2-1221
https://pins-ds.atlassian.net/browse/A2-1207
https://pins-ds.atlassian.net/browse/A2-1204
https://pins-ds.atlassian.net/browse/A2-1202
https://pins-ds.atlassian.net/browse/A2-1223
https://pins-ds.atlassian.net/browse/A2-1239

## Description of change

Adds the new log-in routes for Rule 6 Users

## Checklist

- [X New features have tests

## Important


packages/forms-web-app/src/controllers/rule-6/enter-code.js
```diff
@@ -0,0 +1,197 @@
...
+ * @property {boolean} isGeneralLogin - defines if this enter code journey is for a general appeal log in, unrelated to an appeal
+ */
+
+/**
+ * Creates a one time password grant via the auth server
+ * @param {string} email
+ * @param {string} action
+ * @returns {Promise<void>}
+ */
+const createOTPGrant = async (email, action) => {
```
bryn-pins, line 26: lets reuse this function in auth-client perhaps

bryn-pins (commented):
looks fairly clean
Only thing I think we're missing is a means to know if it's a rule 6 enabled user or not, I think currently this allows either an appellant or LPA user into the rule6 dashboard, which I _think_ we should stop?

probably a case of separate middleware
and a separate createRule6UserSession in user service, that checks they have a rule6 role on any appeal

topickering:
Yes, I did wonder whether we want to check for R6 Users specifically via middleware - we've not moved on the the dashboard yet but that would only show appeals linked to that Rule6 user, but I suppose a benefit of the middleware might be that it wouldn't even allow them to see an empty dashboard.

packages/forms-web-app/src/controllers/rule-6/enter-code.js
topickering: Good idea - there is a fair amount of common code which could be refactored here I think.

bryn-pins (approved):

PR #3878, opened 2024-10-17
feat(a2-1239): add rule 6 user middleware

topickering:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/repositories/sql/appeal-user-repository.js
+++ b/packages/appeals-service-api/src/repositories/sql/appeal-user-repository.js
@@ -87,9 +87,6 @@ class AppealUserRepository {
 		return await this.dbClient.appealUser.findUnique({
 			where: {
 				email
-			},
-			include: {
-				Rule6Parties: true
 			}
 		});
 	}
@@ -104,6 +101,23 @@ class AppealUserRepository {
 		return await this.dbClient.appealUser.findUnique({
 			where: {
 				id
+			}
+		});
+	}
+
+	/**
+	 * Get a user with Rule 6 Parties relation by email
+	 *
+	 * @param {string} email
+	 * @returns {Promise<AppealUser|null>}
+	 */
+	async getWithRule6Parties(email) {
+		return await this.dbClient.appealUser.findUnique({
+			where: {
+				email
+			},
+			include: {
+				Rule6Parties: true
 			}
 		});
 	}
--- a/packages/appeals-service-api/src/repositories/sql/appeal-user-repository.js
+++ b/packages/appeals-service-api/src/repositories/sql/appeal-user-repository.js
@@ -87,6 +87,9 @@ class AppealUserRepository {
 		return await this.dbClient.appealUser.findUnique({
 			where: {
 				email
+			},
+			include: {
+				Rule6Parties: true
 			}
 		});
 	}
--- a/packages/appeals-service-api/src/spec/api-types.d.ts
+++ b/packages/appeals-service-api/src/spec/api-types.d.ts
@@ -577,9 +577,7 @@ export interface AppealUser {
 	isLpaAdmin?: boolean;
 	/** if an LPA user, the status of this user, e.g. have they logged in and confirmed their email */
 	lpaStatus?: 'added' | 'confirmed' | 'removed';
-	Appeals?: object[];
 	Rule6Parties?: object[];
-	SecurityToken?: object;
 }
 
 /** A final comment submitted by an appellant */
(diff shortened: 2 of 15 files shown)


packages/forms-web-app/src/controllers/rule-6/enter-code.js
```diff
@@ -51,6 +51,12 @@ const getEnterCodeR6 = (views) => {
 
 		const email = getSessionEmail(req.session, false);
 
+		const isRule6User = await isRule6UserByEmail(req, email);
+
+		if (!isRule6User) {
+			return renderEnterCodePage(`/${views.EMAIL_ADDRESS}`);
```
bryn-pins, line 57: I think this gives away that a user has rule6 status
See last scenario on A2-1202
User should be able to post a code but it should just always fail, not redirect to email
But we don't want them to receive an email with a code, so just render page without generating

packages/forms-web-app/src/controllers/rule-6/enter-code.js
```diff
@@ -105,14 +110,13 @@ const postEnterCodeR6 = (views) => {
...
-			enterCodeId,
-			sessionEmail,
-			action,
-			req.session?.appeal?.lpaCode,
-			true
-		);
+		const isRule6User = await isRule6UserByEmail(req, sessionEmail);
+
+		if (!isRule6User) {
+			return res.redirect(`/${views.EMAIL_ADDRESS}`);
```
bryn-pins, line 116: put this check after the other token checks and do
return renderError('Enter the correct code');

packages/appeals-service-api/src/spec/appeal-user.yaml
```diff
@@ -33,4 +33,18 @@ components:
         lpaStatus:
           type: string
           enum: ['added', 'confirmed', 'removed']
-          description: if an LPA user, the status of this user, e.g. have they logged in and confirmed their email?
+          description: if an LPA user, the status of this user, e.g. have they logged in and confirmed their email
+
+        # relations
```
bryn-pins, line 38: Do we return Appeals on some endpoints?
Security Token isn't anywhere is it, that should only be in Auth server?

packages/appeals-service-api/src/repositories/sql/appeal-user-repository.js
```diff
@@ -87,6 +87,9 @@ class AppealUserRepository {
 		return await this.dbClient.appealUser.findUnique({
 			where: {
 				email
+			},
+			include: {
```
bryn-pins, line 91: I think this would be better done in a separate api call/repo function (or base off of a param)
Majority of users aren't rule 6 and the majority of calls to this won't need this data and this is probably a hot path, so best to avoid repeat lookups to this table

bryn-pins: If interested you could look at using a scope request for the id token that returns if a user has rule6 status from the auth server. We can then check this on api requests without additional db lookups to secure any rule 6 api only endpoints

packages/forms-web-app/src/controllers/rule-6/enter-code.js
topickering: Part of my reasoning for putting it here was to avoid a non-Rule 6. user being able to use the override test code in dev and test.  Any problem with having the check here, before the token checks, but with the renderError rather than redirect?

packages/appeals-service-api/src/repositories/sql/appeal-user-repository.js
topickering: I've added a new api call rather than using the scope request - can investigate that further if you think it's a better approach

packages/forms-web-app/src/controllers/rule-6/enter-code.js
topickering: Actually, have put the rule 6 check after some token checks, makes sense and still prevents non-r6 using the test code, it occurs before the rule6 user session creation

bryn-pins: No problem think I was thinking through a way to get the tooManyAttempts error to still work for non rule 6 users, but that isn't really possible

yes main thing here is to render error not redirect

packages/appeals-service-api/src/spec/api-types.d.ts
```diff
@@ -575,8 +575,11 @@ export interface AppealUser {
 	lpaCode?: string;
 	/** if an LPA user, whether this user is an admin for that LPA */
 	isLpaAdmin?: boolean;
-	/** if an LPA user, the status of this user, e.g. have they logged in and confirmed their email? */
+	/** if an LPA user, the status of this user, e.g. have they logged in and confirmed their email */
 	lpaStatus?: 'added' | 'confirmed' | 'removed';
+	Appeals?: object[];
```
bryn-pins, line 580: need to run the autogen script?

topickering: Yep! Done now

bryn-pins (dismissed):

bryn-pins (approved):

PR #3886, opened 2024-10-22
Refactor/fix question typing

Si3rr4wow:
## Ticket Number

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/dynamic-components/utils/question-utils.js
```diff
@@ -42,3 +42,10 @@ exports.getListedBuildingForQuestion = (journeyResponse, fieldName) => {
 
 	return listedBuildings.filter((listedBuilding) => listedBuilding.fieldName == fieldName);
 };
+
+/** @type {(maybeDivider: import('../../question-props').Option) => maybeDivider is { divider: string }} */
+exports.optionIsDivider = (maybeDivider) => Object.hasOwn(maybeDivider, 'divider');
```
bryn-pins, line 47: ok this was confusing for me, so divider is a prop we use on options and pass to the govuk macro, and our logic needs to account for there not being a consistent object shape for option

Would it be better if we used a const for the string and reused that in the question instantiation + here?

packages/forms-web-app/src/dynamic-forms/dynamic-components/utils/question-utils.js
```diff
@@ -42,3 +42,10 @@ exports.getListedBuildingForQuestion = (journeyResponse, fieldName) => {
...
+
+/** @type {(conditional: unknown) => conditional is { html: string }} */
+exports.conditionalIsJustHTML = (conditional) =>
+	!!conditional && Object.hasOwn(conditional, 'html') && Object.keys(conditional).length === 1;
```
bryn-pins, line 51: when does this happen?

bryn-pins (commented):
mostly seems fine, type docs, checks, moving some more logic into overrides, utilising some spreads.

Main thing that confused me was the conditionals, couple of questions on that

packages/forms-web-app/src/dynamic-forms/dynamic-components/utils/question-utils.js
Si3rr4wow: This is a bit of an annoying one. I don't _think_ this check will ever be true. It's not until a few lines after this that the conditional value might be overwritten with compiled nunjucks.
```
			// handle conditional (dependant) fields & set their answers
			if (optionData.conditional !== undefined) {
				if (conditionalIsJustHTML(optionData.conditional)) continue;

				let conditionalField = { ...optionData.conditional };

				conditionalField.fieldName = getConditionalFieldName(
					this.fieldName,
					conditionalField.fieldName
				);
				conditionalField.value = payload
					? payload[conditionalField.fieldName]
					: journey.response.answers[conditionalField.fieldName] || '';

				optionData.conditional = {
					html: nunjucks.render(`./dynamic-components/conditional/${conditionalField.type}.njk`, {
						payload,
						...conditionalField,
						...customViewData
					})
				};
			}
```
I'd love to make this work in a smoother way at some point but I tried to alter the internals as little as possible to avoid regressions in this.

packages/forms-web-app/src/dynamic-forms/dynamic-components/utils/question-utils.js
Si3rr4wow: Yeah that's exactly it. Happy to constantify the key, it's only used twice

bryn-pins (approved):

PR #3888, opened 2024-10-22
Fix/a2 794 invalid column

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-794

## Description of change

Moves the 'INVALID' tag from 'Due In' to 'To Do' column

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/views/manage-appeals/your-appeals.njk
+++ b/packages/forms-web-app/src/views/manage-appeals/your-appeals.njk
@@ -34,6 +34,7 @@
                   <td class="govuk-table__cell">{{item.address}}</td>
                   <td class="govuk-table__cell">{{item.appealType}}</td>
                   <td class="govuk-table__cell">
+										{# if appeal is new or invalid this column left blank #}
                     {% if item.isNewAppeal %}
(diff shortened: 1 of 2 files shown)


packages/forms-web-app/src/views/manage-appeals/your-appeals.njk
```diff
@@ -42,7 +42,6 @@
                   </td>
                   <td class="govuk-table__cell govuk-table__cell--numeric">
 										{% if item.displayInvalid %}
```
bryn-pins, line 44: should we remove this and shuffle the elif into the first if?

topickering: I had kept it there because we have a default 'else' at the end of the if block, which will display the dueInDays figure and we want this column to remain blank (otherwise it would show 'days').  Might be able to reword that final else block.

bryn-pins: ah ok, can we get some sort of comment to that effect then please

bryn-pins (approved):

PR #3891, opened 2024-10-23
feat: change primary key to uuid for appeal-user relation

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-1352

## Description of change

- Update the PK of AppealToUser table to a uuid
- Need to allow users to have more than 1 role on an appeal to allow for rule6, (change from what was initially expected)
- Warning was given when trying to add role to the key, due to length, so changed to uuid

## Checklist

- [x] New features have tests

## Important

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
 				"eslint-plugin-jest": "^26.5.3",
 				"husky": "^8.0.0",
 				"jest": "^29.4.0",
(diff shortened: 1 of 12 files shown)


topickering (commented):
I think we also have some tests for the appeal-user-repository in `__tests__/developer/repositories` , not sure if one or both of the test files can go?

topickering (approved):

PR #3894, opened 2024-10-24
fix(a2-1236): amend call to confirm whether user is rule 6 user

topickering:
## Description of change

Amend functions and api calls which check whether user is a rule 6 user.  Now uses count to check for Rule6Party role on AppealToUser array on AppealUser.

## Checklist

- [ ] New features have tests

## Important


bryn-pins (dismissed):
think this all makes sense

bryn-pins (approved):

PR #3900, opened 2024-10-25
chore(A2-1478): update to node 20

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-1478

## Description of change

Update to node 20

Node >19 changes the default setting of http/s agents to use keep alive true, and the clients experience failures with this
Have set the global instance value for both to false

https://github.com/nodejs/node/issues/47130
https://github.com/node-fetch/node-fetch/issues/1767
https://github.com/axios/axios/issues/5929

## Checklist

- [ ] New features have tests

## Important


README.md
```diff
@@ -26,9 +26,9 @@ Note: when starting a new appeal as an appellant, if you cannot progress beyond
 Install NodeJS using [nvm](https://github.com/nvm-sh/nvm#installing-and-updating)
 
 ```
-nvm install 18
-nvm use 18
-nvm alias default 18
+nvm install 20.17.0
+nvm use 20.17.0
```
beejjacobs, line 30: Perhaps leave for now, but I think you can just do `nvm install 20` to get the latest minor version?

packages/appeals-service-api/src/app.js
```diff
@@ -11,6 +11,11 @@ require('express-async-errors');
 require('./controllers/appeals-for-submission-to-horizon-scheduler')();
 const app = express();
 
+const https = require('https');
+const http = require('http');
+https.globalAgent = new https.Agent({ keepAlive: false });
+http.globalAgent = new http.Agent({ keepAlive: false });
```
beejjacobs, line 17: Might be worth a comment along with this code to explain why its there? not obvious otherwise - same with other places.

PR #3916, opened 2024-11-01
fix(a2-986): handle parsing and displaying of dates in UK time

bryn-pins:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-986

## Description of change

- Standardises the display of dates from BO
- When parsing user inputted dates treats input as UK time before saving as UTC

## Checklist

- [x] New features have tests

## Important


edpackard-work (dismissed):

packages/common/src/lib/format-date.js
```diff
@@ -0,0 +1,64 @@
+const { formatInTimeZone, zonedTimeToUtc, utcToZonedTime } = require('date-fns-tz');
+const { isValid } = require('date-fns');
+
+const ukTime = 'Europe/London';
```
beejjacobs, line 4: If I'm picky (which I usually am...), a better name would be `ukTimezone`

packages/forms-web-app/src/dynamic-forms/dynamic-components/date/question.js
```diff
@@ -81,7 +84,7 @@ class DateQuestion extends Question {
 			const answerDateString = journey.response.answers[this.fieldName];
 
 			if (answerDateString) {
-				const answerDate = new Date(answerDateString);
+				const answerDate = convertUTCToUK(answerDateString);
 				day = `${answerDate.getDate()}`.slice(-2);
 				month = `${answerDate.getMonth() + 1}`.slice(-2);
 				year = `${answerDate.getFullYear()}`;
```
beejjacobs, line 90: Is this tested? the `get*` methods are local time (which for the server is UTC) - you might need to use `formatInTimeZone` for each element?

bryn-pins: Have added some tests, is that enough to prove it works, or have I misunderstood how convertUTCToUK works?

packages/common/src/lib/format-date.js
bryn-pins: done

beejjacobs (approved):
Comments addressed, new code is clear and avoids issues with date objects being changed.

PR #3966, opened 2024-12-03
fix(a2-1790): amend check on removed files to prevent api error

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-1790

## Description of change

Amends the check when removing files - now makes sure that not all files removed rather than just one (potentially of many)

## Checklist

- [ ] New features have tests

## Important


edpackard-work (approved):
my only question would be - there are a few other multifile upload validation components like this - will they all suffer from the same problem, or is it unique to this one?

Si3rr4wow (approved):

DeanH1990 (approved):

PR #3990, opened 2024-12-16
fix(a2-183): make upload booleans nullable

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-183

## Description of change
Make upload[documents] fields nullable in spec files in order to prevent api errors if deleted uploaded files.

Following the call to remove files (in the getDataToSave function) the array of SubmissionDocumentUploads on the journeyResponse is now updated - this fixes and issue where the removed file continued to be shown on the upload page and the action on the task list remained at change rather than upload, even if the doc was removed.  This was caused by the fact that the journeyResponse was not updated until the next page was called and the previouslyUploadedDocuments array did not take account of the removed files.

## Checklist

- [ ] New features have tests

## Important


BenSurgisonPINS (approved):
LGTM

bryn-pins (approved):

PR #4032, opened 2025-01-22
fix: handle race with document put calls

bryn-pins:
## Description of change

Deadletter in prod for document
This handles a simple retry if we get a unique constraint error in the upsert
This should only occur if an update happens while the create is still running

## Checklist

- [x] New features have tests

## Important


fabfab-kc (approved):

edpackard-work (approved):

edpackard-work (commented):
looks good - as long as you're happy not to have a limit on the number of retries

bryn-pins:
> looks good - as long as you're happy not to have a limit on the number of retries

I think it should be fine, due to the error type it should only happen once...

PR #4054, opened 2025-01-31
Feat/a2 2163

topickering:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-2163

## Description of change

Adds integration function to listen for broadcast of representation and trigger put in appeals api

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/representations/repo.js
```diff
@@ -106,6 +148,59 @@ class RepresentationsRepository {
...
+					? { connect: { caseReference: data.caseReference } }
+					: { create: {} }
+			},
+			update: mappedData,
+			where: {
+				representationId
+			}
+		});
+
+		if (data.documentIds?.length) {
```
topickering, line 174: I am not entirely sure about this.  What I am looking to do is confirm that we have a Document with the relevant document id.  We will then only make a related RepresentationDocument link where this is true. 
I considered findUniqueOrThrow but count seems to work.
There is a question about whether we should upsert in order to create a Document if one doesn't exist, but this requires additional information which is not in the representation message broadcast

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/representations/repo.js
```diff
@@ -106,6 +148,59 @@ class RepresentationsRepository {
...
+				data.documentIds.filter(async (documentId) => {
+					(await this.dbClient.document.count({
+						where: {
+							id: documentId
+						}
+					})) > 0;
+				})
+			);
+
+			await this.dbClient.$transaction(async (tx) => {
```
topickering, line 185: Query whether this deletion is required

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/representations/index.js
```diff
@@ -26,4 +29,6 @@ router.use(
 
 router.get('/', asyncHandler(getAppealCaseWithRepresentations));
 
+// router.put('/:representationId', asyncHandler(putByRepresentationId));
```
bryn-pins, line 32: Assume you're aware of this,
Also can we get tests for this endpoint

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/representations/repo.js
bryn-pins: I think looking at this we need to switch this around
I think we can't know if the document will come in before or after the representation
At which point the representationDocument can't have a foreign key to document

So we need to:

1. Remove the foreign key on representationDocument in prisma

In this function if we have any doc ids
2. Delete all 
3. Upsert all regardless of if the document exists or not

And then handle the chance of missing documents when we pull from a representation

packages/forms-web-app/src/controllers/selected-appeal/representations/index.js
```diff
@@ -56,7 +56,7 @@ exports.get = (representationParams, layoutTemplate = 'layouts/no-banner-link/ma
 		const lpa = await getDepartmentFromCode(caseData.LPACode);
 		const headlineData = formatHeadlineData(caseData, lpa.name, userType);
 
-		const formattedRepresentations = formatRepresentations(representationsForDisplay);
+		const formattedRepresentations = await formatRepresentations(representationsForDisplay, req);
```
topickering, line 59: Query whether better to pass the apiClient rather than the req

packages/forms-web-app/src/lib/representation-functions.js
```diff
@@ -190,53 +191,67 @@ const formatProofsOfEvidenceHeading = (userType, submittingParty, rule6OwnRepres
 
 /**
  * @param {Representation[]} representations
+ * @param {import('express').Request} req
  */
-const formatRepresentations = (representations) => {
+const formatRepresentations = async (representations, req) => {
```
topickering, line 196: query whether should receive an apiclient rather than the whole req

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/representations/repo.js
topickering: Done

packages/appeals-service-api/src/routes/v2/appeal-cases/_caseReference/representations/index.js
topickering: Uncommented

packages/forms-web-app/src/controllers/selected-appeal/representations/index.js
bryn-pins: probably but lets get this in

bryn-pins (approved):

PR #4091, opened 2025-02-24
fix(A2-2272): stop using duplicate fieldname to for back link to func…

DeanH1990:
…tion correctly

## Ticket Number

https://pins-ds.atlassian.net/browse/A2-2272

## Description of change

fixes back link navigation for other appeals question in all appeal types

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/questions.js
+++ b/packages/forms-web-app/src/dynamic-forms/questions.js
@@ -1627,7 +1627,7 @@ exports.questionProps = {
 			type: 'case',
 			title: 'Enter the appeal reference number',
 			question: 'Enter the appeal reference number',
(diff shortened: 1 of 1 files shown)


DeanH1990:
> * So it seems this fieldname is not currently used in the database?
> * Will have to update the submission mapping to use this field

I think fieldname is misleading for the subquestions. as the ones for the nearby/ linked appeals are inserted into the SubmissionLinkedCase table under the fieldname property as they are arrays of case ref numbers so not actually stored on the submission form/ appealcase itself
e.g just from my local one:
image

bryn-pins (approved):
ah sorry for not looking deeper

PR #4112, opened 2025-02-27
feat: a2-2371 - added new question for s20 - what grade is the listed…

peter-uzor-pins:
… building

## Ticket Number

A2-2371

https://pins-ds.atlassian.net/browse/A2-2371

## Description of change

Radio - What grade is the listed building? - S20
New question added with condition based on previous question answer.

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/dynamic-forms/questions.js
```diff
@@ -2499,6 +2499,28 @@ exports.questionProps = {
...
+		type: 'radio',
+		title: 'What grade is the listed building?',
+		question: 'What grade is the listed building?',
+		fieldName: 'listedBuildingGrade',
+		url: 'listed-building-grade',
+		validators: [new RequiredValidator('Select the grade of the listed building')],
+		options: [
+			{
+				text: 'Grade I',
+				value: 'Grade I'
```
bryn-pins, line 2513: for the values can we please use:

- 'I'
- 'II*'
- 'II'

And maybe put them as an enum/const map in packages/common/src/constants.js
I am hoping this will help avoid any mistakes in subsequent mappings 
Shame they are not already on the data model

peter-uzor-pins: Ok cool. Makes sense. I will get it done.

peter-uzor-pins: @bryn-pins  the only issue i see here is that if the values are I,II* and II then on the questionnaire, these values are displayed. Its not entirely user friendly but i guess it can work.
But as a user i would probably want to see Grade I instead of I on the questionnaire page. Unless there's a way around this to display the text instead.

peter-uzor-pins:
Ahhhhh, i did a rebase and for some reason it has added all these other changes to my PR.

packages/forms-web-app/src/dynamic-forms/questions.js
bryn-pins: hmm good point, we should probably be displaying the text value on the task list.
happy to ignore on this one, but we shouldn't be tying what we show to users to what is held in the database, I'll have a look into it

PR #4114, opened 2025-02-28
feat: a2-2371 - added new question for s20 - what grade is the listed…

peter-uzor-pins:
… building

## Ticket Number

A2-2371

https://pins-ds.atlassian.net/browse/A2-2371

## Description of change

Radio - What grade is the listed building? - new S20 question

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/appeals-service-api/src/spec/api-types.d.ts
+++ b/packages/appeals-service-api/src/spec/api-types.d.ts
@@ -1038,7 +1038,8 @@ export interface LPAQuestionnaireSubmission {
 	SubmissionDocumentUpload?: object[];
 	appealNotification?: boolean | null;
 	demolishAlterExtend?: boolean | null;
-	consultHistoricEngland?: boolean;
+	consultHistoricEngland?: boolean | null;
+	listedBuildingGrade?: string;
 }
 
 /** A statement submitted by an LPA */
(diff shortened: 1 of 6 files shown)


bryn-pins (dismissed):

packages/appeals-service-api/src/spec/api-types.d.ts
```diff
@@ -1038,7 +1038,8 @@ export interface LPAQuestionnaireSubmission {
 	SubmissionDocumentUpload?: object[];
 	appealNotification?: boolean | null;
 	demolishAlterExtend?: boolean | null;
-	consultHistoricEngland?: boolean;
+	consultHistoricEngland?: boolean | null;
```
toyin-pins, line 1041: what's the need of explicitly defining that boolean could b null value here? Since we store this value as a bit in the DB anyway it's only 2 states that we're interested in no?

peter-uzor-pins: The booleans were defined as nullable. I just thought i'd stick to the way things were done already. I don't think it matters.

toyin-pins (approved):

packages/appeals-service-api/src/spec/api-types.d.ts
bryn-pins: So the way we've been doing these is to define the api spec manually and then autogenerating this file with `npm run gen-api-types` in the api package

PR #4134, opened 2025-03-07
A2-Fix/a2 505 live issue

peter-uzor-pins:
## Ticket Number

A2-505

https://pins-ds.atlassian.net/browse/A2-505

## Description of change

Live Issue - after changing the invalid decision date to valid date the user is taken back to start of the appeal journey.

Implemented fix so that on link click, the session will be cleared instead of on page load.

## Checklist

- [ ] New features have tests

## Important


packages/forms-web-app/src/public/js/clear-appeal-session-event.js
```diff
@@ -0,0 +1,14 @@
+document.addEventListener('DOMContentLoaded', () => {
```
bryn-pins, line 1: Unfortunately I don't think we should use a client side request
We need to use [progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement) and I think this is fundamental enough to the user journey it should be handled server side

bryn-pins (approved):

PR #4159, opened 2025-03-13
fix: ensure if user is already created and removed then we just updat…

peter-uzor-pins:
…e the status on recreate

## Ticket Number

A2-1767

https://pins-ds.atlassian.net/browse/A2-1767

## Description of change

Can't re-add an LPA user

Issue: When user is removed the user is marked as removed but when the user is readded, it tries to create a new user with the same email.

Solution: Checked if there was an existing user and then reset the status from Removed to Added.

## Checklist

- [ ] New features have tests

## Important


bryn-pins (commented):
Can we get a test case for this in users.spec.js please

bryn-pins (approved):

PR #4173, opened 2025-03-18
feat: A2-1217 added new lpa table and v2 api endpoints

peter-uzor-pins:
## Ticket Number

 A2-1217 

https://pins-ds.atlassian.net/browse/A2-1217 

## Description of change

Create an LPA model prisma model + migration that uses the same fields as Mongo

V2 API endpoint that returns retrieves all lpas from sql

V2 API endpoint that fetches single LPA by lpaCode

V2 API endpoint that fetches single LPA by lpa19CD

V2 API endpoint that replaces the existing list of lpas with a new one
POST, takes a semi colon separated csv file, deletes all records and adds all new ones

## Checklist

- [ ] New features have tests

## Important


packages/appeals-service-api/data/populate-db.sh
bryn-pins, line 1: Could we add the equivalent of:
dev/data/data/appeals-service-api/001-lpa.json

To
packages/database/src/seed/data-dev.js

So that the two test lpas are added by default when running the site locally and you don't need to run the separate script

packages/appeals-service-api/src/routes/v2/lpa/index.js
```diff
@@ -0,0 +1,12 @@
+const express = require('express');
+const router = express.Router();
+const asyncHandler = require('@pins/common/src/middleware/async-handler');
+const controller = require('./controller');
+
+router.get('/', asyncHandler(controller.getAll));
```
bryn-pins, line 6: could we add openApiValidatorMiddleware to each of the endpoints please
am aware some endpoints don't have this but we want it to be the default going forward

packages/appeals-service-api/src/routes/v2/lpa/service.js
```diff
@@ -0,0 +1,75 @@
...
+ * @returns {Promise<void>}
+ */
+exports.createLpaList = async (csv) => {
+	const mapper = new LpaMapper();
+	const lpaEntities = mapper.csvJsonToLpaEntities(csv).map((lpaEntity) => lpaEntity.toJson());
+	logger.debug(lpaEntities, 'LPA entities as JSON');
+
+	try {
+		await repo.remove();
+		const lpaChunks = chunkArray(lpaEntities, 10);
```
bryn-pins, line 47: can we move the batch size into a const

packages/appeals-service-api/src/routes/v2/lpa/repo.js
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * @param {any[]} data
+	 */
+	async upsertMany(data) {
+		const mappedData = data.map((record) => {
+			return mapDataModelToEntity(record);
+		});
+		try {
+			await this.dbClient.lPA.createMany({
```
bryn-pins, line 76: I have a feeling that this won't actually upsert?
Feels as if the test for this only checks the result of the mapped data not what is actually added to the database as that is what is returned in the response?

packages/appeals-service-api/src/routes/v2/lpa/service.js
```diff
@@ -0,0 +1,75 @@
...
+		throw ApiError.lpaUpsertFailure();
+	}
+};
+
+/**
+ * @param {any} myArray
+ * @param {number} chunk_size
+ * @returns {Promise<any>}
+ */
+const chunkArray = (myArray, chunk_size) => {
```
bryn-pins, line 64: feels like this could be shared function with some unit tests as it's been written to be fairly generic, perhaps in common package?

bryn-pins (commented):
Few minor comments, but it mostly looks good!

packages/appeals-service-api/src/routes/v2/lpa/repo.js
bryn-pins: On second look, we're doing what I originally suggested and deleting all of the existing LPAs
I think that's a mistake as it stops us from adding a foreign key ref in future
Although I'm not sure prisma has an efficient means to bulk upsert

peter-uzor-pins: Oh this works fine. I ran a manual test locally and it works fine.

The test confirms the number of records upsert are the same as the number added. Also manual tests confirm the DB is updated without any issues

bryn-pins: it will break when we add a foreign key ref to other tables, but I guess we can fix it then

bryn-pins (approved):

PR #4186, opened 2025-03-21
feat: a2-1854 increased default max characters on input to 8000

peter-uzor-pins:
## Ticket Number

A2-1854

https://pins-ds.atlassian.net/browse/A2-1854

## Description of change

Increased default max characters on input to 8000 - Changes have to be done in BO before this is released.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/questions.js
+++ b/packages/forms-web-app/src/dynamic-forms/questions.js
@@ -65,7 +65,7 @@ const formatNumber = require('./dynamic-components/utils/format-number');
 /** @typedef {import('./question-props').QuestionProps} QuestionProps */
 /** @typedef {import('./question')} Question */
 
-const inputMaxCharacters = Math.min(Number(configInputMaxCharacters), 1000);
+const inputMaxCharacters = Math.min(Number(configInputMaxCharacters), 8000);
 
 /**
  * @param {'past' | 'future'} tense
--- a/packages/database/src/migrations/20250325121436_increased_string_maximum_for_new_character_limit/migration.sql
+++ b/packages/database/src/migrations/20250325121436_increased_string_maximum_for_new_character_limit/migration.sql
@@ -1,0 +1,27 @@
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[AppellantSubmission] ALTER COLUMN [siteAreaUnits] NVARCHAR(max) NULL;
+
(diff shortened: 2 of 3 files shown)


bryn-pins (commented):
We'll need to update the database fields to match this change
To allow us to merge this we could make this an environment variable

peter-uzor-pins:
> We'll need to update the database fields to match this change To allow us to merge this we could make this an environment variable

peter-uzor-pins:
> We'll need to update the database fields to match this change To allow us to merge this we could make this an environment variable

@bryn-pins, thanks, i forgot about the DB changes. I have made the changes. But with regards to environment variable, we already have an environment variable we can use, the issue is that it gets the min between the environment variable and the value entered i.e. 8000. So technically, we can control this using the environment variable - CHARACTER_LIMIT_FINAL_COMMENT, this is set to 32500 by default and then depending on the environment, its set to whatever its set to. So i would suggest leaving it as is and if there's any request to reduce it then we can use the environment variable currently in place.

packages/forms-web-app/src/dynamic-forms/questions.js
```diff
@@ -65,7 +65,7 @@ const formatNumber = require('./dynamic-components/utils/format-number');
 /** @typedef {import('./question-props').QuestionProps} QuestionProps */
 /** @typedef {import('./question')} Question */
 
-const inputMaxCharacters = Math.min(Number(configInputMaxCharacters), 1000);
+const inputMaxCharacters = Math.min(Number(configInputMaxCharacters), 8000);
```
bryn-pins, line 68: Sorry, shouldn't this PR only be changing the value for the addNewConditions question, or have I misunderstood the ticket, so the other values should remain at 1000?

We'll need to update the equivalent column in appealCase table (newConditionDetails) as well

I suggested a specific env var to allow us to merge this. We can also just leave this blocked/unmerged until the back office PR is approved. Although even then that ties our releases together more closely than I would like

Looking at the limits more closely we could make some simplifications as well:
- we could remove inputMaxCharacters and use the values from characterLimits.appealFormV2
- we could then also remove finalComment: configInputMaxCharacters as it won't be used anymore

peter-uzor-pins: Ok i don't know what i was doing while reading the ticket. I have decided to create a new branch and redo this and raise another PR. I will close off this PR.

PR #4211, opened 2025-03-31
Fix/a2 2655

peter-uzor-pins:
## Ticket Number

A2-2655

https://pins-ds.atlassian.net/browse/A2-2655

## Description of change

Fix for Tab focus skips to continue button instead of error message.

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/views/before-you-start/enforcement-notice.njk
+++ b/packages/forms-web-app/src/views/before-you-start/enforcement-notice.njk
@@ -18,12 +18,13 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-
       {% if errors %}
+      <div {% if focusErrorSummary %} tabindex="-1" autofocus {% endif %}>
         {{ govukErrorSummary({
           titleText: "There is a problem",
           errorList: errorSummary
         }) }}
+        </div>
       {% endif %}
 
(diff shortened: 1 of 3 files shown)


packages/forms-web-app/src/views/before-you-start/enforcement-notice.njk
```diff
@@ -18,12 +18,13 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-
       {% if errors %}
+      <div {% if focusErrorSummary %} tabindex="-1" autofocus {% endif %}>
```
bryn-pins, line 22: So the govuk framework auto focuses the error summary already, with tabindex=-1 and js.
How does this interact with that?

peter-uzor-pins: > So the govuk framework auto focuses the error summary already, with tabindex=-1 and js. How does this interact with that?

@bryn-pins , for this particular view, i started testing in safari on my phone and then resolved the issue because the focus always moved back to the button after error and rerender. After this fix, i tested on chrome on my phone and also the web browser and it works as expected. So personally i dont think the framework is correctly autofocusing in this case, so i am basically forcing it to focus.

bryn-pins (approved):

PR #4240, opened 2025-04-08
fix: a2-2970 using jsdom to sanitize html for the pdf document download

peter-uzor-pins:
## Ticket Number

A2-2970

https://pins-ds.atlassian.net/browse/A2-2970

## Description of change

Used jsdom to sanitize html for the pdf document download

## Checklist

- [ ] New features have tests

## Important


packages/pdf-service-api/src/lib/generatePdf.js
```diff
@@ -20,7 +20,7 @@ const generatePdf = async (html) => {
 			waitUntil: 'domcontentloaded'
 		});
 
-		await page.emulateMediaType('screen');
+		await page.emulateMediaType('print');
```
bryn-pins, line 23: If we're actually using the print styling can we not just use a css media query rather than removing with jsdom?

```css
@media print {
  .govuk-back-link,
  .govuk-phase-banner,
  footer {
    display: none !important;
  }
}
```

peter-uzor-pins: Makes sense, i have tested and it works fine. The approach is a lot lighter so i will delete this PR and branch and raise another one

PR #4244, opened 2025-04-09
Feat/a2 2564 login redirect

DeanH1990:
## Ticket Number

https://pins-ds.atlassian.net/browse/A2-2564

## Description of change

redirects lpa/appellant/rule-6 user to requested page after login

## Checklist

- [x] New features have tests

## Important


packages/forms-web-app/jest.config.js
```diff
@@ -18,5 +18,9 @@ module.exports = {
 			lines: 82
 		}
 	},
-	setupFilesAfterEnv: ['<rootDir>/__tests__/setupTests.js']
+	setupFilesAfterEnv: ['<rootDir>/__tests__/setupTests.js'],
+	moduleNameMapper: {
+		'^@pins/common(.*)$': '<rootDir>/../common$1',
+		'^@pins/business-rules(.*)$': '<rootDir>/../business-rules$1'
+	}
```
DeanH1990, line 25: Found webstorm wouldnt resolve functions from packages outside of forms web app without this.

Not sure if leaving it in would cause issues for people using other IDE's? Happy to revert if so.

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -208,6 +211,17 @@ const postEnterCode = (views, { isGeneralLogin = true }) => {
 			return handleCustomRedirect();
 		}
 
+		if (isSelectedPageRequest) {
```
bryn-pins, line 214: if this becomes a check for any redirect scenario
we can do:

```js
if (hasLoginRedirect) {
  return handleCustomRedirect();
}
```

We can then lose `isV2DocRequest`
And remove the subsequent `if (req.session.loginRedirect)`
And may as well combine `delete req.session.tempBackLink;` into `deleteTempSessionValues`

packages/forms-web-app/src/controllers/rule-6/enter-code.js
```diff
@@ -140,6 +142,19 @@ const postEnterCodeR6 = (views) => {
...
+			if (req.session.tempBackLink) {
+				req.session.navigationHistory = [
+					req.session.tempBackLink,
+					...req.session.navigationHistory
+				];
+				delete req.session.tempBackLink;
+			}
+			const redirect = req.session.loginRedirect;
+			delete req.session.loginRedirect;
+			return res.redirect(redirect);
```
bryn-pins, line 155: combine and use `delete req.session.loginRedirect` into `deleteTempSessionValues();`

packages/forms-web-app/src/lib/login-redirect.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @param {string} basePath
+ * @returns {(req: import('express').Request) => void}
+ */
+const storeAppealPageRedirect = (basePath) => {
```
bryn-pins, line 5: I don't think this needs to return a function?

packages/forms-web-app/src/lib/login-redirect.js
```diff
@@ -0,0 +1,26 @@
...
+
+		if (
+			urlSegments[0] === basePath &&
+			urlSegments[1] &&
+			urlSegments[1].match(/^\d{7}$/) &&
+			urlSegments.length > 2
+		) {
+			tempBackHistory = `/${urlSegments[0]}/${urlSegments[1]}`;
+		}
+		if (req.originalUrl.startsWith(`/${basePath}/`) && urlSegments.length > 1) {
```
bryn-pins, line 19: do we need basePath, can it just be done for all urls
the login just won't work in other scenarios?

packages/forms-web-app/src/lib/login-redirect.js
```diff
@@ -0,0 +1,26 @@
...
+
+		const urlSegments = req.originalUrl.split('/').filter(Boolean);
+
+		if (
+			urlSegments[0] === basePath &&
+			urlSegments[1] &&
+			urlSegments[1].match(/^\d{7}$/) &&
+			urlSegments.length > 2
+		) {
+			tempBackHistory = `/${urlSegments[0]}/${urlSegments[1]}`;
```
bryn-pins, line 17: Which pages are left that are behind a login middleware and use a session navigation history back button?
Would really like to avoid using it completely, if anything still does can we look to hardcode them

bryn-pins (commented):
I think we can rework and simplify this

packages/forms-web-app/src/lib/login-redirect.js
DeanH1990: Maybe dont need a basePath however it doesnt work for all urls.

If user inputs only the basePath e.g. just /appeals/, manage-appeals/  etc it will try redirect to this page which does not exist hence checking the length is more than just that segment

packages/forms-web-app/src/lib/login-redirect.js
DeanH1990: This tempBackLink was mainly for the sub sections of an appeal  - appeal-details, questionnaire etc. these pages use the nav history for the back button. Can definitely get them hardcoded - just back to the appeal overview page of the related appeal?

As the back link currently uses the nav history, if a user then inputs one of these subsections for a different appeal, the back button on that page would go back to where they were. Changing it will then make it go back to the main page for the associated appeal, shall I double check this is what we want with design?

bryn-pins: There isn't another route into these pages right that I am aware of

packages/forms-web-app/src/lib/login-redirect.js
bryn-pins: hmm ok, I feel as if it would be clearer for those page to redirect to login or the listing page based on if the user is logged in or not

packages/forms-web-app/src/lib/login-redirect.js
DeanH1990: I dont think so, no.

packages/forms-web-app/src/lib/login-redirect.js
DeanH1990: I kept basePath in as a param as this allows specifying the user where its being used in each users require-user/ check logged in function.

The length check ignores if the url entered is just /appeals/ etc so it does just go to the dashboard after logging in

packages/forms-web-app/src/controllers/common/enter-code.js
```diff
@@ -168,6 +168,8 @@ const postEnterCode = (views, { isGeneralLogin = true }) => {
 
 		const isV2DocRequest = req.session?.loginRedirect?.startsWith('/appeal-document/');
 
+		const isSelectedPageRequest = req.session?.loginRedirect?.startsWith('/appeals/');
```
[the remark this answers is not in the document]
DeanH1990: Have combined these to just use a isLoginRedirect variable

bryn-pins (approved):

PR #4281, opened 2025-04-29
fix: included redacted documents in zip download

peter-uzor-pins:
## Ticket Number

A2-1664

https://pins-ds.atlassian.net/browse/A2-1664

## Description of change

Ensure all published documents are downloadable in zip

## Checklist

- [ ] New features have tests

## Important

### DIFF
--- a/packages/forms-web-app/src/controllers/selected-appeal/representations/index.js
+++ b/packages/forms-web-app/src/controllers/selected-appeal/representations/index.js
@@ -69,7 +69,10 @@ exports.get = (representationParams, layoutTemplate = 'layouts/no-banner-link/ma
 		let pdfDownloadUrl;
 		let zipDownloadUrl;
 		const ipDocuments = caseData?.Documents?.filter(
-			(doc) => doc.documentType === documentTypes.interestedPartyComment.name && doc.published
+			(doc) =>
+				doc.documentType === documentTypes.interestedPartyComment.name &&
+				doc.published &&
+				doc.redacted
 		);
 		if (
 			(userType === APPEAL_USER_ROLES.APPELLANT || userType === LPA_USER_ROLE) &&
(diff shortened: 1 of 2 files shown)


packages/forms-web-app/src/controllers/selected-appeal/representations/index.js
```diff
@@ -69,7 +69,7 @@ exports.get = (representationParams, layoutTemplate = 'layouts/no-banner-link/ma
 		let pdfDownloadUrl;
 		let zipDownloadUrl;
 		const ipDocuments = caseData?.Documents?.filter(
-			(doc) => doc.documentType === documentTypes.interestedPartyComment.name && !doc.redacted
```
bryn-pins, line 72: I think we should be checking both of these booleans

peter-uzor-pins: @bryn-pins I had a call with @Lpsatti  and she said that it should be displayed to download whether its redacted or not. But as long as its published, it should be displayed in the FO, hence my change.

peter-uzor-pins: @bryn-pins  Ok, going through the code, first of all we only include published documents in the caseData. So all documents displayed are published. Also Looking at BO, the document owner is never set. Meaning that we do not know who the document owner is.

So technically the solution i guess would be to ensure that the documents are published and redacted.

bryn-pins: To clarify a little, most of the time we can figure out the owner based on the document type, i.e. certain doc types are only uploaded by the appellant or lpa.

In this case they are from Interested Parties who are essentially anon (they have no associated user record) so will never be 'owned' by anyone, so should always be redacted

PR #4302, opened 2025-05-09
fix: removed published flag check

peter-uzor-pins:
## Ticket Number

A2-1644

https://pins-ds.atlassian.net/browse/A2-1644

## Description of change

Removed the published check because the flag is not returned in the query but all documents returned are all published. So there is no need for the flag and if we include it, it returns published=false.

## Checklist

- [x] New features have tests

## Important


bryn-pins (approved):
If it's checked by the API then this is fine

PR #4316, opened 2025-05-16
fix: remove unredacted documents from zip

peter-uzor-pins:
### Description of change

Removed unredacted documents from zip download file

Ticket: https://pins-ds.atlassian.net/browse/A2-1664

### Checklist

- [ ] Feature complete and ready for users, or behind feature-flag
- [ ] Commit history is linear with no merge commits

### Important

### DIFF
--- a/packages/document-service-api/src/routes/v2/back-office/_caseReference/case_stage/_caseStage/controller.js
+++ b/packages/document-service-api/src/routes/v2/back-office/_caseReference/case_stage/_caseStage/controller.js
@@ -25,6 +25,7 @@ async function getBlobCollection(caseStage, caseReference) {
 		)
 	)
 		.flat()
+		.filter((doc) => doc.redacted)
 		.map((document) => {
 			const { filename, documentURI, documentType } = document;
 			return {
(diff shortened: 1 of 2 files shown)


packages/document-service-api/src/routes/v2/back-office/_caseReference/case_stage/_caseStage/controller.js
```diff
@@ -25,6 +25,7 @@ async function getBlobCollection(caseStage, caseReference) {
 		)
 	)
 		.flat()
+		.filter((doc) => doc.redacted)
```
bryn-pins, line 28: Sorry but I think this has to be made more complicated...

We have to filter redacted out only if the user does not 'own' the document
See checkDocAccess in /src/routes/v2/back-office/_id/controller.js
As I believe this endpoint is also used in a scenario where a user downloads their own documents before they've been checked on BO, at which point unredacted is fine to show

So we need to loop over each doc and check the user is currently able to download them

Note that I'm of the opinion that the logic for grouping documents together using these FOLDERS should be moved to properties on the each of the document types 
But that doesn't need to be done here, just mentioning for future, I want to remove that const eventually

peter-uzor-pins: Makes sense. For now i will just add logic to only be able to return redacted documents for certain documentTypes. Depending on the user and documentTypes, but this would be determined by the documentType then we can split them if needed in the near future.

bryn-pins:
handled here https://github.com/Planning-Inspectorate/appeal-planning-decision/pull/4328

PR #4425, opened 2025-06-30
feat(A2-3744): removing application-reference from BYS journey

toyin-pins:
### Description of change
- Redirecting to email-address page instead
- Ensured that for V1 versions, the user still gets redirected to application number page since this isn't asked inside the appeal form
- Updated tests since now checking feature flags so function has become async

Ticket: https://pins-ds.atlassian.net/browse/A2-3744

### Checklist

### Important


bryn-pins (changes requested):
We'll need to check for v1/v2
v1 does not ask the question in the appeal form so will be missing this data if we don't ask in BYS

PR #4475, opened 2025-07-31
feat: increase test coverage for dynamic forms

bryn-pins:
### Description of change

Prep for dynamic forms refactor
- cas-adverts/adverts use the same journey, removing duplicate
- increase test coverage with snapshot testing for dynamic forms
- increase test coverage for multifile-upload
- remove appeals specific doc type object from dynamic forms types
- removed some magic strings

### Checklist

### Important


packages/forms-web-app/src/lib/documents-api-wrapper.js
```diff
@@ -108,6 +108,8 @@ const createDocument = async (submission, data, fileName, documentType, sectionT
 
 	if (isTheFormDataBuffer(data)) {
 		body.append('file', fs.createReadStream(data.tempFilePath), documentName);
+	} else if (Buffer.isBuffer(data?.data)) {
```
bryn-pins, line 111: this seemed to behave differently under super test to an actual run, so added another variation here. Not sure the case below this is ever hit, but reluctant to remove it while v1 exists

andy-mair121 (approved):

PR #4504, opened 2025-08-12
Feat/a2 3911 linked appeal appellant dashboard

topickering:
### Description of change

Adds linked cases functionality for FO. Includes:
- api - functions to retrieve linked cases for either single or multiple appeals
- web - functions to establish whether an appeal is a linked-lead or linked-child and display accordingly

This PR concerns the Appellant dashboards and Appeal pages, a subsequent PR will implement for the LPA pages (the code is in place)

Seed data cases 0000066 (lead), 0000666 (child)  and 0006666 (child) have been designated as linked cases.

Ticket: https://pins-ds.atlassian.net/browse/A2-3911

### Checklist

### Important

### DIFF
--- a/packages/appeals-service-api/src/spec/appeal-case.yaml
+++ b/packages/appeals-service-api/src/spec/appeal-case.yaml
@@ -1,7 +1,7 @@
 components:
   schemas:
     AppealCaseDetailed:
-      description: An appeal case from the Back Office, with users, relations and documents
+      description: An appeal case from the Back Office, with users, relations, documents and linkedCase info
       allOf:
         - $ref: '#/components/schemas/AppealCase'
         - type: object
@@ -22,6 +22,20 @@ components:
               type: array
               items:
                 $ref: '#/components/schemas/Document'
+            linkedCases:
+              type: array
+              items:
+                type: object
+                required:
+                  - childCaseReference
+                  - leadCaseReference
+                properties:
+                  childCaseReference:
+                    type: string
+                    description: the child case reference
+                  leadCaseReference:
+                    type: string
+                    description: the lead case reference
 
     AppealCase:
       description: An appeal case from the Back Office
--- a/packages/database/src/schema.prisma
+++ b/packages/database/src/schema.prisma
@@ -429,6 +429,7 @@ model AppealCaseRelationship {
   // indexes
   @@index([caseReference])
   @@index([caseReference2])
+  @@index([type])
 }
 
 /// Lookup table for types of appeals processes, and appeal types such as D (Householder) and W (S78).
--- a/packages/forms-web-app/src/controllers/selected-appeal/index.js
+++ b/packages/forms-web-app/src/controllers/selected-appeal/index.js
@@ -34,6 +34,7 @@ const { getDepartmentFromCode } = require('../../services/department.service');
 const logger = require('#lib/logger');
 const config = require('../../config');
 const { formatDateForDisplay } = require('@pins/common/src/lib/format-date');
+const { formatDashboardLinkedCaseDetails } = require('#lib/linked-appeals');
 
 /** @type {Partial<import('@pins/common/src/view-model-maps/sections/def').UserSectionsDict>} */
 const userSectionsDict = {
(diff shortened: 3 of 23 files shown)


packages/appeals-service-api/src/spec/appeal-case.yaml
```diff
@@ -22,6 +22,20 @@ components:
...
+              type: array
+              items:
+                type: object
+                required:
+                  - caseReference
+                  - caseReference2
+                properties:
+                  caseReference:
+                    type: string
+                    description: the child case reference
```
bryn-pins, line 35: is it worth it to return the child? won't that always be the same as the caseRef.
I think it makes sense to simplify this for API response and contain the whole caseRef1/2 logic to as small a layer as possible in the code e.g. just repo if not too tricky

Potentially map back to the same structure as the data model for consistency e.g.
"linkedCaseStatus",
"leadCaseReference",

packages/database/src/schema.prisma
```diff
@@ -429,6 +429,7 @@ model AppealCaseRelationship {
   // indexes
   @@index([caseReference])
   @@index([caseReference2])
+  @@index([type])
```
bryn-pins, line 432: nice, think I thought this was already here

packages/forms-web-app/src/controllers/selected-appeal/index.js
```diff
@@ -95,15 +102,18 @@ exports.get = (layoutTemplate = 'layouts/no-banner-link/main.njk') => {
...
 
-			shouldDisplayQuestionnaireDueNotification: isLPA && isLPAQuestionnaireDue(caseData),
-			shouldDisplayStatementsDueBannerLPA: isLPA && isLPAStatementOpen(caseData),
-			shouldDisplayProofEvidenceDueBannerLPA: isLPA && isLPAProofsOfEvidenceOpen(caseData),
-			shouldDisplayFinalCommentsDueBannerLPA: isLPA && isLPAFinalCommentOpen(caseData),
+			shouldDisplayQuestionnaireDueNotification:
+				!isChildAppeal && isLPA && isLPAQuestionnaireDue(caseData),
+			shouldDisplayStatementsDueBannerLPA: !isChildAppeal && isLPA && isLPAStatementOpen(caseData),
+			shouldDisplayProofEvidenceDueBannerLPA:
+				!isChildAppeal && isLPA && isLPAProofsOfEvidenceOpen(caseData),
```
bryn-pins, line 109: can we move this check into the various open rules, so all logic for if a journey is open is confined to the respective isOpen function

packages/forms-web-app/src/controllers/selected-appeal/index.js
```diff
@@ -95,15 +102,18 @@ exports.get = (layoutTemplate = 'layouts/no-banner-link/main.njk') => {
 			titleSuffix: formatTitleSuffix(userType),
 			backLinkToDashboard,
 
-			shouldDisplayQuestionnaireDueNotification: isLPA && isLPAQuestionnaireDue(caseData),
-			shouldDisplayStatementsDueBannerLPA: isLPA && isLPAStatementOpen(caseData),
-			shouldDisplayProofEvidenceDueBannerLPA: isLPA && isLPAProofsOfEvidenceOpen(caseData),
-			shouldDisplayFinalCommentsDueBannerLPA: isLPA && isLPAFinalCommentOpen(caseData),
+			shouldDisplayQuestionnaireDueNotification:
+				!isChildAppeal && isLPA && isLPAQuestionnaireDue(caseData),
```
bryn-pins, line 106: I thought they would submit a different LPAQ? Is it that it's just possible to link after they have separate LPAQs?

packages/appeals-service-api/src/spec/appeal-case.yaml
topickering: The thinking behind including the child caseReference was so that this property would give all child cases for a lead appeal, ie a child appeal would have one object in the linkedCases array, and the lead would have an object for each child without a need for an additional db call - all info needed for FO display would be there.

Can see that it might be confusing though, so could return to mapping to data model and then making a subsequent call (if required for display) for the lead.

packages/forms-web-app/src/controllers/selected-appeal/index.js
topickering: Will do

packages/appeals-service-api/src/spec/appeal-case.yaml
topickering: Good point, childCaseReference and leadCaseReference might be somewhat easier to decipher

PR #5371, opened 2026-04-29
fix: back links on comment-planning-appeal pages (A2-7989)

julietteorpen:
### Description of change

Changes in this PR: 

- Fixes "back" link behaviour across comment-planning-appeal pages so that enter appeal ref back link goes to GOV.uk page and enter postcode back to enter appeal ref 
- Refactors controllers to supply explicit back link context to the views/templates.
- Updates templates to use the new back link logic for consistent user navigation.
- Modifies and updates related tests to cover new back link implementation.

Ticket: https://pins-ds.atlassian.net/browse/A2-7989

### Checklist

### Important

### DIFF
--- a/packages/forms-web-app/src/routes/file-based-router/comment-planning-appeal/enter-appeal-reference/controller.js
+++ b/packages/forms-web-app/src/routes/file-based-router/comment-planning-appeal/enter-appeal-reference/controller.js
@@ -1,6 +1,13 @@
 /** @type {import('express').RequestHandler} */
 const enterAppealReferenceGet = (req, res) => {
-	res.render(`comment-planning-appeal/enter-appeal-reference/index`);
+	const previousUrl = req?.session?.navigationHistory[1]?.trim();
+	const backLink =
+		!previousUrl || previousUrl === '/comment-planning-appeal'
+			? 'https://www.gov.uk/'
+			: previousUrl;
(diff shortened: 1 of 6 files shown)


bryn-pins (commented):
I played around with this locally and got a few issues, seemed to have a blank url for the back link

packages/forms-web-app/src/routes/file-based-router/comment-planning-appeal/enter-appeal-reference/controller.js
```diff
@@ -1,6 +1,13 @@
 /** @type {import('express').RequestHandler} */
 const enterAppealReferenceGet = (req, res) => {
-	res.render(`comment-planning-appeal/enter-appeal-reference/index`);
+	const previousUrl = req?.session?.navigationHistory[1]?.trim();
```
bryn-pins: think this may need a navigationHistory defined check before the array lookup

bryn-pins: Also had some weird results with some 404s entering the history e.g. a missing .min.map.js call
That's not a result of this change though. If it only happens locally, then can probably ignore
Otherwise think we need to change how nav history works internally

PR #5382, opened 2026-05-07
fix(A2-7873): update decision tags for enforcement and elb

topickering:
### Description of change

Updates decision tags for enforcement and elb cases

Ticket: https://pins-ds.atlassian.net/browse/A2-7873
https://pins-ds.atlassian.net/browse/A2-8053

### Checklist

### Important

### DIFF
--- a/packages/business-rules/src/utils/decision-outcome.js
+++ b/packages/business-rules/src/utils/decision-outcome.js
@@ -25,10 +25,9 @@ function mapDecisionColour(decision) {
  *
  * @param {string|undefined} decision
  * @param {boolean} isEnforcement
- * @param {boolean} isLpa
  * @returns {string | null}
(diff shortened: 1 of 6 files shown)


zoebutterworth (commented):
Are there any tests which cover the tags in more detail? e.g. to check that ELB show the correct tags

topickering:
> Are there any tests which cover the tags in more detail? e.g. to check that ELB show the correct tags

ELB and Enforcement show the same tags, and the only difference to other appeal types is the wording on a split decision.  The 'isEnforcement' param covers both enforcement notice and elb and is passed rather than the caseType and so I think ELB are covered, but I can perhaps name some of the tests better!

zoebutterworth (approved):

PR #5407, opened 2026-05-29
[A2-7733] Updating e2e tests with multi environment config 

marcpodgorney81:
### Description of change

Currently in order to run the frontoffice tests across multiple environments it requires repeated updating of values within the `.env` file (e.g. have to keep changing urls etc.) 
This could result in different configs being setup and commented/uncommented, leading to an env file that is messy and difficult to interpret 

This PR introduces a multi config setup similar to the one used in backoffice tests, which provides a cleaner way to maintain environment specific values. It also allows for corresponding scripts/commands to be setup in the package.json  

For example to run the tests in dev environment you can now run 
`cy:open:dev` (UI/explorer mode) 
`cy:run:dev` (headless mode) 

The 'default' commands (i.e. `cy:open` `cy:run`) will use local urls  

Also as part of this I have updated the readme to reflect these changes 

Are a couple of things that would be good if could check 
- I've tried running the commands for different environments and seems to work ok, though would be good if someone could check and confirm 
- I've based the urls on information that found in env and readme files, and also what I have bookmarked, would appreciate a check of these 

Ticket: https://pins-ds.atlassian.net/browse/A2-7733 

### Checklist

### Important

### DIFF
--- a/test-packages/platform-feature-tests/cypress.config.dev.env.js
+++ b/test-packages/platform-feature-tests/cypress.config.dev.env.js
@@ -1,0 +1,21 @@
+// @ts-nocheck
+const { defineConfig } = require('cypress');
+const baseConfig = require('./cypress.config');
+
+require('dotenv').config();
+
+const e2eOverride = {
+	appeals_beta_base_url: 'https://appeals-service-dev.planninginspectorate.gov.uk',
+    back_office_base_url: 'https://back-office-appeals-dev.planninginspectorate.gov.uk',
+	apiBaseUrl: 'https://pins-app-appeals-bo-api-dev.azurewebsites.net/'
+};
+
+module.exports = defineConfig({
+    e2e: {
+        ...baseConfig.e2e,
+        ...e2eOverride
+    },
+    env: {
+        ...baseConfig.env
+    }
+});
--- a/test-packages/platform-feature-tests/cypress.config.local.env.js
+++ b/test-packages/platform-feature-tests/cypress.config.local.env.js
@@ -1,0 +1,27 @@
+// @ts-nocheck
+const { defineConfig } = require('cypress');
+const baseConfig = require('./cypress.config');
+
+require('dotenv').config();
+
+// use front office and backoffice defaults if not set
+if (!baseConfig.e2e.appeals_beta_base_url) {
+	baseConfig.e2e.appeals_beta_base_url = 'https://localhost:9003/';
+}
+
+if (!baseConfig.e2e.back_office_base_url) {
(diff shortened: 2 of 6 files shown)


test-packages/platform-feature-tests/cypress.config.dev.env.js
```diff
@@ -0,0 +1,21 @@
...
+const { defineConfig } = require('cypress');
+const baseConfig = require('./cypress.config');
+
+require('dotenv').config();
+
+const e2eOverride = {
+	appeals_beta_base_url: 'https://appeals-service-dev.planninginspectorate.gov.uk',
+    back_office_base_url: 'https://back-office-appeals-dev.planninginspectorate.gov.uk',
+	apiBaseUrl: 'https://pins-app-appeals-bo-api-dev.azurewebsites.net/'
+};
```
bryn-pins, line 11: Minor point but we may want to figure out how best to align variable casing and trailing `/` differences at some point, think it was ported from BO codebase so has different conventions. Willing to accept that it may be best left as is to allow easier porting of any future code changes

test-packages/platform-feature-tests/cypress.config.local.env.js
```diff
@@ -0,0 +1,27 @@
...
+if (!baseConfig.e2e.appeals_beta_base_url) {
+	baseConfig.e2e.appeals_beta_base_url = 'https://localhost:9003/';
+}
+
+if (!baseConfig.e2e.back_office_base_url) {
+	baseConfig.e2e.back_office_base_url = 'https://localhost:8080/';
+}
+
+if (!baseConfig.e2e.apiBaseUrl) {
+	baseConfig.e2e.apiBaseUrl = 'http://localhost:3000/';
```
bryn-pins, line 17: note that this won't work, we can't run FO/BO alongside each other 
port clashes + need a service bus/function app emulator

Perhaps best to leave local out for now, depends on how many tests actually connect to back office API I guess

bryn-pins (approved):

test-packages/platform-feature-tests/cypress.config.dev.env.js
marcpodgorney81: I agree, left as is, as wanted to get your feedback regarding conventions 
Would be good though to align/have some consistency

test-packages/platform-feature-tests/cypress.config.local.env.js
marcpodgorney81: I added this as works for backoffice repo, and thought may have some use as part of local development  
Could possibly leave local frontoffice urls in there, as assume that would work?

PR #5443, opened 2026-06-09
fix(forms): fixed missing fields in appeal details page (no-ticket)

musabadam-pins:
### Description of change

Added missing fields back into mapping
Added missing fields into appeal details page
added missing fields and section into questionnaire page
updated unit tests
Updated prisma

Ticket: https://pins-ds.atlassian.net/browse/A2-

### Checklist

- [ ] Feature complete and ready for users, or behind feature-flag
- [ ] Commit history is linear with no merge commits

### Important

### DIFF
--- a/packages/common/src/lib/format-questionnaire-details.js
+++ b/packages/common/src/lib/format-questionnaire-details.js
@@ -136,3 +136,13 @@ exports.formatProcedurePreference = (caseData) => {
  */
 exports.formatConditions = (caseData) =>
 	(caseData.newConditionDetails && `Yes\n${caseData.newConditionDetails ?? ''}`) || 'No';
+
+/**
+ * @param {AppealCaseDetailed} caseData
+ */
+exports.formatListOfDocumentsBeforeDecision = (caseData) => {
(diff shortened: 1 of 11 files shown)


packages/common/src/lib/format-questionnaire-details.js
```diff
@@ -136,3 +136,13 @@ exports.formatProcedurePreference = (caseData) => {
  */
 exports.formatConditions = (caseData) =>
 	(caseData.newConditionDetails && `Yes\n${caseData.newConditionDetails ?? ''}`) || 'No';
+
+/**
+ * @param {AppealCaseDetailed} caseData
+ */
```
lachlandow1, line 142: Ah only spotted this now, since this is a free text field I don't think this is needed. Something you might want to remove in another PR

lachlandow1 (approved):

packages/common/src/lib/format-questionnaire-details.js
musabadam-pins: ah true, this was one of the first one i added in, later ones dont follow the format function direction

PR #5445, opened 2026-06-09
fix(A2-8223): handle bst/utc on ip comments service

topickering:
### Description of change

Handle time zones on FO IP comments service.

Ticket: https://pins-ds.atlassian.net/browse/A2-8223

### Checklist

### Important


bryn-pins (commented):
Sorry another one I don't think is needed

Testing on staging the code seems to be accurate as is, to the minute

We are comparing a utc date in the database to a utc timestamp currently with this code so it is fine, converting both from utc into UK time before the comparison won't achieve anything

I think the ticket is misleading, if there was a timezone issue here, it would only lead to it being an hour out, where as they are suggesting an entire day. Maybe need to restore a copy of prod from before they modified dates in the database so we can see what the actual issue was, although that's a lot of work

PR #5554, opened 2026-07-24
fix(a2-8808): bump multer field limit for pdf

bryn-pins:
### Description of change

Bumps limit for multer in pdf service to allow larger pdf generation

Ticket: https://pins-ds.atlassian.net/browse/A2-8808

### Checklist

### Important

### DIFF
--- a/packages/pdf-service-api/src/app.test.js
+++ b/packages/pdf-service-api/src/app.test.js
@@ -1,7 +1,12 @@
 const supertest = require('supertest');
+
+const chunkSize = 512;
+const maxLength = 1 * chunkSize;
+process.env.FILE_MAX_SIZE_IN_BYTES = String(maxLength);
+const generatePdf = require('./lib/generatePdf');
+
 const app = require('./app');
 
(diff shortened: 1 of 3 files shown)


packages/pdf-service-api/src/app.test.js
```diff
@@ -24,13 +27,44 @@ describe('app', () => {
...
+				currentSize += chunk.length;
+			}
+
+			chunks.push('</body></html>');
+
+			const htmlContent = chunks.join('');
 
 			const response = await supertest(app).post('/api/v1/generate').field('html', htmlContent);
 
-			expect(generatePdf).toHaveBeenCalledWith(htmlContent);
```
zoebutterworth, line 31: Why did you remove this?

bryn-pins: ah because it then checks a huge html string and dumps it into console when it fails, which was taking ages to print out
perhaps I should bring back and massively reduce the limit in the test

zoebutterworth (approved):

PR #5625, opened 2026-08-14
fix(infrastructure): remove node installs from test pipelines

m-juckes-pins:
There is a install chrome script(file) that the original script was referencing, which is not needed as our agents have chrome installed already. This has been **removed.**


test-packages/platform-feature-tests/install-chromium.sh
```diff
@@ -1,3 +1,4 @@
+# This file could be removed now
```
m-juckes-pins: Cannot see any other references to this file in the project

azure-pipelines-e2e-appeals.yml
```diff
@@ -19,28 +19,28 @@ variables:
   - group: e2e
   - group: e2e-${{ parameters.Environment }}
 
-# PR trigger disabled, until process has been agreed by team
 pr: none
 trigger: none
 
-# pr:
```
m-juckes-pins: removed as not really adding much and left as comments for 9 months

azure-pipelines-pr.yml
```diff
@@ -19,7 +19,7 @@ resources:
       type: github
       endpoint: Planning-Inspectorate
       name: Planning-Inspectorate/common-pipeline-templates
-      ref: refs/tags/release/3.25.2
+      ref: refs/tags/release/3.29.0
```
m-juckes-pins: update to new tag if other pr is merged in

azure-pipelines-e2e-appeals.yml
```diff
@@ -73,54 +82,41 @@ jobs:
...
-          SMOKE_ONLY: ${{ parameters.SmokeOnly }}
-          TEST_TAG: ${{ parameters.TestTag }}
-        displayName: 'Run Cypress Tests'
+      - template: steps/node_script.yml@templates
+        parameters:
+          displayName: 'Run Cypress Tests for Appeals'
+          nodeVersion: 22
+          script: |
+            npm ci
+            npm audit fix
```
m-juckes-pins: `npm audit fix` needed to be added in. Without it the tests were not properly run, they were pending but still classed as passed.
Without the command: https://dev.azure.com/planninginspectorate/appeal-planning-decision/_build/results?buildId=204166&view=logs&j=6d3431c2-e7d6-5a83-1c79-2069804415b1&t=8e7afb79-6f48-50a0-b5f8-2d48f202278a

---

With the command : https://dev.azure.com/planninginspectorate/appeal-planning-decision/_build/results?buildId=204185&view=logs&j=6d3431c2-e7d6-5a83-1c79-2069804415b1&t=8e7afb79-6f48-50a0-b5f8-2d48f202278a

beejjacobs: That is very odd. We shouldn't be running audit fix in the pipeline IMO, that would be something to run locally and push up a change for if required.

azure-pipelines-e2e-appeals.yml
```diff
@@ -73,54 +82,41 @@ jobs:
...
-          SPEC: $(SPEC)
-          SMOKE_ONLY: ${{ parameters.SmokeOnly }}
-          TEST_TAG: ${{ parameters.TestTag }}
-        displayName: 'Run Cypress Tests'
+      - template: steps/node_script.yml@templates
+        parameters:
+          displayName: 'Run Cypress Tests for Appeals'
+          nodeVersion: 22
+          script: |
+            npm ci
```
beejjacobs: I think we should keep the install step separate, so that the test run is its own step as before. Makes the output easier to follow then too.

beejjacobs (commented):
Looks good if we can remove Node & Chromium install. Just a few comments to consider. Happy to discuss.

azure-pipelines-e2e-appeals.yml
m-juckes-pins: Fixed - non-issue, self made in a sense

azure-pipelines-e2e-appeals.yml
m-juckes-pins: Test is separated. Originally coded that way as a method to shorten the pipeline code

Medini-Aithal (approved):

PR #5638, opened 2026-08-18
fix(a2-8993): fail silently on missing document for delete

bryn-pins:
### Description of change

Avoid failing if delete message comes through for a document that does not exist

Ticket: https://pins-ds.atlassian.net/browse/A2-8993

### Checklist

### Important


topickering (approved):
Looks good (now we just need Prisma to randomly change their error codes!)

PR #5657, opened 2026-08-27
fix(A2-8132): remove duplicate application date question

topickering:
### Description of change

For full planning appeals (s78) which are not following the expedited route, we currently ask the 'when did you submit your planning application' question twice - during the BYS stage and during the Prepare Appeal section of the appeal form.  This PR removes the question from the Appeal Form.

Note, we do ask the question in some appeal forms, where we do not ask it during BYS.

Ticket: https://pins-ds.atlassian.net/browse/A2-8132

### Checklist

### Important

### DIFF
--- a/packages/forms-web-app/src/dynamic-forms/display-questions.js
+++ b/packages/forms-web-app/src/dynamic-forms/display-questions.js
@@ -10,6 +10,7 @@ const { isExpeditedAppealDate } = require('#lib/is-expedited-part1-eligible');
 /**
  * @typedef {import('@pins/dynamic-forms/src/journey-response').JourneyResponse} JourneyResponse
  * @typedef {import('@pins/dynamic-forms/src/question')} Question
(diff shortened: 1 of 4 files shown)


zoebutterworth:
Might be worth a quick check of whether this breaks the e2e tests :)

topickering:
Good point!

lachlandow1:
I don't think this can be removed from the S78 form completely - possibly could have a check for certain appeal types to still display or to update those journeys to include the question in the BYS. As far as I can see this would include:
MINOR_COMMERCIAL_DEVELOPMENT
LISTED_BUILDING
REMOVAL_OR_VARIATION_OF_CONDITIONS
There may be more that I've missed and it's only certain paths that need this.


This has been checked for the expedited types in a previous ticket. After we made the updates to the appeal forms:
S78_EXPEDITED
HOUSEHOLDER
CAS_ADVERTS
CAS_PLANNING

lachlandow1 (approved):

PR #5676, opened 2026-09-07
chore: use @planning-inspectorate/coding-standards

beejjacobs:
### Description of change

https://pins-ds.atlassian.net/browse/RDSD-264

also updates eslint to v10


eslint.config.mjs
```diff
@@ -0,0 +1,119 @@
+import { defineConfig, globalIgnores } from 'eslint/config';
+import { eslintConfig } from '@planning-inspectorate/coding-standards';
+import jest from 'eslint-plugin-jest';
+import cypress from 'eslint-plugin-cypress';
+import globals from 'globals';
```
bryn-pins, line 5: should globals be listed as a dep in package json

bryn-pins: already done

bryn-pins (approved):

